Issue #228, opened 2014-11-01
Full screen applications minimized


Christilut:
Some full screen applications, mostly games, get taken out of full screen mode when a notification happens. In my case I've noticed it in Battlefield 3 and 4. 

I think the notification should not happen if there is a full screen program running.

Aokromes:
I agree for a setting for this.

programcoder:
I also go for a setting

jonnius:
On Ubuntu this also happens when watching a Movie in the Browser.

Issue #127, opened 2014-08-21
Notification minimizes full screen applications


Christilut:
Some full screen applications lose their full screen state when a notification occurs. Notably this happens in Battlefield 3 for me, but there may be other applications that have this issue. 

Maybe suppress notifications when something is fullscreen or make sure they pop up below the full screen application.

telegramdesktop:
Notifications should not be shown now while fullscreen app is shown (from 0.7.10 version).

programcoder:
please make it a setting or something, because It's really important for me to see notifications on a video playblack, which didn't annoy me neither did it minimized my video.

PR #376, opened 2014-12-31
Suppress notifications when in D3D full screen and presentation mode

brynthomas:
Addresses issues:
telegramdesktop/tdesktop#228
telegramdesktop/tdesktop#127

Updated to include support for the Vista and later feature that allows one to query the OS to determine if notifications should be suppressed. http://msdn.microsoft.com/en-us/library/bb762242%28VS.85%29.aspx

Checks SHQueryUserNotificationState to see if the system is running a full screen exclusive D3D application, and hides popup notifications. If the Windows Mobility Center feature of  "I am currently giving a presentation" is enabled, sounds will be suppressed as well.

This does not impact regular full screen apps, or games running in Windowed (Borderless).

Tested on Windows 8.1 and Windows XP (where XP doesn't support the feature, but is not harmed by the presence of it).

### DIFF
--- a/Telegram/SourceFiles/pspecific_wnd.cpp
+++ b/Telegram/SourceFiles/pspecific_wnd.cpp
@@ -55,6 +55,7 @@ namespace {
 	bool useDWM = false;
 	bool useTheme = false;
 	bool useOpenAs = false;
+	bool useNotifySuspend = false;
 	bool themeInited = false;
 	bool finished = true;
 	int menuShown = 0, menuHidden = 0;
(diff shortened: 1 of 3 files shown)


Aokromes:
@brynthomas this PR cannot be automaticaly merged :)

brynthomas:
@Aokromes it was mergeable before and not merged, so not sure if it's a feature that @telegramdesktop necessarily wants to add. Maybe there's some other method planned to deal with this that will be implemented later.

telegramdesktop:
That should be working already for a long time.

Issue #377, opened 2015-01-01
Message Field does not always correctly receive focus on Windows


brynthomas:
if another app is in focus and you click on the text of a History Message inside Telegram, the Message Field does not receive focus. Anything you type gets lost. If you click on the History Message again then Telegram will correctly give the Message Field focus. If you enter the app via clicking on anything except the History Message (say the Cats 'n Dogs background) then the Message Field also correctly gains focus.

On Ubuntu 12.04 if one clicks on the text of a History Message while in another app, the Message Field correctly receives focus.

Seems to be a Windows only problem (but have not tried Mac OS).

telegramdesktop:
Should be fixed in 0.7.8 (the idea was not remove the text selection, if it was, when activating app by click in Windows, only by a second click on already active app), but if nothing is selected the message field should now get focus.

PR #378, opened 2015-01-01
Fixed focus problem when reentering the app by clicking History Message.

brynthomas:
Addresses issue: telegramdesktop/tdesktop#377

On Windows _dragWasInactive is set to true while clicking into the Telegram window. This prevents App::main()->activate() from being called, and instead of refocusing the Message Field, leaves the History Message incorrectly with focus even when nothing has been selected.

Tested on Windows 8.1 and Windows XP and it now appears to work correctly. (tested with selecting portions of items, as well as entire items to ensure focus isn't incorrectly given to the Message Field)

However, Ubuntu worked correctly with the previous code, and so unsure what this change will do on that platform. Do not have an Ubuntu or Mac OS dev environment in place to test there.


auchri:
Should be fixed, thank you :)

Issue #351, opened 2014-12-23
Permanent high CPU usage


mokrates:
Always uses 20 to 35% of my CPU even while doing nothing.
May be related to the constant "playing" of sound through pulseaudio.
(Xubuntu 14.10 i386)

ChrisCA:
With 0.7.7, this problem has been solved in Windows 8.1.
May you check this with Xubuntu?

halega:
Can you fix it for Linux too? PulseAudio uses 8-9% of the CPU permanently + Telegram 2-3% = 10-12% CPU when idle. It's really annoying when you work on a notebook: fans always on.

Telegram 0.7.7
Fedora 21 Workstation
Linux 3.17.8-300.fc21.x86_64
pulseaudio 5.0-25.fc21

paveless:
Linux openSUSE 13.2 i586 KDE.
PulseAudio uses 50% CPU, kmix 25% CPU.
Telegram not usable at all.
Multiple messages in console:
"ALSA lib pcm.c:7897:(snd_pcm_recover) underrun occurred"

marcoklv:
Permanent high CPU usage under Windows 8.1 64-bit Sytem Process usage 15-20 %. If i close
telegram desktop program fan stop and the system process in task monitor going down to normal work. If i start telegram: system process coming up to 15-20 % on cpu usage and fan start on laptop and never stop!

Telegram 0.7.17
XPS Dell Laptop
Windows 8.1 Pro 64bit
Intel® Core™ i7 CPU

mahmoodn:
Same 20%-25% CPU usage on windows 7 (64bit). It is really annoying but unlike linux, there is no precise tool on windows to pinpoint the root cause of this cpu usage. 

Desktop version is 0.7.17

telegramdesktop:
From version 0.8.22.dev pulseaudio (and any other audio) output should stop when there is nothing playing in tdesktop.

mahmoodn:
Still I see 15%-20% (on system process) cpu usage when telegram is running and the system is idle. I tested with 0.8.23.dev

halega:
It's OK with 0.8.24 version now. PulseAudio consumes 0% of CPU while there's no sound playing. Tested in Fedora Workstation 22 and Ubuntu 15.04. Thanks!

mahmoodn:
No. The problems still exists in Windows 7. I don't think this problem is going to be solved!
[...]

Issue #258, opened 2014-11-14
Tdesktop does not allow windows to suspend


santiquetzal:
If I have open Tdesktop my computer does not suspend automatically after x minutes (I have configurated 1 min to do this tests).

If I close Tdesktop my computer suspends automatically after x minutes

telegramdesktop:
Should be fixed, try the latest version 0.7.10.

Issue #229, opened 2014-11-02
Telegram prevents Windows 8.1 automatic sleep


thepillow86:
When telegram is opened, it prevents windows 8.1 from going into sleep mode, because it opens an audiostream. 
I found this through running the powercfg /requests command in CMD. 
Telegram opens a Realtek HD Audio stream:
---------------
C:\Windows\system32>powercfg /requests
DISPLAY:
None.

SYSTEM:
[DRIVER] Realtek High Definition Audio (HDAUDIO\FUNC_01&VEN_10EC&DEV_0668&SUBSYS
_102805FD&REV_1000\4&325a6d8c&0&0001)
An audio stream is currently in use.

AWAYMODE:
None.

EXECUTION:
None.

PERFBOOST:
None.

fefrei:
Let me add that this is extremely annoying.

In addition to the Audio Stream entry shown above, Telegram also causes this entry for me:

    [DRIVER] Legacy Kernel Caller

telegramdesktop:
Should be fixed, try the latest version 0.7.10.

fefrei:
I can confirm this is fixed. The audio stream opens on demand and closes soon after the sound has finished.

terraquian:
this problem was solved in the past but now occurs again (i'm a windows 10 user)

fefrei:
Are you sure this is the same problem? I cannot reproduce it.
Can you post an output of `powercfg /requests`?

terraquian:
certainly, it just try and does not happend. sorry :-(

Issue #191, opened 2014-10-07
use of audiodg.exe CPU


Digitalizer:
When you run the Telegram will begin "AUDIODG.EXE" CPU usage. 

The CPU Clock rise of laptops affected.

telegramdesktop:
From version 0.8.22.dev any audio output should stop when there is nothing playing in tdesktop.

PR #380, opened 2015-01-02
Changed to open audio stream on demand, rather than keeping it permanently open

brynthomas:
**Why:**

Keeping an audio stream open on Windows 8.1 prevents Windows from entering sleep.
Depending on the audio driver it can significantly increase CPU usage, in the form of AUDIODG.EXE on Windows or pulseaudio on Linux.
Depending on the audio driver it can force Windows to use high Timer Resolution, further draining battery.
(I sometimes have Telegram running in a Windows VM on my Mac, and having Telegram open and idle would use an extra 16% of a core. With this change it uses a negligible amount.)

**Addresses the following issues:**

telegramdesktop/tdesktop#229
telegramdesktop/tdesktop#258
telegramdesktop/tdesktop#351
telegramdesktop/tdesktop#176
telegramdesktop/tdesktop#191
telegramdesktop/tdesktop#179

**What changed:**

Split the audio initialisation (audioInit()) into separate methods:

1. Basic OpenAL setup was moved into AudioControl::setupCoreAL()
2. Loading the notification sound's WAV data was moved into AudioControl::loadNotificationWaveData()
3. Setting up the OpenAL buffers to play the notification sound was moved into AudioControl::setupNotificationAL()

This was split so that the notification sound's WAV data could be loaded once, even if the OpenAL specific commands needed to be rerun each time the audio stream was shut down and recreated.

Added a method AudioControl::audioRequired() that is called when the Notification Sound or a Voice Message is played, that resets up the audio stream if it has been shut down. It also resets the audio stream shutdown timer.

Added a timer method which gets executed after a few seconds of inactivity, and if no audio is playing (paused Voice Messages are still counted as playing), it will free buffers, sources, the audio context, and then close the device.

**Known implementation issue:**

If a Voice Message is paused in playback the audio stream will not shut down so as to maintain the context and position of playback. When you finish listening to the Voice Message, then the audio stream will shut down as normal.

Tested on Windows 8.1, Windows XP and Ubuntu 14.04

### DIFF
--- a/Telegram/SourceFiles/audio.cpp
+++ b/Telegram/SourceFiles/audio.cpp
@@ -109,15 +109,15 @@ VoiceMessages::~VoiceMessages() {
 
 	for (int32 i = 0; i < AudioVoiceMsgSimultaneously; ++i) {
 		alSourceStop(_data[i].source);
-		if (alIsBuffer(_data[i].buffers[0])) {
-			alDeleteBuffers(3, _data[i].buffers);
+        if (alIsSource(_data[i].source)) {
+            alDeleteSources(1, &_data[i].source);
+            _data[i].source = 0;
+        }
+        if (alIsBuffer(_data[i].buffers[0])) {
+            alDeleteBuffers(3, _data[i].buffers);
 			for (int32 j = 0; j < 3; ++j) {
 				_data[i].buffers[j] = _data[i].samplesCount[j] = 0;
 			}
-		}
-		if (alIsSource(_data[i].source)) {
-			alDeleteSources(1, &_data[i].source);
-			_data[i].source = 0;
 		}
 	}
 	_faderThread.quit();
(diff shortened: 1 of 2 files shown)


telegramdesktop:
I've tried to fix that by pausing audio thread in Windows mmdevapi audio backend, check out 0.7.7 version. It should handle paused voice messages as well.

ChrisCA:
I've tested 0.7.7 and it works correctly for me now.

halega:
Can you fix it for Linux too? PulseAudio uses 8-9% of the CPU permanently + Telegram 2-3% = 10-12% CPU when idle. It's really annoying when you work on a notebook: fans always on.

Telegram 0.7.7
Fedora 21 Workstation
Linux 3.17.8-300.fc21.x86_64
pulseaudio 5.0-25.fc21

brynthomas:
@telegramdesktop Great stuff, I see it's all working nicely on Windows now and with pauses. Hadn't really seen too much success from other people trying to get OpenAL to play nicely with suspension (think somebody had tried to get it to work with Blender that had a similar issue).

I suppose I was a little hesitant with doing a patch for a dependency like OpenAL just because it's one more thing to look after when updating to a newer version of OpenAL.

Was there a problem with the other backends (say Pulseaudio) and suspension so that the SuspendAndProcessSupported was restricted to Windows/MMDEVAPI?

telegramdesktop:
>Was there a problem with the other backends (say Pulseaudio) and suspension so that the SuspendAndProcessSupported was restricted to Windows/MMDEVAPI?

I've just didn't check that, because in mmdevapi I've made "Suspend" just with a "Stop" call and "Process" just with "Start" call, and I'm afraid different backends could be not ready for such an approach :)

Seems like I should check for pulseaudio as well, because it gives some problems to the users. I've made only mmdevapi because it was the only case where users had the problem.

vitaminace33:
@telegramdesktop I hope you can use this to solve #176.

auchri:
I think @telegramdesktop fixed that, thankt you @brynthomas for the PR :)

PR #383, opened 2015-01-03
Added Korean Translate(delete some commit)

itstake:


telegramdesktop:

PR #390, opened 2015-01-07
Create lang_fa.strings

Jav4dR:
Lang FA


telegramdesktop:

PR #409, opened 2015-01-17
Update lang.h

Nyvki:


telegramdesktop:

PR #410, opened 2015-01-17
Update lang.h

Nyvki:
Please update this file for add Ukrainian language.


telegramdesktop:

Issue #307, opened 2014-12-09
[Enhancement] "Replace emoji" should replace emoji without hyphen dashes :)


ADTC:
I don't add the hyphen to the smiley text `:)` to make it `:-)`. Telegram Desktop currently only replaces the second one, but I would like it to replace the first one too.

I would love to make this change myself but I might break something not knowing anything more than basics of C++ and not knowing even the basics of Objective-C++. But I did some searches, and looks like you only need to copy-paste-reindent a bunch of lines here:

**Add support:**
* [For `B)`](https://github.com/telegramdesktop/tdesktop/blob/7a600e03d0549cdc3f034a6ae86ac6929523d104/Telegram/SourceFiles/gui/emoji_config.cpp#L4364)
* [For `;P`, `;)`](https://github.com/telegramdesktop/tdesktop/blob/7a600e03d0549cdc3f034a6ae86ac6929523d104/Telegram/SourceFiles/gui/emoji_config.cpp#L4417)
* [For `:p`, `:P`, `:D`, `:*`, `:)`, `:(`](https://github.com/telegramdesktop/tdesktop/blob/7a600e03d0549cdc3f034a6ae86ac6929523d104/Telegram/SourceFiles/gui/emoji_config.cpp#L4575)
* [For `8)`](https://github.com/telegramdesktop/tdesktop/blob/7a600e03d0549cdc3f034a6ae86ac6929523d104/Telegram/SourceFiles/gui/emoji_config.cpp#L4654)
* [For `3)`, `3-(`](https://github.com/telegramdesktop/tdesktop/blob/7a600e03d0549cdc3f034a6ae86ac6929523d104/Telegram/SourceFiles/gui/emoji_config.cpp#L4669)

**Related issue:** #116 

***Sidenote:*** *The [findEmoji method](https://github.com/telegramdesktop/tdesktop/blob/7a600e03d0549cdc3f034a6ae86ac6929523d104/Telegram/SourceFiles/gui/emoji_config.cpp#L4319) looks like a maintenance nightmare... isn't there a better way, like some sort of a hash map with key-value pairs?*

joshery420:
:+1: because almost every other chat program I've used that does this doesn't require the hyphen nose.

Also, I totally agree that a hash/dictionary of some kind would probably be a much better way to implement this.

Sidenote: The ability to choose whether we see the iOS or Android emoji set would be awesome, but I'm sure that's extremely low priority. :wink:

ADTC:
[...]

PR #454, opened 2015-01-27
Updated genemoji to support short version smilies

bjaskj:


drasive:
Nice, I wanted to do the same thing

nmat:
+1

joshery420:
Looks like at least a partial (not sure it covers 100% of cases) answer to issue #307 
Thanks @bjaskj!

joshery420:
@bjaskj Curious, is there a specific reason why you didn't add short versions for things like 8-) and B-)?

bjaskj:
I added those that I missed. Sorry if I overlooked something others use and miss :)

rastersoft:
Please, please, please!!!!!! Merge this ASAP!!!!!!!

ADTC:
@telegramdesktop And we're not merging this yet because.... ?

@bjaskj Thanks for this! Very nice. Regarding the 8-) and B-) (etc etc) you can add them and make a second commit to the same branch. It will automatically add to this pull request.

telegramdesktop:
Perhaps some day this replacement array will be configured by user, but right now it is exactly the way it meant to be. I personally use :) and don't want it to be converted automatically and use "xD" and do want it to be converted. Sorry, but that stays the way it is.

I've added [new contributing policy](https://github.com/telegramdesktop/tdesktop/blob/master/CONTRIBUTING.md), so it is clearly stated now, that no features will be accepted in format of pull requests, only bug fixes and code improvements.

drasive:
The first point is subjective and I understand if you prefer to not have short versions of smileys,
but I personally feel a bit disappointed about not accepting any features from the community. I would imagine that some features of the desktop client wouldn't necessarily have a strong impact on the Telegram project as a whole.

ADTC:
Frankly, your response is very disappointing. We all have preferences, likes and dislikes. We only ask that you provide to us the ability to configure and customize as per each of our preferences but you shot us all down with a response that you will dictate and enforce your personal preference on everyone else.

I hope you will reconsider. At the very least, a setting to enable or disable short smiley conversion would be adequate.

telegramdesktop:
@drasive some times features can be accepted - when they do not conflict at all with our view of the desired progress. But it is so unpredictable from the side point of view, that much often case will be disappointment from declining of already prepared pull request with new features.

@ADTC I don't say that there won't be customization of emoji replacements ever, I say that they won't be changed in this pull request format.

Issue #438, opened 2015-01-23
Support for default file browser on Linux


rumatoest:
Telegram should launch default file manager when to show downloaded files in folder.
I have Kubuntu 14.10 
My default file manager is Dolphin, but I also have nautilus installed.
Telegram opening downloads in folder via Nautilus - this is not correct

bil-elmoussaoui:
You should use `xdg-open` instead of `nautilus`, hope this can be fixed in the next version!

stek29:
Isn't this fixed?

PR #478, opened 2015-02-10
Switches show in folder to use user's preferred file manager

tushar-suresh:
xdg-open is the preferred way to do this.
http://linux.die.net/man/1/xdg-open

Addresses  #438


tushar-suresh:
I used my work email for this. I'll issue another pull request with my personal email.

PR #479, opened 2015-02-10
Switches show in folder to use user's preferred file manager

tushar-suresh:
xdg-open is the preferred way to do this.
http://linux.die.net/man/1/xdg-open

Addresses #438


tushar-suresh:
@telegramdesktop could you review this?

tushar-suresh:
@telegramdesktop did you get a chance to look at this and include it in the next release?

joshery420:
+1 for using the user's default file manager instead of hardcoding a specific one. I see no reason why this shouldn't be merged.

telegramdesktop:
I've added some escaping in that line, so it can't be merged like that, but I will replace "nautilus" with "xdg-open", thanks for the report.

PR #519, opened 2015-02-25
Added complete Slovak localization

WURMiL:
Hello,
I've added complete Slovak translation of all strings. I hope it will be useful ;)
Please review.
Thank you


robin0van0der0v:
@WURMiL You should translate it to Slovak here: https://www.transifex.com/projects/p/telegram-desktop/
If it is translated there, it can be much easier added to Telegram Desktop. :wink:

WURMiL:
@robin0van0der0v Oh, thank you for the hint :-) I didn't know, that something like Transifex service exist at all. ;-)

telegramdesktop:

Issue #557, opened 2015-03-13
Non-ascii passes are not recognized in linux


Yanpas:
I have path to my app /home/yanpas/Документы/Дистрибутивы/Telegram/, but in the launch file (telegramdesktop.desktop) it is shown as
 home/yanpas/ÐÐ¾ÐºÑÐ¼ÐµÐ½ÑÑ/ÐÐ¸ÑÑÑÐ¸Ð±ÑÑÐ¸Ð²Ñ/Telegram so it is unable to launch it from menu.

Vdragon:
I can confirm this issue.  I installed Telegram to /home/Vdragon/軟體/Telegram but the Exec field in the generated freedesktop.org desktop entry is "/home/Vdragon/è»é«/Telegram/Telegram -- %u" instead.

Recommended rename this issue to "Incorrect-path freedesktop.org desktop entry installed when Telegram launched in a path contains non-ASCII characters"

Vdragon:
AFAICS the problem is at https://github.com/telegramdesktop/tdesktop/blob/ad04bf77d1c95a9f53dc9dbf7d12b8e12746e759/Telegram/SourceFiles/pspecific_linux.cpp#L1658

It shouldn't use toLocal8Bit() method to convert the string, maybe toUtf8() instead.

PR #567, opened 2015-03-15
pspecific_linux.cpp: replace all toLocal8Bit() call to toUTF8(), fix I18N and #557

Vdragon:
Currently Telegram uses toLocal8Bit() in many GNU/Linux operations such as
generating command line, directory path and file path, this will cause
issues on non-ASCII environments like garbled path in Exec field of
freedesktop desktop entry(issue #557).

This commit simply replace all toLocal8Bit() calls to toUTF8().

WARNING: This patch is by a person with little knowledge on Qt
developing and is also not tested at all, please review before merging.

Reported-by: Yanpas <https://github.com/Yanpas>
Signed-off-by: Ｖ字龍(Vdragon) <Vdragon.Taiwan@gmail.com>

### DIFF
--- a/Telegram/SourceFiles/pspecific_linux.cpp
+++ b/Telegram/SourceFiles/pspecific_linux.cpp
@@ -1606,7 +1606,7 @@ void psOpenFile(const QString &name, bool openWith) {
 
 void psShowInFolder(const QString &name) {
     App::wnd()->layerHidden();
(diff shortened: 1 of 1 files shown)


telegramdesktop:
Fixed in last versions together with some escaping of command line arguments, so can't merge that, but thanks for the report.

PR #605, opened 2015-03-27
Changed the German translation of typing to be more common

SphinxC0re:
Because people aren't always on a keyboard. Also it should be the same translation as on the mobile version.


telegramdesktop:

SphinxC0re:
Could you handle this change anyway? :pray:

PR #613, opened 2015-03-29
Update QTCREATOR.md

Streppel:
updated correct path to patch, which is actually under /Telegram and not on /tdesktop


telegramdesktop:
Thanks.

PR #682, opened 2015-04-29
[linux] Telegram puts a 'fontconfig' folder in the user's home on every start.

agusdallalba:
Deleting the folder only makes it show up again the next time you launch Telegram from the desktop shortcut.

The folder shows up in the app's working directory, so included is a pull request that changes the working directory to where Telegram is installed. Unfortunately this will only work for new installs.


telegramdesktop:
I now change working directory to data files directory like ~/.TelegramDesktop (in the last versions).

PR #723, opened 2015-05-16
Update lang_es.strings (minor proposal)

kant:
Semantic issues


eduardosm:
I would keep "icono" rather than changing it to "ícono".

telegramdesktop:

PR #746, opened 2015-05-22
Remove lines in genemoji.cpp that prevents building on non-macs

luvemil:
The two line removed seem unnecessary. I have successfully built and run the software in Debian with this configuration.


telegramdesktop:
This software requires Mac OS colored emoji font from OS X system to run and generated proper sprites, so there is no point in running it on other OS, you won't get valid sprite images.

PR #774, opened 2015-05-31
[Feature Implementation] - Implemented right click context menu for dialog widget (and ability to specify notification tone)

kireledan:


RadialAdmin:
+1 
Two month wait this commit

telegramdesktop:
I've added [new contributing policy](https://github.com/telegramdesktop/tdesktop/blob/master/CONTRIBUTING.md), so it is clearly stated now, that no features will be accepted in format of pull requests, only bug fixes and code improvements. Sorry. Some of the features will be made in the future, some will not, but not those together and not in the current Settings design.

PR #819, opened 2015-06-11
Update README.md with a new supported OS

ner0x652:
It also work on Fedora 22. Tested!


grandcat:
It also perfectly works with Windows 10 (Home).

PR #365, opened 2014-12-27
added Polish translation file

Morgaroth:

PR #837, opened 2015-06-17
Added Polish (pl_PL) translation.

zal-any:
Hi!

I thought that it would be a good idea to add Polish translation to Telegram Desktop. This is initial translation, it needs some proof reading and polishing, but it should be a good start. In the future, I also want to provide such translation for Telegram for Android.

Best regards,
Karol Zalewski


zal-any:
Nevermind.

After some hours of fighting with this I realized two things:
1) I need to modify more files to achieve working translation (lang.h, telegram_linux.qrc and Telegram.pro).
2) Someone created pull request for Polish translation few months ago (see https://github.com/telegramdesktop/tdesktop/pull/365).

Sorry for bothering you.

telegramdesktop:

PR #846, opened 2015-06-23
Replace two dots with ellipsis character (U+2026)

headcrab-in-my-room:
Two-dot ellipsis are used in mathematics, but not in ordinary text.


aluminiumgeek:
AFAIK two dots instead of ellipsis is some kind of "corporate identity"

auchri:
Please resolve the conflicts

telegramdesktop:
Yeah, we just like two dots, sorry ¯\_(ツ)_/¯

PR #920, opened 2015-07-17
Allow to disable the updater using a marco

eduardosm:
Adding "DEFINES += TDESKTOP_DISABLE_AUTOUPDATE" to Telegram.pro would disable the updater. This eases creating packages for Linux distributions, where updates are handled by the corresponding package manager.


auchri:
@eduardosm Please create a new pull request with only your commit.

Consider to rebase the commit and don't add all the other commits to your pull request.

PR #1055, opened 2015-09-03
Allow to disable 'register custom scheme' functions using a macro.

eduardosm:
Adding "DEFINES += TDESKTOP_DISABLE_REGISTER_CUSTOM_SCHEME" to Telegram.pro would disable these functions. This can be useful when the application is distributed using an installer or a package manager that can apply these settings system-globaly.

### DIFF
--- a/Telegram/SourceFiles/pspecific_linux.cpp
+++ b/Telegram/SourceFiles/pspecific_linux.cpp
@@ -1165,6 +1165,7 @@ namespace {
 }
 
 void psRegisterCustomScheme() {
+    #ifndef TDESKTOP_DISABLE_REGISTER_CUSTOM_SCHEME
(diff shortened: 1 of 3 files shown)


auchri:
Please update the description :)

eduardosm:
Oops! That's why I have to be careful when doing copy-paste.
Updated.

telegramdesktop:
I've added [new contributing policy](https://github.com/telegramdesktop/tdesktop/blob/master/CONTRIBUTING.md), so you'll need to make a pull request with the new "Signed-off-by:" signature in the commit message, like it is described in [sign your work](https://github.com/telegramdesktop/tdesktop/blob/master/CONTRIBUTING.md#sign-your-work) section.

PR #1056, opened 2015-09-03
Commented string, which creates an error while compiling on non MacOS OS

LibertyPaul:
I wasnt able to build MetaEmoji on my Linux system until I've commented this error macro, which seems to be wrong.
But I'm newbie, this is my first commit into not my project, and I dont't know all about how things get done here, just want to help.
If my commit is bad, comment below, I want to know your oppinion.

### DIFF
--- a/QTCREATOR.md
+++ b/QTCREATOR.md
@@ -82,6 +82,7 @@ In Terminal go to **/home/user/TBuild/Libraries** and run
 
     sudo apt-get install xutils-dev bison python-xcbgen
     git clone https://github.com/xkbcommon/libxkbcommon.git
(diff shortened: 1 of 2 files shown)


auchri:
So you build MetaEmoji  successfully on linux?

telegramdesktop:
There is no need to build MetaEmoji project for building Telegram. I'll remove that line from QTCREATOR.md instructions, so that you won't even try. Emoji sprites are generated only once and are used in all other builds and they are available in the repository already. They can be built successfully only in Mac OS X.

LibertyPaul:
@auchri yes

PR #1178, opened 2015-10-13
Resolving "failed retrieving file" error

Alireza-Mim:
The command "pacman -S mingw64/mingw-w64-x86_64-opus" gives an error and based on the comment on this ticket: http://sourceforge.net/p/msys2/tickets/186/ , we should sync the package database first.
It worked for me.


telegramdesktop:
Thanks for your fix.

PR #1221, opened 2015-10-23
Do not depend on QPixmap and QGuiApplication in MetaStyle.

eduardosm:


telegramdesktop:
Does it actually work? I remember, that something about working with images pulled the QGuiApplication dependency so I introduced it. But perhaps it is not needed after all?

eduardosm:
I tested it on Linux and it works. I had read somewere that QImage worked without QGuiApplication as long as you don't use QPainter::drawText(). The purpose of this patch is to avoid depending on a X11 connection to run MetaStyle.

eduardosm:
Maybe using '-platform offscreen' would be a better idea?

PR #1272, opened 2015-11-05
Add c++11 config flag to Telegram.pro

eduardosm:
It is needed to use UTF-16 string literals (u"string"), which are generated by MetaLang.


telegramdesktop:
So it does not work right now? The lang_auto.cpp does not get any problems when I build it right now..

PR #1299, opened 2015-11-15
Initialise _firstLoadRequest to prevent jumps from depending on uninitialised memory.

michaelsproul:
When running Telegram under Valgrind, I found that there were jumps depending on the value
of _firstLoadRequest (in historywidget.cpp) before it was initialised.

It seems like 0 is used as the default value for this variable elsewhere.


telegramdesktop:
It is really a bug, thank you for your fix, but it can't be a reason for anything: before even first usage _firstLoadRequest it is cleared in "clearAllLoadRequests();" call. All it could do was canceling some MTP request before history load (probability of that was very low, but it was possible theoretically), and after that it just was getting zero value in "clearAllLoadRequests();" call in "HistoryWidget::showHistory".

PR #1346, opened 2015-11-30
Do not depend on MFC

OZ1:
Telegram.rc depends on afxres.h. This file is installed with MFC. For the project to compile it is sufficient to use winres.h
https://connect.microsoft.com/VisualStudio/feedback/details/759972/rc-files-include-mfc-headers-even-when-mfc-is-not-installed

PR #1374, opened 2015-12-10
Do not depend on MFC #2

OZ1:
'Found out that there is another .rc file that has an MFC dependency. Updater.rc also depends on MFC. It prevents 0.9.15 from compiling successfuly without having MFC installed.
same as https://github.com/telegramdesktop/tdesktop/pull/1346 for Telegram.rc
Also I'd liked to notice that the version in Updater.rc is set to 0.9.0.0. Don't know whether it's by design or just by oversight.


OZ1:
Don't know how it managed to compile previous time

auchri:
What did you changed in `Telegram.rc`?

PR #1394, opened 2015-12-14
Fix a build error when TDESKTOP_DISABLE_AUTOUPDATE preprocessor defined.

heejune:
If the 'TDESKTOP_DISABLE_AUTOUPDATE' defined, a build is failing because empty moc_autoupdater.cpp is being generated and the msvc compiler will complain about 'stdafx.h not found'. In addition, aboutbox.cpp also refers to countBetaVersionSignature() which will not be included when TDESKTOP_DISABLE_AUTOUPDATE defined.

Added a dummy UpdateDownloader inherits QObject when TDESKTOP_DISABLE_AUTOUPDATE not defined, and moved the location of countBetaVersionSignature(uint64 version) to outside of #ifndef/endif scope.

Fixed & Tested on dev branch w/ VS2015.

### DIFF
--- a/Telegram/SourceFiles/autoupdater.cpp
+++ b/Telegram/SourceFiles/autoupdater.cpp
@@ -569,6 +569,8 @@ bool checkReadyUpdate() {
 	return true;
 }
 
+#endif
+
 QString countBetaVersionSignature(uint64 version) { // duplicated in packer.cpp
 	if (cBetaPrivateKey().isEmpty()) {
 		LOG(("Error: Trying to count beta version signature without beta private key!"));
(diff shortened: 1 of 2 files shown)


auchri:
Thanks for the PR! :+1: 

It's enough to include the Signed-off-by in the commit message, it's not necessary to include it also into the description of the PR :)

telegramdesktop:
@heejune now "QString countBetaVersionSignature(uint64 version);" is repeated twice in autoupdater.h at the end.

heejune:
@telegramdesktop Right... Thanks for the comment. I thought and even double checked that I moved out the both function definitions to the outside of #ifndef/#endif scope before the commit, but it seemed that I have committed the wrong repo after testing. I'll close this and resubmit again.

auchri:
Please squash the commits: http://stackoverflow.com/a/5201642/4752807

heejune:
@auchri Thanks for pointing that out.

PR #1409, opened 2015-12-16
Create Localizable.strings

mamadoo:
Persian language support


auchri:
What's the sense of this change?

telegramdesktop:
The translations are not managed through github, sorry. They are on Transifex and currently the supported languages list is freezed.

PR #1436, opened 2015-12-25
Build instructions for Visual Studio (Qt 5.5.1)

alejodosr:
When executing this command:

perl init-repository --module-subset=qtbase,qtimageformats

In my case, just only one of those modules was taken into account, so that a had to execute it separately with every module. Maybe, it a note shall be included in the instructions or directly splitting the command.


auchri:
Thanks for the PR!

Where do you add the note?

For more information about pull requests, see https://help.github.com/articles/using-pull-requests/

PR #1648, opened 2016-02-19
Include some reference XDG files

hobarrera:
Include a .desktop file (for app menu in xdg-complaint desktops), and tg.protocol (for having xdg-complaint apps pick up the tg: protocol).

The intention of including these files is for package maintainers to re-use them, instead of each or them re-writing (or re-bundling for somewhere else) these files.

### DIFF
--- a/lib/xdg/telegramdesktop.desktop
+++ b/lib/xdg/telegramdesktop.desktop
@@ -1,0 +1,12 @@
+[Desktop Entry]
+Encoding=UTF-8
+Version=1.0
+Name=Telegram Desktop
+Comment=Official desktop version of Telegram messaging app
+Exec=/usr/bin/telegram-desktop -- %u
+Icon=telegram-desktop
+Terminal=false
+StartupWMClass=Telegram
+Type=Application
+Categories=Network;
+MimeType=application/x-xdg-protocol-tg;x-scheme-handler/tg;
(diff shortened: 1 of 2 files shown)


auchri:
Hi,
thanks for the pull request!

Please read our [contributing policy](https://github.com/telegramdesktop/tdesktop/blob/master/CONTRIBUTING.md). You'll need to make a pull request with the "Signed-off-by:" signature being the last line of your commit message, like it is described in [sign your work](https://github.com/telegramdesktop/tdesktop/blob/master/CONTRIBUTING.md#sign-your-work) section. That will grant your work into the public domain.

hobarrera:
Done :)

telegramdesktop:
The distributed app currently contains of two binaries, Telegram + Updater. There is no sense in just adding such files to the repository, they won't affect users that download the app in any way. But such files are created by the app itself and are placed (at least should be) in the places they belong, you can see it in the source code [here](https://github.com/telegramdesktop/tdesktop/blob/master/Telegram/SourceFiles/pspecific_linux.cpp#L1301) and [here](https://github.com/telegramdesktop/tdesktop/blob/master/Telegram/SourceFiles/pspecific_linux.cpp#L1348).

hobarrera:
The intention is to make them available for downstream packagers from one central location (rather than each one having to find/bundle it from some external location). They would effectively provide no use for the "build a re-distributable binary" scenario.

These files are created by telegram upon first run, but:

1. They're per user; other users can't find it
2. Users won't have a first-run if they can't see an icon (assuming average user, and not someone how runs via a terminal).

Issue #1615, opened 2016-02-15
"Disable Notifications" from tray does not disable sound


Royal2000H:
Title is pretty self-explanatory, but clicking on "Disable Notifications" from the tray does not disable the notification sound, only the popup. One would expect it to do both.

Windows 8.1. Telegram Desktop 0.9.18 stable. Upgraded to 0.9.23 dev - same issue.

theus:
The only way that works with me is disable the notification in "Info" tab at conversation. But, I believe that "Disable Notifications" in tray should disable all notifications of all conversations.

tdesktop-closer:
Hey there!

We're automatically closing this issue since there was no activity in this issue since 360 days ago. We therefore assume that the user has lost interest or resolved the problem on their own. Closed issues that remain inactive for a long period may get automatically locked.

Don't worry though; if this is in error, let us know with a comment and we'll be happy to reopen the issue.

Thanks!

<sub>(Please note that this is an [automated](https://github.com/auchri/github_auto_closer) comment.)</sub>

PR #1654, opened 2016-02-19
disabling notifications now disables sound too

Royal2000H:
In reference to issue #1615.

Please note, this was not tested as I do not currently have a build environment set up. It should be extremely quick and easy to test if the build environment is already configured.

### DIFF
--- a/Telegram/SourceFiles/settingswidget.cpp
+++ b/Telegram/SourceFiles/settingswidget.cpp
@@ -237,6 +237,7 @@ SettingsInner::SettingsInner(SettingsWidget *parent) : TWidget(parent)
 	// notifications
 	_senderName.setDisabled(!_desktopNotify.checked());
 	_messagePreview.setDisabled(_senderName.disabled() || !_senderName.checked());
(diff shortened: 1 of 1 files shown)


telegramdesktop:
I don't like it that way :( If you go to settings and disable just the popup notifications it will disable sound as well -- but they are different checkboxes, you can disable them separately and it is not looking good, when changing one checkbox automatically checks another.

Royal2000H:
When you disable the notifications, it also disables "Sender's name" and "Message Preview" so it's not that different for the user. It only makes sense for the user to have a way to disable all notifications in one click.

If you disagree with this solution, how about a solution that clicking "Disable Notifications" in the tray would also disable Sound? So in the Settings menu, disabling Desktop notifications won't do anything but in the tray, disabling notifications will also disable sound.

telegramdesktop:
@Royal2000H yes, this could be done, but in more complicated way. If you click in tray Disable notifications and then Enable notifications they should return to exact same state as they were before. So the sound settings should be saved in some way just for this switch in the tray icon menu. I'll think about that.

john-preston:
This finally should work :)

Issue #1783, opened 2016-03-15
An option to copy user's first name and last name


Kolahzary:
The problem is when one of my friends writes his name in Chinese and I want to search his name in google in order to find what is his real name. It's impossible! because simply I can't read or write Chinese characters and Telegram doesn't have **An option to copy user's first name and last name**!
Regards,

telegramdesktop:
@Kolahzary You can select his message (right click -> Select Message) and then Ctrl+C, his message text will be copied together with his name and date the message was received.

DaFri-Nochiterov:
![image](https://cloud.githubusercontent.com/assets/10401817/13810738/836692ce-eb9b-11e5-81af-b878577cbddd.png)

Can you add this to menu, when right click on username, please? :blush:

> * Copy name
> * Copy phone number
> * Copy username

k4ml:
Found this issue when I want to copy my contact's phone number, but I can't select it and use Command + C to copy it. Turn out you can copy it using context menu like in screenshot above. What I can say is that the feature is not easily discoverable. It's common when people want to copy something, they will select (highlight) it first, and then open up the context menu, or immediately using Ctrl/Cmd + C to copy it.

stek29:
@auchri It was added in last version (0.9.34)

auchri:
@stek29 oh yes, thank you!

stek29:
@auchri Np

PR #1797, opened 2016-03-18
Add "Copy name" option (Requested in #1783)

stek29:
Also add `lng_profile_copy_fullname` to lang files

PS: Please check other languages, translation may be inaccurate.


stek29:
Some screenshots:
Name only
Name and Username
Name and Number
Name, Number and Username

Also when `last_name` is empty it copies `First `, not `First`, but it's `lng_full_name` issue.

john-preston:
@stek29 there should not be any translations here, they will be done by translators on Transifex. Just remove all changes in all "lang_...strings" files.

stek29:
@john-preston Oh, i've broken my repository, so I had to [duplicate pull request](../pull/1801).

PR #1918, opened 2016-04-18
Fixed inverted checkbox for supergroup members notification about pinned message

VBelozyorov:
It seems that `Notify all members` checkbox was inverted.

I'm talking about this one:
notify

If I set it to `unchecked`, other supergroup members say they recieved notification, but when I leave it `checked` — they hadn't recieved anything. 

As you can see in source, `silent` flag was used when checkbox was checked, but text of the checkbox is about notification, not about silence :)
    
Signed-off-by: Vladimir Belozyorov <VBelozyorov@users.noreply.github.com> (github: vbelozyorov)


VBelozyorov:
oops, bad commit message via github editor

PR #1954, opened 2016-04-24
Add zooming in media viewer on mouse wheel

### LINKED ISSUE #1013: [Feature request] mouse control when viewing images
right btn+scroll = zoom
scroll = next/prev

EXL:
Hello!
In many standard image viewers mouse wheel uses for zoom in/zoom out.
This is more convenient than Ctrl+Plus/Ctrl+Minus/Ctrl+0.

Fix #1013

### DIFF
--- a/Telegram/SourceFiles/mediaview.cpp
+++ b/Telegram/SourceFiles/mediaview.cpp
@@ -520,6 +520,74 @@ void MediaView::step_radial(uint64 ms, bool timer) {
 	}
 }
 
+void MediaView::zoomIn() {
+	int32 newZoom = _zoom;
+	if (newZoom == ZoomToScreenLevel) {
+		if (qCeil(_zoomToScreen) <= MaxZoomLevel) {
+			newZoom = qCeil(_zoomToScreen);
+		}
+	} else {
+		if (newZoom < _zoomToScreen && (newZoom + 1 > _zoomToScreen || (_zoomToScreen > MaxZoomLevel && newZoom == MaxZoomLevel))) {
+			newZoom = ZoomToScreenLevel;
+		} else if (newZoom < MaxZoomLevel) {
+			++newZoom;
+		}
+	}
+	zoomUpdate(newZoom);
+}
+
+void MediaView::zoomOut() {
+	int32 newZoom = _zoom;
+	if (newZoom == ZoomToScreenLevel) {
+		if (qFloor(_zoomToScreen) >= -MaxZoomLevel) {
+			newZoom = qFloor(_zoomToScreen);
+		}
+	} else {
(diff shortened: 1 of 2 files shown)


auchri:
I would prefer to zoom only with <kbd>CTRL</kbd> and mouse wheel.

Mouse wheel without the key could be used to switch to the next/last image.

#1013

DRSDavidSoft:
Please add mouse wheel functionality, developers.
Without holding <kbd>CTRL</kbd> to navigate through images, and with holding <kbd>CTRL</kbd> to zoom in/out.

stek29:
Wait, `Qt::ControlModifier` won't work for OS X <kbd>Cmd</kbd>, so you need to check meta too...

auchri:
> Note: On OS X, the ControlModifier value corresponds to the Command keys on the Macintosh keyboard, and the MetaModifier value corresponds to the Control keys. The KeypadModifier value will also be set when an arrow key is pressed as the arrow keys are considered part of the keypad.

http://doc.qt.io/qt-5/qt.html#KeyboardModifier-enum

DRSDavidSoft:
When will this commit be merged into the master branch?

john-preston:
I'm not sure about it.

For example, it is not a good idea for touchpads. At least for OS X ones (where the vertical two finger swipe will be mouse wheel), because vertical two finger swipe has nothing to do with horizontal movement between the images.

Perhaps it is good for a physical mouse with a classic wheel, but maybe even on Windows with some multitouch touchpads it doesn't make sense. On Windows with touchscreen perhaps two finger vertical gestures are as well translated to the mouse wheel events (or not, I don't even have information about all the devices available).

auchri:
So @EXL should check if the `QWheelEvent` comes from a mouse and only then switch to the next image? :)

EXL:
@auchri sorry for the long wait.

This can be checked with using `QWheelEvent::source()` and `Qt::MouseEventNotSynthesized`, right?
http://doc.qt.io/qt-5/qwheelevent.html#source
http://doc.qt.io/qt-5/qt.html#MouseEventSource-enum

I am a little bit confused that method `QWheelEvent::source()` was introduced only with Qt 5.5

auchri:
Thanks!

> This can be checked with using `QWheelEvent::source()` and `Qt::MouseEventNotSynthesized`, right?

Yes, I think so.

Please http://rebaseandsqua.sh :)

EXL:
@auchri 

> Please http://rebaseandsqua.sh

Ha-ha! Nice service, big thanks for sharing, but I know how to use `git rebase -i`.

Done.

EXL:
Fixed conflicts with `setZoomLevel()` function, but don't tested this :(

john-preston:
Thanks!

Issue #2032, opened 2016-05-16
About dialog on dark desktop theme


poofeg:
About dialog on dark desktop theme looks like that:
![telegr-dark-theme](https://cloud.githubusercontent.com/assets/586294/15281524/ed4275b8-1b5d-11e6-894d-24699b6fec6b.png)
### Configuration

**Operating system:** Kubuntu 16.04

**Version of Telegram Desktop:** 0.9.49

john-preston:
@poofeg There is no dark desktop theme in telegram desktop yet, only in some forked applications based on it.

poofeg:
I just want to make the text can be read.

john-preston:
@poofeg Well, it can be read in the official version from https://desktop.telegram.org

poofeg:
@john-preston I think we do not understand each other. I'm using official version of Telegram. And I can't read text in about dialog.

poofeg:
@john-preston I think #2033 can fix this issue.

z3ntu:
@poofeg How did you set a "dark mode" in Telegram desktop?

poofeg:
@z3ntu I set it in system, not in Telegram.

stek29:
And did it work? Can you provide some screenshots?
Is About popup the only one with white background?

poofeg:
The only color from system theme is text color in about dialog. All other colors is redefined.

PR #2033, opened 2016-05-16
Set color of text in about dialog explicitly

poofeg:
Fix issue #2032


john-preston:
There is no "color" field in "textStyle" class, so that will not precompile. But I understand what you're trying to achieve, I didn't see such problem in any environment I've tried, but I'll set the color explicitly for the next version.

PR #2039, opened 2016-05-16
Add Generic user pics (See #1905)

### LINKED ISSUE #1905: Better generic avatar
### Steps to reproduce
1. Feature request
   2.
   3.
### Expected behaviour

How about include initials of contact inside generic avatar like telegram android? Maybe setting or even allow to add one picture to the contact to use has avatar.
### Actual behaviour

You get a generic avatar making searching when fast scrolling more hard.
### Configuration

**Operating system:** Any.

**Version of Telegram Desktop:** Unrelated.
### Logs

Insert logs here (if necessary)

You can type "debugmode" in settings and then see ~/.TelegramDesktop/DebugLogs/log_...txt for log files.
Type "debugmode" in settings again to disable logs.

stek29:
Fixes #1905.

I'm not really sure where `drawInitialsPixmap` and `initialsByName` should be placed and how they should be named, so please correct me if I place or name them wrong.

### Screenshots ###
In dialog
Word Split
Contact
Mention

### Known issues ###

- Some letters like `J` or Arabic/Persian ones isn't really centered _(looks like iOS doesn't center Arabic/Persian too)_ _(looks like manual centering is needed, Qt::alignCenter isn't suitable)_  
JJ  
Arabic
- Emojis and special symbols aren't filtered _(like on Android app)_ _(May be solved by adding `isValidForInitials(char)` method and iterating through `first` and `last` string's until it return's true? iOS app uses this method)_    
Emoji


auchri:
+1827272728292

Nawuko:
Would like to see that :+1:

john-preston:
Currently it was a design decision to leave the empty userpics as they are right now (last time it was discussed when we moved to round userpics in tdesktop). Perhaps some day this will change, but currently empty userpics are displayed the way they are not because this feature was not implemented, it was not implemented intentionally.

stek29:
@john-preston So I can close it?

stek29:
@john-preston all official apps have them, and only tdesktop has that old style weird pics...

stek29:
Well, I think I'd just leave this in [my fork](/stek29/tdesktop-plus).

stek29:
Well, I think I'd just leave this in [my fork](/stek29/tdesktop-plus).

PR #2043, opened 2016-05-17
Fix travis

auchri:


stek29:
Yaaaay!

auchri:
If you don't mind :)

stek29:
Hm, but it's based on dev, while this patches master...

Issue #1908, opened 2016-04-16
Make tray icon themeable


andia89:
### Steps to reproduce
1. Open telegram on Linux
2. Enable tray icon
### Expected behaviour

Icon from generic place like `/usr/share/icons/hicolor` is used instead of a hardcoded one, so that different icon themes can provide tray icons for telegram instead of a hardcoded png. 
### Actual behaviour

An icon that is hardcoded in the binary is used, with absolutely no way to change apart from building telegramDesktop from source with changed png files...
### Configuration

Ubuntu 15.10 with appindicator installed

Telegram version 0.9.33

auchri:
Shouldn't #1431 fix this issue?

wa4557:
I don't think so because this is only for the app icon (the icon that shows up in the unity launcher for example), whereas the tray icon is still hardcoded.

If I made a pull request, is this something you would consider merging?

auchri:
Duplicate #1199 

> If I made a pull request, is this something you would consider merging?

Yes, a PR would be great 👍

john-preston:
What will happen with the coloured unread counters that are placed upon the icon (currently red or grey, depending on the muted status of the conversations with a positive unread counter)? They will be placed that way above an arbitrary tray icon?

wa4557:
I suppose that's up to the icon theme to decide. If they  ship icons for all different states, they should show up

john-preston:
@wa4557 I just don't understand (and afraid a bit) how this is going to look in the pull request — the counter is added uniformly there using internal styles.

wa4557:
Is it? As far as I understand the code there are a lot of different png files for all different counter values. So for me there's not really a difference if this is done internally (this will be the fallback oc), or via an icon theme 

I could be wrong though, and misunderstand  the code...

DavidVentura:
Any news on this? Plasma uses the theme icon but the status bar still shows the original blue icon

![teleg1](https://cloud.githubusercontent.com/assets/3650670/15986265/f3a01830-2fd9-11e6-8690-54ac71a42b37.png)
![teleg2](https://cloud.githubusercontent.com/assets/3650670/15986269/04ec1634-2fda-11e6-9474-71dd16164f6f.png)

auchri:
@wa4557 Did you make some progress on this issue?

wa4557:
[...]

PR #2137, opened 2016-06-12
Partially addresses #1908

wa4557:
This allows changing the tray icon, like it was before, by changing icons in ~/.TelegramDesktop/tdata/ticons. Cf. #1908

### DIFF
--- a/Telegram/SourceFiles/pspecific_linux.cpp
+++ b/Telegram/SourceFiles/pspecific_linux.cpp
@@ -152,6 +152,9 @@ namespace {
     typedef GtkStatusIcon* (*f_gtk_status_icon_new_from_pixbuf)(GdkPixbuf *pixbuf);
     f_gtk_status_icon_new_from_pixbuf ps_gtk_status_icon_new_from_pixbuf = 0;
 
+    typedef GtkStatusIcon* (*f_gtk_status_icon_new_from_file)(const gchar *icon_name);
+    f_gtk_status_icon_new_from_file ps_gtk_status_icon_new_from_file = 0;
+
     typedef void (*f_gtk_status_icon_set_from_pixbuf)(GtkStatusIcon *status_icon, GdkPixbuf *pixbuf);
     f_gtk_status_icon_set_from_pixbuf ps_gtk_status_icon_set_from_pixbuf = 0;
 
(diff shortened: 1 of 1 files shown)


auchri:
Hi,
thanks for this pull request :)

Please read our [contributing policy](https://github.com/telegramdesktop/tdesktop/blob/master/.github/CONTRIBUTING.md). You'll need to make a pull request with the "Signed-off-by:" signature being the last line of your commit message, like it is described in [sign your work](https://github.com/telegramdesktop/tdesktop/blob/master/.github/CONTRIBUTING.md#sign-your-work) section. That will grant your work into the public domain.

(See travis build)

john-preston:
Unfortunately I've changed the way (the place where) the libraries are loaded, new version is available in the dev branch -- so this pull request won't merge. But I think I can add this by myself.

wa4557:
Fine with me, as long as the feature gets added :)

Telegram/SourceFiles/pspecific_linux.cpp
```diff
@@ -641,18 +653,35 @@ void PsMainWindow::psUpdateCounter() {
...
-            ps_gtk_status_icon_set_from_pixbuf(_trayIcon, _trayPixbuf);
+            QFileInfo f(_trayIconImageFile());
+            if (f.exists()) {
+                QByteArray path = QFile::encodeName(f.absoluteFilePath());
+                _trayIcon = ps_gtk_status_icon_new_from_file(path.constData());
+                }
+            else{
+                loadPixbuf(_trayIconImageGen());
+                _trayIcon = ps_gtk_status_icon_new_from_pixbuf(_trayPixbuf);
+            }
```
john-preston: You've replaced a ..set_from_pixbuf(_trayIcon) call with a _trayIcon = ...new_from_...(), which means you create each time a new icon, just forgetting a pointer to the old one -- this should cause memory leaks and other bugs (for example, you don't connect callbacks, like it is done on 792 (old) / 820 (new) line. I think you need not only ..._new_from_file(), but both that one and ..._set_from_file().

john-preston:
You've replaced a ..set_from_pixbuf(\_trayIcon) call with a \_trayIcon = ...new_from_...(), which means you create each time a new icon, just forgetting a pointer to the old one -- this should cause memory leaks and other bugs (for example, you don't connect callbacks, like it is done on 792 (old) / 820 (new) line. I think you need not only ..._new_from_file(), but both that one and ..._set_from_file().

john-preston:
I've added such code to the dev branch in the new files.

PR #2163, opened 2016-06-17
Tuning of locating of winRT functions

kleuter:
Both WindowsCreateStringReference & WindowsDeleteString are also
in
COMBASE.DLL, using of "api-ms-win-core-winrt-string-l1-1-0.dll"
looks
not nice plus the dll name can theoretically be
changed


john-preston:
Unfortunately I've changed the way (the place where) the libraries are loaded, new version is available in the dev branch -- so this pull request won't merge. But I think I can add this by myself.

john-preston:
I've loaded them from combase in dev branch in the new windows_dlls.cpp module. Thanks!

PR #2165, opened 2016-06-18
Repair of parallel build; reconstruction of release build of satellites

mymedia2:
Fixes a race when `make` is running in multiprocess mode with the flag `-j`, but files that are produced by `codegen_style`, have not exist yet. A build fails if the compilation of `Telegram` are started when `codegen_style` has not been finished.

Also the automatically generated code will be created in the release build.


john-preston:
Wow! Thanks for clearing the way those qmake things work.

PR #2181, opened 2016-06-21
strip out .desktop file generation (and xdg uri association)

### LINKED ISSUE #306: [Linux] Stop creating ".desktop" file automatically!
Because I already made a custom ".desktop" file that uses an icon from my current icon theme. But when I launch the app, it just replace the file with the one that uses the default icon from the app. That's annoyying.

Could you just give the user an option?

AndydeCleyre:
The two elements I've stripped out (runtime-generated `.desktop` launcher and xdg mime URI association) are in my view at best redundant functions and at worst meddling gremlins in the machinery.

Packagers who previously relied on the generated-at-launch `.desktop` file should instead use the one provided in `lib/xdg/`, modifying if necessary.

Fixes #306

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ env:
   - BUILD_VERSION="disable_register_custom_scheme"
   - BUILD_VERSION="disable_crash_reports"
   - BUILD_VERSION="disable_network_proxy"
+  - BUILD_VERSION="disable_desktop_file_generation"
 
 arch:
   packages:
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -79,6 +79,10 @@ prepare() {
 		options+="\nDEFINES += TDESKTOP_DISABLE_NETWORK_PROXY"
 	fi
 
+	if [[ $BUILD_VERSION == *"disable_desktop_file_generation"* ]]; then
+		options+="\nDEFINES += TDESKTOP_DISABLE_DESKTOP_FILE_GENERATION"
+	fi
+
 	options+='\nINCLUDEPATH += "/usr/lib/glib-2.0/include"'
 	options+='\nINCLUDEPATH += "/usr/lib/gtk-2.0/include"'
 	options+='\nINCLUDEPATH += "/usr/include/opus"'
--- a/Telegram/SourceFiles/pspecific_linux.cpp
+++ b/Telegram/SourceFiles/pspecific_linux.cpp
@@ -434,6 +434,7 @@ void psRegisterCustomScheme() {
     QString home(_psHomeDir());
     if (home.isEmpty() || cBetaVersion()) return; // don't update desktop file for beta version
 
+    #ifndef TDESKTOP_DISABLE_DESKTOP_FILE_GENERATION
     DEBUG_LOG(("App Info: placing .desktop file"));
     if (QDir(home + qsl(".local/")).exists()) {
         QString apps = home + qsl(".local/share/applications/");
@@ -480,6 +481,7 @@ void psRegisterCustomScheme() {
             LOG(("App Error: Could not open '%1' for write").arg(file));
         }
     }
+    #endif // TDESKTOP_DISABLE_DESKTOP_FILE_GENERATION
 
     DEBUG_LOG(("App Info: registerting for Gnome"));
     if (_psRunCommand("gconftool-2 -t string -s /desktop/gnome/url-handlers/tg/command " + escapeShell(escapeShell(QFile::encodeName(cExeDir() + cExeName())) + " -- %s"))) {


auchri:
Hi,
thanks for the pull request!

Please read our [contributing policy](https://github.com/telegramdesktop/tdesktop/blob/master/.github/CONTRIBUTING.md). You'll need to make a pull request with the "Signed-off-by:" signature being the last line of your commit message, like it is described in [sign your work](https://github.com/telegramdesktop/tdesktop/blob/master/.github/CONTRIBUTING.md#sign-your-work) section. That will grant your work into the public domain.

(See travis build)

john-preston:
For the packagers who will build their version of the binary you can opt-out this piece of code by #ifdef TDESKTOP_DISABLE_DESKTOP_FILE_GENERATION (or similar) directives. Currently that is done for autoupdate code and some other. There is even a TDESKTOP_DISABLE_REGISTER_CUSTOM_SCHEME key that is used by packagers, that disables some special actions on-launch, including .desktop file generation.

The main code branch is used to build a version that is distributed through https://desktop.telegram.org/ in a simple compressed way and it should work just by download-unpack (like it does right now).

This is the only version that is currently officially offered by Telegram, anyone else can build their own versions with or without .desktop file generation and distribute through any package manager they prefer.

AndydeCleyre:
@john-preston Can you point to anything regarding this `TDESKTOP_DISABLE_DESKTOP_FILE_GENERATION` ? I don't see any evidence of it.

The other one you mention, `TDESKTOP_DISABLE_REGISTER_CUSTOM_SCHEME`, seems to prevent extra things that we don't want or need to prevent.

You say it works "just by download-unpack" as it is, but I disagree for at least two reasons.

- If the binary package included the static desktop file and did not insist on generating one, that would be the case (download, unpack to the proper directories for your system, launch normally). Instead, it's download, unpack, open a terminal, launch telegram, quit telegram, launch telegram normally.

- For another, if a user has multiple Telegram clients installed, installing this one as officially packaged messes with their xdg settings without permission, doesn't it?

So yes, I am very interested in the compile time option you're referring to, if it exists. But it's worth noting that a complicating factor is that compiling this program is an enormous task, considering the gigabytes of custom qt required, as well as plenty of libraries that are otherwise unnecessary for non-unity, non-gtk desktops.

john-preston:
I mention the way this could be added to the source code -- not by stripping out the code, but by surrounding it by

#ifdef TDESKTOP_DISABLE_DESKTOP_FILE_GENERATION
.. stripped by you code here ..
#endif // TDESKTOP_DISABLE_DESKTOP_FILE_GENERATION

then it could be merged and anyone who doesn't need this while compiling could disable it for himself.

What exactly do you disagree with? There is a huge amount of users, who just download, unpack in any directory (like Desktop or Downloads), and launch the current version - and it works, with silent background autoupdate. So any change to the source code that is suggested should not break anything for them.

AndydeCleyre:
OK, I've made changes to implement the bugfix the way you suggest.

> What exactly do you disagree with?

As I wrote on one of the bug reports for this:

> . . . on Linux desktops, it's not the role of a program to place files like that on the user's system in the proper place. That's what build systems and package managers do.

The way things are causes unexpected meddling on systems, in a way that is unavoidable once compiled. It's an impolite software-neighbor.

The ability to compile without that is good, thank you. But it is such a PITA to compile that  it would be nice if the binaries provided included those suitable for friendly interaction with Linux desktop standards.


Why is it an PITA to compile?

- Huge custom Qt framework. Would be much simpler if this could use the standard.
- AFAIK, libunity and libappindicator-gtk2 (and everything those entail) must be installed even if the resulting binary will not be used in an environment with Unity, or AppIndicators, or GTK2.

So I think that behavior of unprompted launcher generation and placement is 100% wrong, and the PITA aspect exacerbates the problem.

AndydeCleyre:
It looks like this pull request is not getting my repo changes, probably because it's closed.  I don't know if it will resume if the request is re-opened, or if I need to submit a new one.

auchri:
@AndydeCleyre let's try :)

AndydeCleyre:
Alright, I've made those changes. I also swapped in `ifndef` instead of `ifdef` -- it should be `ifndef`, right?

auchri:
yes, lgtm :+1:

auchri:
@AndydeCleyre do you mind if I ask you to [squash the commits](https://github.com/telegramdesktop/tdesktop/blob/master/.github/CONTRIBUTING.md#squash-your-commits-to-a-single-commit)?

AndydeCleyre:
Squashed!

henning-schild:
I would like to suggest a global config-file in /etc/ where this feature and the auto-updater can be disabled globally without having to rebuild telegramdesktop.

The application is hard to build from source, because it requires patching and rebuilding qt as well. (at least that was the case, did not check in the last year) And at the same time the .desktop-file and the auto-updater are issues for distributions that ship telegramdesktop-packages.

The global config file could be shipped only by distributions, turning off the problematic features. And people that just download the application from telegram.org would still have those features enabled by default.

PR #2200, opened 2016-06-28
New define: TDESKTOP_DISABLE_UNITY_INTEGRATION

eegorov:
Make possible build on system without unity (like Gentoo)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ env:
   - BUILD_VERSION="disable_crash_reports"
   - BUILD_VERSION="disable_network_proxy"
   - BUILD_VERSION="disable_desktop_file_generation"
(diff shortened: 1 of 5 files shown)


eegorov:
Make possible build on system without unity (like Gentoo)

auchri:
The new flag should also added to the travis build ([here](https://github.com/telegramdesktop/tdesktop/blob/master/.travis.yml) and [here](https://github.com/telegramdesktop/tdesktop/blob/master/.travis/build.sh#L78)).

eegorov:
I make that. But I do not know how chech this.

auchri:
looks good :+1:

john-preston:
Better to make one commit only (squash).

eegorov:
Done

PR #2211, opened 2016-07-02
Add comment in Windows build instructions that gclient sync error can be ignored

splintor:
In my case, I got this error, and was afraid that something went
wrong,
but I was able to successfully build and run Telegram.

### DIFF
--- a/doc/building-msvc.md
+++ b/doc/building-msvc.md
@@ -214,6 +214,8 @@ and run
     ..\depot_tools\fetch breakpad
     ..\depot_tools\gclient sync
 
+`gclient sync` might print `Error: client not configured; see 'gclient config'` - You can safely ignore it.
+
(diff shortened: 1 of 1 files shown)


auchri:
I think the correct way to fix this is to run `gclient config "https://chromium.googlesource.com/breakpad/breakpad.git"` before `gclient sync`


...I will create another PR.

auchri:
#2217

splintor:
Even better. Thanks!

Zenexer:
@auchri  
> I think the correct way to fix this is to run gclient config "https://chromium.googlesource.com/breakpad/breakpad.git" before gclient sync

Not quite.  fetch actually runs client config as part of is process.  That needs to take place in the breakpad directory, which is why it's run there.  Running gclient config with that URL in the depot_tools directory is akin to changing the primary remote of a git repo to point to an unrelated repo, whereas when we don't run gclient config, we're lacking a remote altogether.  You can configure it for depot_tools if you must, but you'll end up with a lot of stuff you don't need, if I recall correctly.  When I wrote those steps, I found documentation somewhere by Google that recommended only partially installing depot_tools to gain access to fetch and gclient; then, if the rest of the toolkit was needed, gclient could be used to retrieve the rest.

Issue #1367, opened 2015-12-08
Pressing ESC on any window should close it


marcopaganini:
Pressing ESC on the conversation window brings back the main window with the contacts. Pressing ESC on the main window does nothing. It should ESCape from this window (i.e, close it). Not having a way to close the window quickly could represent a privacy problem for those using telegram in a low privacy environment.

auchri:
No, just closing the current conversation is the right behavior. There will be a shortcut to close the window, but for that see #1060 

fyi alt + F4 is also fast

marcopaganini:
Hmm, any reason why not close the window with ESC? ESC is somewhat
"universal" for "get me out of here" and it would make sense to close all
windows with it. Of course, I'm not talking about exiting telegram, just
closing all the windows.

splintor:
I agree. Would also like Esc to close the window when no chat is open. Alt+F4 is less convenient.

ralesk:
I'm very much "No" on this one.

It's popup dialogs and other modal things that can be dismissed with Esc.  Every main window you encounter is not closeable with Esc (look at pretty much any other chat application, browsers, etc.)

On the other hand, Esc should only ever close one level of modal stuff — i.e. if I have an image open, pressing Esc should only close the lightbox and not the chat it was opened from.  But that's a different bug really.

marcopaganini:
What about Ctrl-W? Seems to be the gnome recommended way (but I could be
wrong.)

ralesk:
That one seems to be fairly common for "Close Window" (also known as, nowadays I think less often, Ctrl+F4 in the Windows world).

splintor:
@ralesk OK, so why not make it optional? For me, it is very convenient to use Esc to close the current conversation, and then another Esc to close the Telegram window and go back to my work (note that I use Telegram as a "pop-up" to do some chat, and not as a main working tool like a browser or a text editor).
However, I really don't mind having it optional, and even have it disabled by default, as long as I have the easy option to enable it, so I can easily press Esc to dismiss the entire Telegram window. Currently, I have to use Alt+F4 to close it, which is less convenient.

marcopaganini:
+1 on making ESC an option.

ANSI-C:
Any news on this? Are they willing to make it optional? The desktop version is useless for me without this basic behaviour that dates back to nineties.

john-preston:
[...]

PR #2212, opened 2016-07-02
Add option to close Telegram when Escape is pressed

splintor:
Should fix https://github.com/telegramdesktop/tdesktop/issues/1367

Missing settings string translations

### DIFF
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -259,6 +259,7 @@ Copyright (c) 2014-2016 John Preston, https://desktop.telegram.org
 "lng_settings_auto_start" = "Launch Telegram when system starts";
 "lng_settings_start_min" = "Launch minimized";
 "lng_settings_add_sendto" = "Place Telegram in «Send to» menu";
+"lng_settings_close_main_on_escape" = "Close Telegram when Escape is pressed";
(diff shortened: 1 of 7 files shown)


auchri:
I think this would be a nice addition :+1:

splintor:
Any news?

auchri:
@john-preston

john-preston:
@auchri Perhaps after the settings are redesigned this option will be added. Right now the settings page is already too complex, so I don't want to fill it with additional controls for different minor settings.

splintor:
@john-preston If this is the problem, can we at least add the option without adding the change in the Settings UI, so those who desperately want this option will be able to have it?

splintor:
Can you please explain why won't you mergeg my pull request? I really love Telegram and this is the only thing that bugs me (and others).

john-preston:
@splintor Because the settings are different now and this pr is not mergeable anyway. Also a checkbox in settings for that is not an option right now as I've mentioned above. Some way of geek-only settings file should be added for such things I guess.

splintor:
@john-preston If I file another PR that used a "geek-only settings file" for that, will you reconsider merging it?

stek29:
@splintor If you add "geek-only settings file" I'm 99% sure it would be merged, because lots of issues aren't fixed to avoid settings cluttering.

PR #2213, opened 2016-07-02
Add fast finish to travis

auchri:
> Now, a build will finish as soon as a job has failed, or when the only jobs left allow failures.

See https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing


john-preston:
oops..

PR #2220, opened 2016-07-03
Automatic comment for missing signatures

auchri:
With this PR, travis will add a comment to all PR's which have a commit which is missing the signature.

**Depends on #2210 and #2213 (merge first!)**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ sudo: required
 language: cpp
 
 env:
+  global:
+   # GitHub auth token (GH_AUTH_TOKEN)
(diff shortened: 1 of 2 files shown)


john-preston:
oops..

stek29:
I'm not sure, but would this really work?
[From Travis docs:](https://docs.travis-ci.com/user/environment-variables/#Encrypted-Variables)
> Encrypted environment variables are not available to pull requests from forks due to the security risk of exposing such information to unknown code.

auchri:
😢 😢 😢

Issue #982, opened 2015-08-09
Can the Linux client use GTK FileChooser?


orschiro:
Hi all, 

Great work on the Linux Desktop client. Thanks so much!

I have one question:

Currently, the client uses a very impractical and rather ugly file choosing window. 

![image](https://cloud.githubusercontent.com/assets/170947/9154656/cdf396aa-3e9b-11e5-9cdb-6afa57c1f5a2.png)

Would it for the client be possible to use the native GTK File Chooser?

See this example of how this could look using when opening a file, for instance in Chrome:

![image](https://cloud.githubusercontent.com/assets/170947/9154660/2542962c-3e9c-11e5-90cc-db0c0891bca6.png)

Thankfully,

Robert

mareg:
:+1:

ManuelSchneid3r:
Technically this is nonsense since, tgram is powered by Qt. In case of the desirable goal to dynamically link on linux this would introduce unneccesary dependencies.

Normally qt styles are consistent with gtk. But somehow the author decided to set the qt style to win. This affects the systray menu too.

orschiro:
@ManuelSchneid3r 

Thanks for pointing that out! I haven't changed or customised the Qt settings. I am using the defaults as provided by elementaryOS. 

I have asked the question on [elementaryOS StackExchange](http://elementaryos.stackexchange.com/questions/1240/how-to-set-qt-style-to-default) how to change the default style.

ManuelSchneid3r:
I am not experienced with this, but there is an option "-style" which is gereral to all qt applications. Unfortunately it is not working. After further reading, I guess this is due to the fact that the author wrote its own style (checkbuttony, slider,etc), which is based on the win style. The ugly xp parts must be parts that he has not overridden yet.

lunix4:
This software use Qt toolkit with static linking lib, Qt/C++ classes is inherited and changed from Qt5 original classes , so the style option will not work except the developer add style option.

auchri:
duplicate #438

auchri:
Discuss further at #987 :)

ManuelSchneid3r:
These are not dups. #987 is about the style of the application this one is about the filechooser type.

telegramdesktop:
@ManuelSchneid3r They both are about the file chooser style.

ManuelSchneid3r:
[...]

Issue #1404, opened 2015-12-16
File Chooser on Linux is Butt Ugly


mypetbirdrules:
The file chooser in telegram desktop doesn't respect system-wide themes from gtk.

DaFri-Nochiterov:
Duplicate #987

ManuelSchneid3r:
I am the author of #987. #987 is not about the filechooser. #987 is about the application style in general. Check #982 for native filechooser.

john-preston:
In the latest 0.9.57 version gtk file chooser should be used if it is available. Please check.

Issue #1140, opened 2015-09-29
The native file browser is not used for sending images etc.


adiultra:
I have installed telegram on elementary OS Freya.
Everything is working fine except that telegram uses an awkward prompt for selecting files or images to send. 

![choose images_085](https://cloud.githubusercontent.com/assets/12121575/10162805/7ea8849a-66cb-11e5-8c4f-21071a61017c.png)

whereas the background choosing dialogue is quite nice

![telegram_084](https://cloud.githubusercontent.com/assets/12121575/10162818/967e2a7a-66cb-11e5-8741-5319c4545d83.png)

Can you please change the default file chooser or use the one of the native OS.

auchri:
Duplicate #438

Issue #987, opened 2015-08-10
Make tdesktop look native / Do not set Qt style to win by default


ManuelSchneid3r:
Telegram does not use the gtk style which uses gtk+ engine to render the widgets. Please do not do this. Just do not set it manually at all. As far as I know qt takes the correct style automatically.

Further reads:
http://doc.qt.io/qt-5/qstyle.html#details
http://doc.qt.io/qt-4.8/qgtkstyle.html#details

Yanpas:
Also stop using Open Sans, distance between symbols in some alphabets (russian e.g.) is too big

evgenymarkov:
Linux users want native file dialog and tray menu! QFileDialog does not use the native dialog by default?

telegramdesktop:
I do nothing with the styles, but I use Qt static build, perhaps that is the cause of non-native look.

nortexoid:
Proper theming would be nice, but native picker would be even better. I don't how they do it, but Chrome on linux manages to chooses the gtk picker on Gnome and the kde picker ("KDE File Dialog Widget") on KDE. I would at least be happy with proper theming!

ManuelSchneid3r:
@nortexoid This would introduce unneccesary dependencies. Qt way is to do it platform independent. 

@telegramdesktop At lest the style set manually get [overidden](https://github.com/telegramdesktop/tdesktop/blob/85a2fb18764a4b75a6443636814ee6dbf7349a5c/Telegram/SourceFiles/main.cpp#L69).

Yanpas:
@ManuelSchneid3r Why would it introduce dependecies? Make algorithm like: if there is gtkpeeker - use gtkpeeker, if there is KDE-file-dialog - use kdefile dialog. Otherwise use native. Chrome doesn't have any dependencies

nortexoid:
@Yanpas Sounds right to me. The Qt5 build of QupZilla I'm using doesn't have any GTK/KDE dependencies but it still allows the use of native file picker.

solbjorn:
@Yanpas, Qt-based applications should automatically use system-wide look without any algorithms. As @ManuelSchneid3r said, currently there is an overrride of default style in Telegram code (through fake args), that prevents it.

Lesik:
[...]

PR #2221, opened 2016-07-03
Native look of file choosing dialog on GTK Linux systems

BOOtak:
before
after

### DIFF
--- a/Telegram/SourceFiles/platform/linux/linux_libs.cpp
+++ b/Telegram/SourceFiles/platform/linux/linux_libs.cpp
@@ -130,28 +130,17 @@ void start() {
 	bool gtkLoaded = false;
 	bool indicatorLoaded = false;
 	QLibrary lib_gtk, lib_indicator;
-	if (loadLibrary(lib_indicator, "appindicator3", 1)) {
-		if (loadLibrary(lib_gtk, "gtk-3", 0)) {
+	if (loadLibrary(lib_indicator, "appindicator", 1)) {
+		if (loadLibrary(lib_gtk, "gtk-x11-2.0", 0)) {
+			gtkLoaded = indicatorLoaded = false;
 			gtkLoaded = setupGtkBase(lib_gtk);
 			indicatorLoaded = setupAppIndicator(lib_indicator);
-		}
-	}
-	if (!gtkLoaded || !indicatorLoaded) {
-		if (loadLibrary(lib_indicator, "appindicator", 1)) {
-			if (loadLibrary(lib_gtk, "gtk-x11-2.0", 0)) {
-				gtkLoaded = indicatorLoaded = false;
-				gtkLoaded = setupGtkBase(lib_gtk);
-				indicatorLoaded = setupAppIndicator(lib_indicator);
-			}
 		}
 	}
 
 	// If no appindicator, try at least load gtk.
 	if (!gtkLoaded && !indicatorLoaded) {
-		if (loadLibrary(lib_gtk, "gtk-3", 0)) {
-			gtkLoaded = setupGtkBase(lib_gtk);
-		}
-		if (!gtkLoaded && loadLibrary(lib_gtk, "gtk-x11-2.0", 0)) {
+		if (loadLibrary(lib_gtk, "gtk-x11-2.0", 0)) {
 			gtkLoaded = setupGtkBase(lib_gtk);
 		}
 	}
--- a/Telegram/Telegram.pro
+++ b/Telegram/Telegram.pro
@@ -25,6 +25,7 @@ macx {
 linux {
     SOURCES += ./SourceFiles/pspecific_linux.cpp
     HEADERS += ./SourceFiles/pspecific_linux.h
+    QTPLUGIN.platformthemes += qgtk2
 }
 
 CONFIG(debug, debug|release) {


BOOtak:
Screenshots attached

auchri:
Fixes #982 #1140 #1404 #929

melvinvermeeren:
I'm not sure how this affects the possibility of `-no-gtkstyle` being removed to fix the long-standing `-style=0` problem that has been an issue for a while.

It also appears that the GTK version used for the file picker is now 2 instead of 3, which is afaik not the preferred choice for a DE that natively uses 3.

This still doesn't give a native filepicker for e.g. KDE and probably other DEs.

Consider the following example.

```c++
#include <QtWidgets/QApplication>
#include <QtWidgets/QFileDialog>

int main(int argc, char **argv)
{
	QApplication *qa = new QApplication(argc, argv);

	QFileDialog::getOpenFileName
	(
		nullptr,
		"Open File",
		"/home",
		"Images (*.png *.xpm *.jpg)"
	);

	delete qa;
	qa = nullptr;

	return 0;
}
```

compile it with and run `./a.out`
```
g++ -I/usr/include/qt -fPIC -std=gnu++14 -lQt5Core -lQt5Widgets main.cpp
```

kdefilepicker

As you can see, the native system filepicker is used automatically. In my case KDE 5's file picker. I suspect this behaviour is consistent across OSes and DEs, since Qt prefers native when possible.
http://doc.qt.io/qt-5/qfiledialog.html#getOpenFileName
http://stackoverflow.com/a/6405275

Perhaps it is a better idea to let Qt itself choose the system native dialog. It is only when a lot of `QFileDialog` options are specified that Qt is **FORCED** to use the built-in dialog since it only then can be absolutely sure that the requested options are available.

If you ask me this is the solution. Less option forcing and hacking around a problem by inserting 3 other problems at the same time.

View https://github.com/telegramdesktop/tdesktop/issues/987#issuecomment-230170321 for some background and the possible fix for `-style=0`.

melvinvermeeren:
Asked a friend using Gnome 3 to also try the above code.

image_2016-07-03_23-03-18

ManuelSchneid3r:
Is it just me, or is this PR basically just removing everything o GTK3?

Note: Qt 5.7 does heavily mess up qt - gtk interaction, by moving GTK+ style to the qtstyles plugin.

BOOtak:
@ManuelSchneid3r it's more you then PR as Qt 5.7 [is unsupprted](https://github.com/telegramdesktop/tdesktop/blame/master/README.md#L19) by TDesktop client. Compile proect with 5.6 and everything will be just fine.

BOOtak:
@melvinvermeeren please explain wtf is `-style=0` issue or give me the link.

BOOtak:
@melvinvermeeren TL;DR: please read project's source code carefully before writing comments next time. I have no idea why I should spend my time to explain obvious things.
> It also appears that the GTK version used for the file picker is now 2 instead of 3

No. I've changed GTK version for absolutely different reason (read my commit message carefully) and it has no connection at all with file picker's theme.
> This still doesn't give a native filepicker for e.g. KDE and probably other DEs.

Again, qgtk2 is a  __theme__ __engine__ and will be applied __only__ to GTK file picker if it will be used. As written in Qt docs, `By default, a platform-native file dialog will be used if the platform has one` which is __not__ GTK file picker on KDE. So, my PR has no connection to and no influence on native KDE file picker. I have no idea why KDE doesn't use it's own file picker but I strongly confident that it's completely separate issue and I can't wait for your PR to fix this.

> Perhaps it is a better idea to let Qt itself choose the system native dialog. It is only when a lot of QFileDialog options are specified that Qt is FORCED to use the built-in dialog since it only then can be absolutely sure that the requested options are available.

I see you haven't read project source code at all as project's [QFileDialog utilizes only common set of options which every native implementation has for sure.](https://github.com/telegramdesktop/tdesktop/blob/master/Telegram/SourceFiles/ui/filedialog.cpp#L84)

> Consider the following example.
> g++ -I/usr/include/qt -fPIC -std=gnu++14 -lQt5Core -lQt5Widgets main.cpp

Your example will be compiled with system Qt which already built with all possible platforms ant theme engines. Tdesktop uses it's own static version of Qt patched and built from source code. Let me enhance your example and show what would happen if we compile it with static Qt:

```qmake
QT += core gui network widgets

CONFIG += plugin static c++14

QT_TDESKTOP_VERSION = 5.6.0
QT_TDESKTOP_PATH = $$(QT_TDESKTOP_PATH)

INCLUDEPATH += $${QT_TDESKTOP_PATH}/include/QtGui/$${QT_TDESKTOP_VERSION}/QtGui \
               $${QT_TDESKTOP_PATH}/include/QtCore/$${QT_TDESKTOP_VERSION}/QtCore \
               $${QT_TDESKTOP_PATH}/include

SOURCES += \
    ./test.cpp

INCLUDEPATH += \
               /usr/local/include

QTPLUGIN.platformthemes += qgtk2

LIBS += $${QT_TDESKTOP_PATH}/plugins/platforminputcontexts/libcomposeplatforminputcontextplugin.a \
        $${QT_TDESKTOP_PATH}/plugins/platforminputcontexts/libibusplatforminputcontextplugin.a \
        $${QT_TDESKTOP_PATH}/plugins/platforminputcontexts/libfcitxplatforminputcontextplugin.a
```

This is test.pro file copy-pasted from Telegram.pro with slight modification (some parts are thrown away). 

Let's build project:
`qmake test.pro`
`make`

And launch it:
`./test`

Here is the result:


WTF, right?
And now we add one line to our `test.pro`:

```qmake
QTPLUGIN.platformthemes += qgtk2
```

And here is the result:


So, solution provided in this PR only applies correct theme to GTK file picker, no less, no more. It doesn't influence KDE or other platforms which have their own picker.

melvinvermeeren:
@BOOtak Thanks for explaining more about the state of things with tdesktop itself, I didn't know all of this yet.

Even so, instead of changing even more inside tdesktop itself, wouldn't it be a better idea to change the configuration of the static Qt build to include support for native filepickers in the same way the system Qt does? Even if this increases the binary by a few M it shouldn't pose a problem.

I just can't help but feel that the problems are being solved in a nonstandard ways and will most likely cause further issues later on, there have been too many cases of this already with tdesktop imo. Kind of off-topic, but I'd like to see a truly native tdesktop one day.

john-preston:
@BOOtak I don't like the idea of removing the "appindicator3" loading (which requires "gtk-3", if I understand correctly the difference between "appindicator" and "appindicator3" libraries).

For example, Ubuntu 15.04 has libappindicator3.so but doesn't have libappindicator.so, so removing the gtk-3 version of libappindicator will turn off the tray icon support on all the systems, that rely on libappindicator3.so, which I think is not acceptable.

BOOtak:
@john-preston 
> requires "gtk-3", if I understand correctly the difference between "appindicator" and "appindicator3" libraries

Exactly!

> removing the gtk-3 version of libappindicator will turn off the tray icon support on all the systems,

That is an issue indeed. Is it clear why I've removed gtk-3 libs? (Hint: to avoid mixing gtk-2 and gtk-3 symbols which would cause app's crash). So, static linking with libappindicator should do the trick/ i'll rebase my changes on top of latest master branch and try to link appindicator statically.

john-preston:
libappindicator can't be statically linked as I know (at least I could not do that at the time I tried), so I started to load it dynamically and load the needed version of gtk library.

I'm afraid the only way to make things working with both appindicator, appindicator3 and gtk file picker is to write specific code for gtk file picker usage instead of Qt file pickers, which will work both with dynamically loaded gtk2 lib and gtk3 lib :(

BOOtak:
@john-preston why you load gtk-3 for libappindicator3 on runtime when it can find gtk-3 itself as it's linked with it dynamically?

john-preston:
@BOOtak I don't see this comment on github.. I call methods from libgtk together with libappindicator methods, I guess they should be from the same version of libgtk.

I'll try to implement (copy paste :) my own gtk file dialog that'll work with both gtk2 and gtk3, I hope this will work.

john-preston:
Another problem is that adding "QTPLUGIN.platformthemes += qgtk2" statically links a huge amount of libraries, which were not in dependencies for the currently deployed build.

BOOtak:
@john-preston those new libs should be mentioned in documentation and travis build agent as soon as we handle all other issues.

john-preston:
@BOOtak The problem with them is not the inconsistency of docs or travis failing. The problem is that the binary may stop launching on some systems where it does work now. A new version can't be deployed that will just stop working on some systems.

john-preston:
I've pushed to master branch a version with self-made GtkFileDialog (based on Qt theme plugin code), it works for me in Ubuntu 12.04 for both gtk-2 and gtk-3.

BOOtak:
That is definitely better solution than mine! I'll test that. Thank you anyway for paying attention to this problem as previous look of file choosing was horrible  :)

john-preston:
In the latest 0.9.57 version gtk file chooser should be used if it is available. Please check.

kvaster:
File chooser on gentoo with KDE is broken now. GTK file chooser is also available, cause I have some GTK apps installed, but native file chooser is really much better.

john-preston:
@kvaster was it native file chooser before? All systems I've tested with old tdesktop version were using Qt inner file chooser implementation, that was not like "kdialog" in KDE, for example. Can you test

https://telegram.org/dl/desktop/linux (without updating to alpha -- disable autoupdate in Settings)

version and say if it is really better (native?), than the new 0.9.57 alpha? If it does -- maybe you can tell me, how I can understand in the code (read some env variable?) do I need to try to use GTK file chooser or not.

kvaster:
Seems you're right - it was opening Qt inner file chooser.

http://imgur.com/a/j7EiJ

I will try to figure out today if there is reliable method to detect KDE and to open Qt file chooser instead of gtk. On kde Qt chooser looks much better in my opinion.

AndydeCleyre:
@kvaster 

> if there is reliable method to detect KDE

```bash
$ echo $DESKTOP_SESSION
```

```
/usr/share/xsessions/plasma
```

PR #2242, opened 2016-07-11
Replace hard-coded paths with pkg-config

gavv:
Hi,

This PR replaces hard-coded paths in `Telegram.pro` with [`pkg-config`](https://en.wikipedia.org/wiki/Pkg-config) invocation. All remaining `/usr/local/` references (which usually don't present on user system) are moved to `build/build.sh`, so that `Telegram.pro` is now generic and should work on any system without patching paths.

### DIFF
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -58,8 +58,6 @@ prepare() {
 	fi
 
 	sed -i 's/CUSTOM_API_ID//g' "$srcdir/tdesktop/Telegram/Telegram.pro"
-	sed -i 's,LIBS += /usr/local/lib/libxkbcommon.a,,g' "$srcdir/tdesktop/Telegram/Telegram.pro"
-	sed -i 's,LIBS += /usr/local/lib/libz.a,LIBS += -lz,g' "$srcdir/tdesktop/Telegram/Telegram.pro"
 
 	local options=""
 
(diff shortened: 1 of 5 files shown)


john-preston:
I'll have to remove xkbcommon from PKGCONFIG and return "/usr/local/lib/libxkbcommon.a" link, because PKGCONFIG usage results in "-L/usr/local/lib -lxkbcommon" flags, who link libxkbcommon dynamically, instead of static link. Which leads to that segfault:
image

gavv:
You probably could replace:

    PKGCONFIG += xkbcommon

with

    QMAKE_CXXFLAGS += `$$PKG_CONFIG --cflags xkbcommon`
    QMAKE_LFLAGS += `$$PKG_CONFIG --libs --static xkbcommon`

to force static linking.

If it doesn't work, I suggest to keep `pkg-config` usage optional instead of completely removing it (and probably use it by default).

Since `/usr/local/lib/libxkbcommon.a` is specific to your system and is not present on user system, it doesn't allow to build telegram without patching it, which complicates life for users and distro maintainers that would like to package telegram.

john-preston:
Looks like gentoo kde build was also affected #2278 I don't know what else could break it.

gavv:
Uh, great caught, sorry for that.

How can one prepare environment for production build?

To avoid introducing such bug next time, I'd like to run [`Telegram/build/build.sh`](https://github.com/telegramdesktop/tdesktop/blob/master/Telegram/build/build.sh) in the same environment as the official build. I guess my environment differs somehow.

Could you provide instructions for preparing such an environment? I think it will help contributors a lot. Ideally, it can be a script or maybe a docker container, but just a text description will be ok too.

john-preston:
@gavv It didn't crash on my build environment (on Ubuntu 12.04) or on any other configurations I tried, it happened only on an end-user machine. So I don't see how this could help.

But I guess it could be predicted, because when Qt was built there was "OPENSSL_LIBS='-L/usr/local/ssl/lib -lssl -lcrypto'" param passed to Qt configuration, so the final executable should have been built with the same ssl/crypto libraries.

The environment is just almost like in QtCreator build instructions, I don't remember, if there were any changes.

PR #2248, opened 2016-07-14
Improve travis build process

auchri:
- Build directly on ubuntu
- Cache custom qt and breakpad

Things to improve:

- Cache should be deleted after a certain amount of time (e.g. 1 week)
- QT should be rebuilded if the patch is changed


john-preston:
:like:

PR #2321, opened 2016-08-06
Remove rounded rectangle message bubble style

wbsdty331:
added signature.

* advise 1

`macOS` version use Rounded rectangle style
`Windows` & `Linux` keep original style.

* advise 2

make it optional.
If you like rounded rectangle bubble style, Just enable it.

Which do you prefer?

### DIFF
--- a/Telegram/Resources/basic.style
+++ b/Telegram/Resources/basic.style
@@ -665,7 +665,7 @@ scrollDef: flatScroll {
 	hiding: 1000;
 }
 
-msgRadius: 16px;
+msgRadius: 3px;
 dateRadius: 10px;
 buttonRadius: 3px;
 


matteocontrini:
Can you post a screenshot of how this "3px" rounded style looks exactly?

wbsdty331:
Please use 0.9.56 Version
That's normal message bubble style.

lgommans:
Can someone with commit access please give us an update on this? Like, it's a config file change, it can't break anything. Can we merge this already or are there still considerations to be made?

Geobert:
What are we waiting for?

nocturnal3d:
+1 👍

wbsdty331:
Nope?

chermenin:
Old style of messages really better. Give it back!

daslapovsky:
+1 for old style.

Akiyamka:
Now the rounded corners look very clumsy. Old design telegram though was nothing special, but it was rather restrained so as not to annoy the user. Let's return everything as it was, and experimentation with alternative themes design leave registration

PR #2337, opened 2016-08-08
Wide screen view improvements

timocov:
Hi guys,
this merge request refers to #2060 (and similar issues).
Below you can find comparison screenshots.

Messages before:
Messages before

Messages after:
Messages after

Group/Bot/Channel info before:
Group/Bot/Channel info before

Group/Bot/Channel info after:
Group/Bot/Channel info after

Issue #2060, opened 2016-05-20
Allow to resize message block in chat window


f1ztech:
Long messages has small width and large height, like a newspaper column. Sometimes It's hard to read such messages, also they take a lot of space on screen and make chat history less usable for me. Besides as a programmer I often send a code snippets in telegram, and it is practically impossible to read them because of small message width. 

 
### Steps to reproduce

Open chat window
### Expected behaviour

User can change width of message with mouse
### Actual behaviour

Message width is fixed
### Configuration

**Operating system:**
Windows
**Version of Telegram Desktop:**
0.9.49

eode:
Same issue.  Message width adapts appropriately to narrowing window size, but stops expanding after a certain point.  on my screen (1920x1080), that means messages use only 1/3rd of the available space.  As f1ztech pointed out, this also affects code block comments.

lexa971:
Same issue. Very annoying. Must fix it asap please...

kotep:
Providing an example of wasted space with fixed message width.
Expected behavior would be message width maximum is maximum width of message pane.  
What's the purpose of maximizing if the messages remain fixed-width?

![untitled-1](https://cloud.githubusercontent.com/assets/7033407/17412897/550c3d10-5a3c-11e6-8c25-ba35fbd9dbd4.jpg)

Alukardd:
Version 1.0 released and we still seeing ugly narrow column instead of code block, backtrace, etc...
I can understand such behavior on mobile client, but on my laptop with 2880x1620 resolution it is a absurd!

rotanov:
I vouch for this. Was waiting for it so long so I decided to make an issue. Since I didn't found this one.

valcker:
+1 for this.

francescortiz:
Pasting code becomes really annoying with this limited width. It seems kind of contradictory to have the possibility to quote code but not being able to properly see any line with more than 52 characters. Also, some programming languages don't follow the 80 column styling, so limiting the width is limiting usability.

linermgn:
Just do it, please, logs and code blocks looks awfull

kotep:
In v1.1, the client now uses more screen space on large windows to display a cluttered and distracting emoji window, and still does not let the message size increase beyond its current width.

sbagan:
Yay! Issue is opened for more than one year.
Seriously please fix it

proea:
Please fix it!

Xorboo:
[...]

PR #2339, opened 2016-08-09
Wide screen view improvements

timocov:
Hi guys,
this merge request refers to #2060 (and similar issues).
Below you can find comparison screenshots.
(old pr #2337)

Messages before:
Messages before

Messages after:
Messages after

Group/Bot/Channel info before:
Group/Bot/Channel info before

Group/Bot/Channel info after:
Group/Bot/Channel info after

Max size of photo's dimension is 420px, so we can use it space for photos.

Photo with caption before:
before

Photo with caption after:
after

### DIFF
--- a/Telegram/SourceFiles/history.cpp
+++ b/Telegram/SourceFiles/history.cpp
@@ -2623,13 +2623,6 @@ void HistoryMessageUnreadBar::paint(Painter &p, int y, int w) const {
 	p.setFont(st::unreadBarFont);
 	p.setPen(st::unreadBarColor);
 
-	int left = st::msgServiceMargin.left();
-	int maxwidth = w;
-	if (Adaptive::Wide()) {
-		maxwidth = qMin(maxwidth, int32(st::msgMaxWidth + 2 * st::msgPhotoSkip + 2 * st::msgMargin.left()));
-	}
-	w = maxwidth;
-
 	p.drawText((w - _width) / 2, y + marginTop() + (st::unreadBarHeight - 2 * st::lineWidth - st::unreadBarFont->height) / 2 + st::unreadBarFont->ascent, _text);
 }
 
@@ -4017,7 +4010,12 @@ void HistoryDocument::initDimensions() {
 
 	if (auto named = Get<HistoryDocumentNamed>()) {
 		_maxw = qMax(tleft + named->_namew + tright, _maxw);
-		_maxw = qMin(_maxw, int(st::msgMaxWidth));
+
+		if (Adaptive::Wide() && captioned) {
+			_maxw = qMax(captioned->_caption.maxWidth(), _maxw);
+		} else {
+			_maxw = qMin(_maxw, int(st::msgMaxWidth));
+		}
 	}
 
 	if (thumbed) {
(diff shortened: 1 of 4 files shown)


stek29:
Can you squash this commits into one?

john-preston:
This is not going to be merged, sorry. The layout like "Messages after:" was explicitly avoided by setting the maximum width of the text message bubble, because it is less readable (reading the line up till the end in the right side of the monitor you can't quickly jump to the beginning of the next line, you need to follow all the line back). When you use small width of the message layout it is easy to read.

timocov:
So, maybe we will use increased max width of messages for wide layout? For example, twice as much.

stek29:
It would be good solution to make it configurable IMO.
But it would make settings more complicated.

Scukerman:
I often send my code via Telegram or even some traces so that would be a good thing to make it work at least as an option.
So, +1

Kondra007:
@Scukerman Then maybe messages containing only code (formatted text) should be wide, not every message.

vstepchik:
@john-preston Could you please tell if it makes sense increasing the width for code snippets? While I agree that narrow balloons are easier to read, on the other hand it makes reading some code the opposite.

timocov:
@john-preston @auchri Telegram for Mac has no constrains of message's width and seems everything is OK.
I guess if someone want to see more - he will maximize (or increase size of) the window. But we forcedly constrains width of the message and everybody cannot see more that he see when size window's size is (approx.) 760-780px.

stek29:
One more place where that _geeky settings file_ is needed.

cauebs:
I understand that it may be less readable, but in some situations it's the exact opposite.
Sometimes I just want so share some code and it looks awful.
Would it be too hard to just make it configurable? I see nothing wrong with that

msva:
@timocov What do you think about reworking it a bit to make this behaviour configurable?

@john-preston what do you think about merging such PR if that behaviour will be configurable? Take a notice, that there are many users in this and similar issues on this there wants full-width text instead of "default" behaviour too.

Yes, *some* people may be "hard to find strat of the next line, when you'll finish read long line", but there is also many people who will be ok with that.

And also many people who need to paste logs, code snippets and so on, which is **totally** unreadable in current layout.

timocov:
@msva I'm agree with your suggestion

kirsan31:
Any news?

msva:
@kirsan31 nope. Devs just silently ignores the issue :(

// by the time, I just ported the patch to the new codebase (although, I still didn't added the checkbox), and also I did the patchset to make tdesktop to build against system Qt5, without requiring to patch. All of them are in my gentoo overlay, if any :)

Fusl:
I don't know who does design change decisions for the desktop app here but they definitely have no idea about UI and UX designing. It really s*cks that text keeps wrapping around whereas my display has enough space to display the entire message in just a single line or a few lines. Less horizontal space used means more vertical space used and I, as a user, have to waste time scrolling through the chat to find a text.

Imagine, Firefox would just hardlimit the amount of memory it actually consumes of my available memory, like, 256MB usable from 288GB. You want more? You need to open multiple Firefox processes. Your OS only allows you to write files of up to 1GB to your 10TB disk. Want more? You'll need to `split` it into 1GB sized files.

PS, this is how Telegram looks on a standard 4k monitor with native DPI configured which is a ridiculous waste of available space.

ab0c49839cbb

Now also imagine, you're actually using Telegram to do some productive work. Copy-paste code from your OSS into Telegram with backticks, this is how Telegram messes everything up and makes EVERYTHING hard to read, essentially forcing users to copy-paste text from Telegram into a standard text editor.
075bca0ada9f
OHH, look. Not even your own code fits in here without being line wrapped 😂

I bought a 4k screen for a reason, and that reason is definitely not that I want my apps to only use 15% of my actual displays capable resolution.

msva:
I've patch for tdesktop that makes it to use "wide" baloons (based on work from neighbour pull-request, but tdesktop's code has changed few times since then, so I adapted it few times already).
If you've possibility to re-compile telegram I can give you the link (or, if you use Gentoo, you can just add my overlay (`mva`) and emerge `telegram-desktop` from there).

SmartManoj:
Reopen !

adrav:
I also need this!

PR #2363, opened 2016-08-18
Create newest qt creator doc for ubuntu 16.04

sujunmin:
1. Qt Creator 4.0.3 on Ubuntu 16.04
2. Change Font Settings for Building

### DIFF
--- a/doc/building-qtcreator-16.04.md
+++ b/doc/building-qtcreator-16.04.md
@@ -1,0 +1,146 @@
+##Build instructions for Qt Creator 4.0.3 under Ubuntu 16.04
+
+###Prepare
+
+* Read [Build instructions for Qt Creator 3.5.1 under Ubuntu 12.04](https://github.com/telegramdesktop/tdesktop/blob/master/doc/building-qtcreator.md) first, and the following instructions is modified by it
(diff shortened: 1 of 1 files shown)


john-preston:
I think there should be single doc for building in QtCreator that should not depend on QtCreator version or system version.

The build instruction docs are prepared just for building the software the way it is in the official builds, not for describing every possible way of modifying the source code.

sujunmin:
This document is not only for newest version for qtcreator, becuase some dev packages are fulfilled for building, no need to download original source codes or dev kits.

The last paragraph is for all user which need to change font, and by rebuilding the codes can easily to change fonts, I think it is so important to rebuild codes.

So please merge the changing fonts to your doc system before there is a place to  change font settings.

stek29:
Why won't you also add intructions on how to change colors/sizes? Or icons?
And don't forget about localization files.
Oh, and about changing date format too.

sujunmin:
Yes, because there are more issues for font changing and render bad, and there is no plan for this project to solve them, so this is my opinion to raise this problem to solve this problem temporarily.

stek29:
@sujunmin It's better to write everything there, in those issues' comments.

PR #2457, opened 2016-09-25
Linux: Make Telegram config folder XDG compatible

cookiengineer:
This is a very small change, but it helps improve Linux compatibility very much.

The XDG specification recommends using `.config/<program name>` and `.cache/<program name>` as its folder structure. This means that the `~/.TelegramDesktop` folder is violating that specification, leading to errors with pretty much all system-integrated tools that do management, cleanup or further integration of those files.

I changed the configuration folder to `~/.config/telegram-desktop`, assuming that telegram-desktop is still the name of the program judging from Windows App Id "TelegramDesktop".

However, XDG would suggest that `telegram` binary opens the config for `telegram`, not for `telegram-desktop` - meaning that it is up for discussion if it makes sense to rename the binary to just Telegram or leave it as-is.

These two changes affect only these files:

- _other/update_linux.cpp
- pspecific_linux.cpp

Tested on ArchLinux, known to work perfectly on Debian Testing, too. Tested with up-to-date GNOME3, MATE (v1.10 with GTK 2.8) and XFCE4.

edit:// Just for the sake of completeness: The [XDG user dirs article on freedesktop](https://www.freedesktop.org/wiki/Software/xdg-user-dirs/)

### DIFF
--- a/Telegram/SourceFiles/_other/updater_linux.cpp
+++ b/Telegram/SourceFiles/_other/updater_linux.cpp
@@ -366,7 +366,7 @@ int main(int argc, char *argv[]) {
                     writeLog("No workdir, trying to figure it out");
                     struct passwd *pw = getpwuid(getuid());
                     if (pw && pw->pw_dir && strlen(pw->pw_dir)) {
-                        string tryDir = pw->pw_dir + string("/.TelegramDesktop/");
+                        string tryDir = pw->pw_dir + string("/.config/telegram-desktop/");
                         struct stat statbuf;
                         writeLog("Trying to use '%s' as workDir, getting stat() for tupdates/ready", tryDir.c_str());
                         if (!stat((tryDir + "tupdates/ready").c_str(), &statbuf)) {
--- a/Telegram/SourceFiles/pspecific_linux.cpp
+++ b/Telegram/SourceFiles/pspecific_linux.cpp
@@ -325,7 +325,7 @@ namespace {
 
 QString psAppDataPath() {
     QString home(_psHomeDir());
-    return home.isEmpty() ? QString() : (home + qsl(".TelegramDesktop/"));
+    return home.isEmpty() ? QString() : (home + qsl(".config/telegram-desktop/"));
 }
 
 QString psDownloadPath() {


cookiengineer:
Damn it, forgot to signoff - hang on a sec

auchri:
Please include the signature to the commit message: https://github.com/telegramdesktop/tdesktop/blob/master/.github/CONTRIBUTING.md#sign-your-work

cookiengineer:
Added signoff line to commit message.

john-preston:
@cookiengineer If I merge such pull request everyone will be signed out with such update, which is not acceptable. I need to migrate the user data from old place to the new.

cookiengineer:
So, what's the suggested solution to achieve this? Is there any kind of infrastructure existing already?

Lesik:
This is a very important change. Thanks for your initiative, @cookiengineer.

z3ntu:
@cookiengineer Shouldn't the application honor `$XDG_CONFIG_HOME` according to the standard?

cookiengineer:
@z3ntu Preferably yes, but please see the comment of @john-preston - it won't be merged in anyways. So I guess it's up to me or someone else to fork tdesktop to get these XDG fixes in.

john-preston:
@cookiengineer It can be done with migrating the existing users. What folder instead of ~/.TelegramDesktop should be used?

z3ntu:
@john-preston `~/.config/telegram-desktop`

john-preston:
@z3ntu You mean $XDG_CONFIG_HOME/telegram-desktop ? Is it fine to use it even for cache of images? Because splitting the location for images, settings and auth data won't be done soon.

z3ntu:
Cached data that isn't precious (so could be lost every reboot or spontaniously) should be located in `~/.cache` or `XDG_CACHE_HOME`. 

Back to the config directory, the environment variable will be not set most likely, where it should default to `~/.config` but if it's set, it should be used instead of `~/.config`. But yes `$XDG_CONFIG_HOME/telegram-desktop` seems correct.

Lesik:
I would like to emphasize that `XDG_CONFIG_HOME` should _not_ be used for cache! The `XDG_CONFIG_HOME` is the folder which Linux users backup when they want to preserve their personal configurations during a system upgrade (for example). If every application dumped their cache in their that would mean unnecessary waste of precious backup space. Some applications, for example Cutegram (alternative Telegram client) wrongly write their cache to `XDG_CONFIG_HOME`, which is 1) not XDG compliant and 2) super annoying and simply the wrong thing to do.

john-preston:
@Lesik What about data? (like local databases). Just to understand.

Well, there is no way to easily migrate to the new location then, because currently tdesktop doesn't divide config / settings / local image / file / voice message storage. So we can't migrate easily all the tdesktop data from ~/.TelegramDesktop neither to XDG\_CONFIG\_HOME nor to XDG\_CACHE\_HOME (because it contains both important data and cache). So this pull request better not to be merged (or all users who heavily backup their ~/.config can become very disappointed).

z3ntu:
So stuff which contains settings or user tokens which could be transferred to a new machine should reside in `.config` but for example the `DebugLogs` directory should definitely not be there. `.local/share/` (=`XDG_DATA_HOME`) should *probably* be used for that.

So in short: Config stuff that doesn't change the whole time into `.config` and the rest into `.local/share` as far as I understood it correctly.

john-preston:
@z3ntu Well, DebugLogs is not that interesting, it can be moved to cache indeed. But "tdata" currently contains both critical data (user authorization key, settings etc) and huge amount of images / voice messages cache (could be tens or hundreds MB). So you can't place it to cache (or auth will be lost for those who purge it) and you can't place it to confit (or users who back up it frequently will freak out). What's left? Use XDG_DATA_HOME=~/.local/share/ for all of them?

stek29:
@john-preston store crucial files and caches in different folders?
And assign those two folders one value for other OSes.

z3ntu:
I would also suggest splitting it if it's somehow possible.

john-preston:
@z3ntu It is not easy for the existing users that will autoupdate to the new version.

stek29:
@john-preston why splitting and moving is harder than just moving?

john-preston:
@stek29 Because it requires full read before moving.

ManuelSchneid3r:
> But yes $XDG_CONFIG_HOME/telegram-desktop seems correct.

Partially. The variables may be empty and implementation have to set fallbacks which are covered by the spec as well. 

While reading I permanently asked myself why John hast to ask all this questions. Finally I realized that the specification you are talking about has never been linked. Its not that much to read. (Less than this issue). 

[XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)

john-preston:
@ManuelSchneid3r Well, also I ask them because currently it is not easy to follow this specification right now, still maintaining backward compatibility, the pull request was just partially following it and I was interested is there any sense in partially following it in some of the upcoming releases (easy to do). But it looks like it will be a wrong idea:

— if XDG\_CONFIG\_HOME is used for all the data, then all users who heavily backup their ~/.config can become very disappointed.
— if XDG\_CACHE\_HOME is used for all the data, then all users who clean it will be logged out which is even worse.

If all the data (both critical auth data and cache of images etc) is moved to XDG\_DATA\_HOME — will it be better than the current solution? Or should it wait till I'll be able to move different types of data to different folders?

z3ntu:
I think `XDG_DATA_HOME` (`~/.local/share`) is still better than the current solution but splitting the files to different directories would be the optimal solution.

hobarrera:
> I think XDG_DATA_HOME (~/.local/share) is still better than the current solution but splitting the files to different directories would be the optimal solution.

Not *much* better*. It avoids littering my home (I really find too many directories in my home annoying, since it's where all my own stuff is), but won't get backed up, and has an uncleanable cache.

john-preston:
Now we should use $XDG_DATA_HOME/TelegramDesktop if there is nothing in $HOME/.TelegramDesktop (if there is we still use the old data location).

PR #2489, opened 2016-10-02
Added pin button on the title

nop144d:
Introduced PinBtn class responsible for window pinning functionality which
is implemented using setWindowFlags function from Qt. Added sysPin style
placeholder in basic.style, so it can be used when the icon for the button
will be ready in sprites.

Signed-off-by: Vahan Stepanyan nop144d@gmail.com (github: nop144d)


nop144d:
Is there a way to look at tests' logs?

nop144d:
Adjusted my changes to reconcile with commit 376941d:
Mixed channel access hash to the game score encrypted data

nop144d:
Just in case, this is how it looks:
telegram-pin-clear

jhasse:
IMHO this should rather be hidden in a context menu or something. Most users probably don't need this feature very often or even don't understand what it means.

Have you tried something like http://www.skybn.com/window-always-on-top/ ?

ADTC:
How about holding Alt while hovering on Minimize or Maximize? The button would change to Pin.

BTW, do you have two separate Pin icons for on/off states?

nop144d:
@jhasse Can't say about most users, but I used to have the chat window at the right bottom corner sized enough, that I can both chat with someone and read something in browser window which is typically maximized. And it is very annoying that I have to switch between browser and chat back and forth to reply in the chat and then continue reading. I think this is very common. Now, having a separate software for pinning - which I agree is not the thing that you do with every window - is kind of wasteful in terms of resources, though I use TopMost (https://sourceforge.net/projects/topmost/) with a software that lacks this functionality that can be implemented in 3 lines of code. Finally, it requires more skills to find, install and use this kind of software than getting used to what that button near the minimize button does. So, hiding it in the context menu makes this feature completely useless.

@ADTC What you suggest is complex both in implementing and using the feature. I would never guess that there is a magic combination of key and mouse that will let me pin the window.
Currently, I use the same icon for pinned and unpinned states, but can make a different one for unpinned state when I get updated about the status of this PR.

jhasse:
On Linux (e. g. MATE) this is a context menu option (for every window). I've checked MPC-HC and it has that option in the normal menu, rather hidden (View -> On Top -> Always). That's why I think a dedicated button is a little bit unusual.

nop144d:
@jhasse I see. Maybe `#ifdef Q_OS_WIN` then?

jhasse:
AFAIK client-side window decorations are Windows only anyway.

delminskii:
Most of (if not every) linux WMs has this feature as built-in. For example, here's a screen for Openbox window manager:

nop144d:
@delminskii ok, I got that, but Windows still does not have it and I'm not sure about Mac. But the second argument (which is more important) is that when I was using my patch for a while, I noticed, that having the window pinned and notifications enabled at the same time is inconvenient, you want only one of them. Thus, there should be a way of automatically disabling notifications when window is pinned and enable them back (if needed) when unpinned or something very similar to that. Window Managers will never do this.

nop144d:
@delminskii BTW, just noticed that the text in your Telegram from the screenshot says the same thing ;) Nevertheless, my comment may be helpful for those who don't understand Russian.

PR #2517, opened 2016-10-10
added zip download link

Kolahzary:
https://github.com/telegramdesktop/tdesktop/archive/master.zip


john-preston:
You need to clone from git so many things, that I'd rather remove this download-in-zip completely -- more clean docs with just "git clone" for everything.

stek29:
> Download in ZIP

https://youtu.be/umDr0mPuyQc

PR #2527, opened 2016-10-13
Fixed Markdown parsing issues in build documentation for Windows

worm202:
The issue is that Markdown parser doesn't handle code blocks after lists right. Instead of putting code into a block, it is shown as regular unformatted text. To resolve this issue, the empty HTML comments were added between lists and code blocks.

Signed-off-by: Sergey Filatov <raxp.worm202@gmail.com> (github: worm202)


john-preston:
Thanks!

PR #2553, opened 2016-10-21
Update updater.cpp

IMAN4K:


john-preston:
What's the point, why is that important?

IMAN4K:
Just for better reading log files nothing more!
(`'\n'` actually does nothing)

john-preston:
I get completely readable logs with '\n'. Maybe you're using notepad to view them, try something like https://www.sublimetext.com/2 

Anyway, the signature should be in the pull request commit message, not in here.

PR #2555, opened 2016-10-21
Update updater.cpp

IMAN4K:
Signed-off-by: Iman iman72411@yahoo.com (github: IMAN4K)


auchri:
I think it's better do not introduce carriage returns here and instead @IMAN4K should instead use a proper editor to view the logs (e.g. sublime or notepad++)

auchri:
I think it's better do not introduce carriage returns here and instead @IMAN4K should instead use a proper editor to view the logs (e.g. sublime or notepad++)

PR #2659, opened 2016-11-22
main-window-linux: parse XDG_CURRENT_DESKTOP as colon separated list

3v1n0:
As per [freedesktop specifications](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html), $XDG_CURRENT_DESKTOP is a "colon-separated list of strings". Thus we need to parse it in that way.

Various desktops are now exporting this variable exposing multiple flavors.


john-preston:
Thanks!

PR #2660, opened 2016-11-22
main-window-linux: try to use AppIndicator in Unity

3v1n0:
AppIndicator is the native format for indicators in Ubuntu unity shell, since telegram supports them, it's better to rely on the system library for that. As it also allows better integration with [snap packages](http://snapcraft.io).

This PR has https://github.com/telegramdesktop/tdesktop/pull/2659 as prerequisite.

### DIFF
--- a/Telegram/SourceFiles/platform/linux/main_window_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/main_window_linux.cpp
@@ -370,7 +370,7 @@ bool MainWindow::psHasNativeNotifications() {
 void MainWindow::LibsLoaded() {
 	QStringList cdesktop = QString(getenv("XDG_CURRENT_DESKTOP")).toLower().split(':');
 	noQtTrayIcon = (cdesktop.contains(qstr("pantheon"))) || (cdesktop.contains(qstr("gnome")));
-	tryAppIndicator = cdesktop.contains(qstr("xfce"));
+	tryAppIndicator = (cdesktop.contains(qstr("xfce")) || cdesktop.contains(qstr("unity")));
 
 	if (noQtTrayIcon) cSetSupportTray(false);
 


john-preston:
Please rebase your commit because I've merged the previous pull request.

Also, I'm not sure this is the right thing to do. I try to minimize the amount of dependencies and I force usage of libappindicator over Qt tray icon support only if Qt tray icon is known to be buggy. This was the case some time before, but right now Qt tray icon works fine in Unity (as far as I know), so I prefer to use it (the existing dependency) instead the other one (libappindicator). Is it known to work better in any way? From my experience they work exactly the same.

3v1n0:
> Please rebase your commit because I've merged the previous pull request.

Done.

> Also, I'm not sure this is the right thing to do. I try to minimize the amount of dependencies and I force usage of libappindicator over Qt tray icon support only if Qt tray icon is known to be buggy. This was the case some time before, but right now Qt tray icon works fine in Unity (as far as I know), so I prefer to use it (the existing dependency) instead the other one (libappindicator). Is it known to work better in any way? From my experience they work exactly the same.

As briefly written in the PR description, Ubuntu is going to use [snap packages](http://snapcraft.io/) more and more. These packages are using sandboxes and currently unity shell can't access to temporary files they produce if they're in `/tmp`, as it is in [this case](http://code.qt.io/cgit/qt/qtbase.git/tree/src/platformsupport/dbustray/qdbustrayicon.cpp?h=5.7#n69).
AppIndicator [is going to support](https://code.launchpad.net/~3v1n0/libappindicator/incons-paths-on-snap) such cases, while the Qt Tray icon would need some patching too.

Speaking of the indicator itself, app-indicator has the ability of pointing an actual icon with various sizes, thus it allows better integration when there's a mix of hidpi and normal displays.

In any case this dependency is already in the code, so if the library is there, why not to use it?

john-preston:
@3v1n0 Ok, I see. Currently my libappindicator support uses "app_indicator_set_icon_full" method and points it to a temp icon file created in app data folder.

Maybe you can tell me how can I provide the pixmaps with different sizes for different icon sizes (hidpi displays support)?

3v1n0:
> Maybe you can tell me how can I provide the pixmaps with different sizes for different icon sizes (hidpi displays support

If providing an SVG isn't feasible, the best way would be using AppIcon custom themes, you can register a folder as your app theme using `app_indicator_set_icon_theme_path` (in this case pointing to `$CACHE_DIR/temp-folder-name`) containing something like:
```
$CACHE_DIR/temp-folder-name/hicolor/64x64/apps/icon-file.png
$CACHE_DIR/temp-folder-name/hicolor/48x48/apps/icon-file.png
$CACHE_DIR/temp-folder-name/hicolor/32x32/apps/icon-file.png
$CACHE_DIR/temp-folder-name/hicolor/22x22/apps/icon-file.png
$CACHE_DIR/temp-folder-name/hicolor/16x16/apps/icon-file.png
```
Then you've only to set the app-icon-name to `"icon-file" `

You can see the gtk2/3 app indicator examples in this repo: https://github.com/3v1n0/indicators-examples-snaps. Something similar is also [here](http://bazaar.launchpad.net/~3v1n0/appmenu-qt5/icons-user-cache-on-snap/view/head:/src/iconcache.cpp#L130).

3v1n0:
@john-preston you need anything else on my side?

john-preston:
@3v1n0 Currently there is a long lasting release being developed, I think I'll merge this and test it when it will be close to being done.

auchri:
@john-preston 👍

john-preston:
Thanks.

PR #2815, opened 2017-01-12
change more idiomatic ko

yangbeom:
change more idiomatic ko '데이트를 다운로드 중' -> '업데이트 다운로드 중'
change missspelling '정ㅂ' -> '정보'


stek29:
Translations are maintained on Transifex.
But maybe @john-preston can forward this to KO translators.

PR #2936, opened 2017-01-28
[WIP] Improvements of emoji panel

mymedia2:
The emoji panel very annoys me if it pops out when I'm trying to go to a last message in a chat or to cancel answering. So I removed the possibility of opening this panel by hovering mouse on attach button (the icon with smile on the left of the text input field). You can open the list of emoji or stickers by clicking on this icon or pressing tabulation on keyboard, and now you can choose a sticker using keyboard.

Also I simplified sending several stickers from the same pack at a time. The panel isn't hiding when you sent a sticker using mouse. But it automatically hides if you use keyboard for sending.

But the work is still not finished. I'd like to implement storing current selected section of the panel between launches the application and displaying on attach icon which section would be opened, just like on Android.

And maybe I'll make a some refactoring of `EmojiPan`, `FieldAutocomplete` classes.

**Work in progress…**

### DIFF
--- a/Telegram/SourceFiles/historywidget.cpp
+++ b/Telegram/SourceFiles/historywidget.cpp
@@ -3191,7 +3191,8 @@ HistoryWidget::HistoryWidget(QWidget *parent) : TWidget(parent)
 	_silent->hide();
 	_botCommandStart->hide();
 
-	_attachEmoji->installEventFilter(_emojiPan);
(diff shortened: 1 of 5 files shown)


auchri:
Please create a new pull request after you finished your work.

mymedia2:
Okay, I'll just push here, and then later, I'll squash all further commits and create new PR

PR #2959, opened 2017-02-02
Edit Korean translation

2DU:
"Delete all from this user" is not "모두에게 메시지 삭제"

"Delete all from this user" is "이 사용자의 모든 메세지 삭제"


stek29:
Translations are managed on Transifex, sorry.

2DU:
I have to inquire there. Okay.

PR #2986, opened 2017-02-08
Update lang_it.strings

f4nu:
Fixed a little typo.


auchri:
Langs are managed on transifex only, sry. Maybe @john-preston can forward it to the translator.

f4nu:
OK, thank you!

PR #2995, opened 2017-02-10
Added missing FileDescrtiption in winrc file for main .exe

LuKePicci:
Following the fix for URI associations it turns out that the only field Windows uses to display a friendly app name and not the exe filename is the FileDescription resource string, actually missing from released binaries. Given that no RC generation script is present in this repo I simply hardcoded the "Telegram Desktop" as such description. Updater is done already the same way.


john-preston:
@LuKePicci It is there in dev branch already. I'll change my long description to a shorter one, but.. it is there in some way.

LuKePicci:
Just seen, I double checked that string was the cuprit, and it is. Yes it definetely needs to be shorter.

PR #3014, opened 2017-02-13
small german translation correction

Emer0ld:
changed the german translation of "unpin". The old one could be easily mistaken with deleting the chat. Now it says "don't pin chat any longer".


auchri:
Langs are managed on transifex only, sry. Maybe @john-preston can forward it to the translator.

PR #3016, opened 2017-02-13
Fix compilation of history widget.

woutershep:
error: inconsistent deduction for 'auto': 'int' and then 'auto'
  auto start = 0, end = list.size();
    ^~~~

### DIFF
--- a/Telegram/SourceFiles/historywidget.cpp
+++ b/Telegram/SourceFiles/historywidget.cpp
@@ -190,7 +190,8 @@ namespace {
 // is applied once for blocks list in a history and once for items list in the found block
 template <bool TopToBottom, typename T>
 int binarySearchBlocksOrItems(const T &list, int edge) {
-	auto start = 0, end = list.size();
+	// static_cast to work around GCC bug #78693
+	auto start = 0, end = static_cast<int>(list.size());
 	while (end - start > 1) {
 		auto middle = (start + end) / 2;
 		auto top = list[middle]->y;


stek29:
Isn't it theoretically possible to get `start` and `end` of different types?
Afaik `auto start = 0` would deduct to `int`, but `list.size()` may (and should) be `unsigned`.

woutershep:
The function returns an int, and only returns start, which will be an int, so i figured that this wouldn't be a problem, if your size cant fit in int you're going to have a problem no matter what type start and end are.

john-preston:
@woutershep I'm curious where doesn't it compile (what compiler version and Qt version)? Because it is used with T=QList<> and with T=QVector<> and they both return int in their size() method (so it works now in VS2015, Xcode 8.2, GCC 4.9).

If you need to fix it is better to cast the result of size() call to int (by using `int(list.size())` or `static_cast<int>(list.size())`), because `start` and `end` are supposed to have the same type.

woutershep:
Using GCC 6.3.0 and qt-5.6.2

Looking closer into it it appears to be a regression in GCC 6 actually:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78693

Now the question is, do you want to work around GCC 6 bugs in your code, if so id be happy to change it to a static_cast with a comment above it that its needed for a GCC 6 quirk? If not i guess you can just close this pr as invalid.

john-preston:
@woutershep I can merge a pr with `static_cast` if it'll make it easier for someone. I think it should be there until a fixed version of GCC is released.

john-preston:
@woutershep please rebase your changes on top of the current master branch so I could merge it with linear history:

https://github.com/telegramdesktop/tdesktop/blob/master/.github/CONTRIBUTING.md#pull-upstream-changes-into-your-fork-regularly

and it could be merged after that.

Issue #361, opened 2014-12-26
Hashtag search is hidden in single column layout


ADTC:
If I click on a hashtag in the double-column layout (in a message in the right column), the hashtag appears in the search box of the left column, and the search results appear below it.

But in single-column layout, if I click on a hashtag in a message, nothing *appears to happen*. But when I click on the Back link at the top, I find that the search had indeed happened in the Chats list view, but was hidden from view.

In single-column layout, clicking on a hashtag should *not only* search for it, **but also** switch to the search results (Chats view) automatically.

telegramdesktop:
Fixed in 0.7.13.

PR #3062, opened 2017-02-23
Added interpreter path

rezasp:
Signed-off-by: MohammadReza Espargham <reza.espargham@owasp.org> (github: rezasp)


auchri:
Duplicate #361

PR #3109, opened 2017-03-11
Update .desktop file on Linux

mymedia2:
What I have done:

 * Removed deprecated _Encoding_ key
 * Specified correct WM class for _StartupWMClass key_
 * Fixed icon name in `lib/xdg/telegramdesktop.desktop`

The _Encoding_ key is deprecated as [the specification says][1].

A WM class, which was specified, represents the name of binary and depends on its file name. Qt sets this class equal to the _ApplicationName_, which you specified in [Telegram/SourceFiles/main.cpp][2]. To show all WM classes use **xprop** utility.

What happens if you specify incorrect _StartupWMClass_: then when you click on Telegram menu item, you'll see two Telegram icon on the taskbar. And one of them will disappear after a while.

Screenshot of the taskbar

---

I also fixed restarting when language is changed and there is no Updater near Telegram executable file. By the way, you need not have a special Updater on Linux. Telegram Desktop can update itself within a single process. When a new version is downloaded, just rename it to `gExeName` and do restart. You can rename any file (if you certainly have permissions) even if it's opened.

Besides, you can find all valid values for the _Category_ key in [annex to menu specification][3].

 [1]: https://standards.freedesktop.org/desktop-entry-spec/1.1/apc.html
 [2]: https://github.com/telegramdesktop/tdesktop/blob/83720d87898e62e4e8ef98d2e410b931a9328562/Telegram/SourceFiles/main.cpp#L29
 [3]: https://specifications.freedesktop.org/menu-spec/latest/apa.html

### DIFF
--- a/Telegram/SourceFiles/platform/linux/specific_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/specific_linux.cpp
@@ -467,16 +467,16 @@ void psRegisterCustomScheme() {
 			QTextStream s(&f);
 			s.setCodec("UTF-8");
 			s << "[Desktop Entry]\n";
-			s << "Encoding=UTF-8\n";
 			s << "Version=1.0\n";
 			s << "Name=Telegram Desktop\n";
 			s << "Comment=Official desktop version of Telegram messaging app\n";
+			s << "TryExec=" << EscapeShell(QFile::encodeName(cExeDir() + cExeName())) << "\n";
 			s << "Exec=" << EscapeShell(QFile::encodeName(cExeDir() + cExeName())) << " -- %u\n";
 			s << "Icon=telegram\n";
 			s << "Terminal=false\n";
-			s << "StartupWMClass=Telegram\n";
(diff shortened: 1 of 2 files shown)


auchri:
Please rebase to latest `dev`.

john-preston:
@mymedia2 I'm afraid a bit to make changes to lib/xdg/telegramdesktop.desktop file, because I don't know quite well how it is used, it was added in a PR #1648 by @hobarrera and I don't know, will it break something for its users when I change the `Exec` and `Icon` values in it. For example for users who've placed their own icons somewhere with the old name and they've been used until now.

auchri:
Maybe @hobarrera can review this PR and leave some comments ☺

hobarrera:
The changes to `lib/xdg/telegramdesktop.desktop` look okay by my.

You might want to mention that the icon was renamed in the changelog (packages will want to know this). Users using custom icons can see that too. I'm in favour of both `.desktop` files using the same name, because otherwise icon theme authors won't know which name to target.

Issue #1586, opened 2016-02-09
File name as time sent not downloaded


PaulNerd:
Hi,
this is very inconvenient to me when I'm trying to back everything up and sort it.
The file name of files sent with Telegram (except uncompressed ones), like images and audio, is the timestamp of when it was downloaded (e.g. photo_2016-02-09_12-16-46 now), even if it happened months after it was originally included in the conversation.
This can lead to huge mixups and it's time-consuming for me to rename every single file.

I'd ask to change that so that the file name reflects the original place in the conversation.
Thanks.

telegramdesktop:
@PaulNerd This seems right, I'm not sure when this will be changed, but the sent time is definitely a better name.

PaulNerd:
Thanks
it should be a pretty straightforward and fast change to make, I believe :-)

PaulNerd:
Sorry to bother you... any news on this one?
I'm not familiar with Github nor Telegram code, and I don't have the time to try and find the place where the filename string is formed, otherwise I could have given it a shot.

PR #3150, opened 2017-03-25
Use message date by default when saving a photo (#1586)

Lonami:


auchri:
@Lonami To get this PR merged, you have to fix the tests.

Lonami:
> you have to fix the tests.

I'm not sure what really failed though? I checked the logs and couldn't find anything relevant. The change is so small I don't understand why that one build failed.

john-preston:
@Lonami click on "Details" link and you'll see the appveyor build log. It has warning (that is treated as an error):

"c:\tbuild\tdesktop\telegram\sourcefiles\mediaview.cpp(788): warning C4003: not enough actual parameters for macro 'qsl'"

because of a "qsl()" expression. For an empty string you should use "QString()" instead.

Lonami:
> because of a "qsl()" expression. For an empty string you should use "QString()" instead.

I see now, that's what I thought which could be the reason. Thank you both! I'll fix it tomorrow and `git commit` a squash.

Aokromes:
@Lonami you don't need to squash anymore, now github allows to squash on merge :)
(unless it's TC contribution requirement)

Lonami:
> now github allows to squash on merge

I know, I often use it, but I forgot to add the "Signed off" thing to the commit message, so I simply `squash`'ed the commits to keep the original message.

PR #3201, opened 2017-04-04
Update libva git URL and QT init-repository invocation

felfert:
Signed-off-by: Fritz Elfert <fritz-github@fritz-elfert.de> (github: felfert)

When attempting to build on my own travis account, those two changes were necessary, because I did not yet had dependencies cached and therefore full builds were necessary.
 - The libva repo has moved.
 - The invocation of the init-repository script in the Qt repo has changed.

### DIFF
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -205,7 +205,7 @@ buildVa() {
   rm -rf *
 
   cd "$EXTERNAL"
-  git clone git://anongit.freedesktop.org/git/libva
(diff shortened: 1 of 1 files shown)


felfert:
Well. the "failed" travis check is **not** due to changes but due to travis being extremely slow.
In fact, my PR would affect the travis build only, if you guys would delete the cache.
All logs end with:
````
The job exceeded the maximum time limit for jobs, and has been terminated.
````

PR #3203, opened 2017-04-04
Add appstream metadata file

grulja:
Add appstream metadata file [1] providing basic information about telegram desktop app. I would also prefer if there is some official screenshot rather than the one I'm currently hosting, but that can be changed later. The most important thing is to agree on appstream app ID, which can be then used to identify telegram desktop app, e.g. when writing reviews in ODRS [2].

[1] - https://www.freedesktop.org/wiki/Distributions/AppStream/
[2] - https://odrs.gnome.org/

### DIFF
--- a/lib/xdg/telegramdesktop.appdata.xml
+++ b/lib/xdg/telegramdesktop.appdata.xml
@@ -8,6 +8,10 @@
     <description>
        <p>Fast and secure desktop app, perfectly synced with your mobile phone.</p>
     </description>
+    <categories>
+        <category>Network</category>
+        <category>InstantMessaging</category>
+    </categories>
     <url type="homepage">https://desktop.telegram.org/</url>
     <url type="bugtracker">https://github.com/telegramdesktop/tdesktop/issues</url>
     <screenshots>
--- a/lib/xdg/telegramdesktop.appdata.xml
+++ b/lib/xdg/telegramdesktop.appdata.xml
@@ -2,13 +2,13 @@
 <component type="desktop">
     <id>org.telegram.TelegramDesktop.desktop</id>
     <metadata_license>CC0-1.0</metadata_license>
-    <project_license>GPL-2.0+</project_license>
-    <name>Telegram desktop</name>
+    <project_license>GPL-3.0</project_license>
+    <name>Telegram Desktop</name>
     <summary>Telegram Desktop messenger</summary>
     <description>
        <p>Fast and secure desktop app, perfectly synced with your mobile phone.</p>
     </description>
-    <url type="homepage">https://github.com/telegramdesktop/tdesktop</url>
+    <url type="homepage">https://desktop.telegram.org/</url>
     <url type="bugtracker">https://github.com/telegramdesktop/tdesktop/issues</url>
     <screenshots>
       <screenshot type="default">
--- a/lib/xdg/telegramdesktop.appdata.xml
+++ b/lib/xdg/telegramdesktop.appdata.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<component type="desktop">
+    <id>org.telegram.TelegramDesktop.desktop</id>
+    <metadata_license>CC0-1.0</metadata_license>
+    <project_license>GPL-2.0+</project_license>
+    <name>Telegram desktop</name>
+    <summary>Telegram Desktop messenger</summary>
+    <description>
+       <p>Fast and secure desktop app, perfectly synced with your mobile phone.</p>
+    </description>
+    <url type="homepage">https://github.com/telegramdesktop/tdesktop</url>
+    <url type="bugtracker">https://github.com/telegramdesktop/tdesktop/issues</url>
+    <screenshots>
+      <screenshot type="default">
+        <image>https://jgrulich.fedorapeople.org/telegram/telegram-screenshot.png</image>
+      </screenshot>
+    </screenshots>
+    <provides>
+        <binary>telegram-desktop</binary>
+    </provides>
+</component>
--- a/lib/xdg/telegramdesktop.appdata.xml
+++ b/lib/xdg/telegramdesktop.appdata.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <component type="desktop">
-    <id>org.telegram.TelegramDesktop.desktop</id>
+    <id>org.telegram.tdesktop.desktop</id>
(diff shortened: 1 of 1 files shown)


lib/xdg/telegramdesktop.appdata.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<component type="desktop">
+    <id>org.telegram.TelegramDesktop.desktop</id>
+    <metadata_license>CC0-1.0</metadata_license>
+    <project_license>GPL-2.0+</project_license>
+    <name>Telegram desktop</name>
```
auchri: `D` esktop

lib/xdg/telegramdesktop.appdata.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<component type="desktop">
+    <id>org.telegram.TelegramDesktop.desktop</id>
+    <metadata_license>CC0-1.0</metadata_license>
+    <project_license>GPL-2.0+</project_license>
```
auchri: We use GPLv3 with OpenSSL exception

lib/xdg/telegramdesktop.appdata.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<component type="desktop">
+    <id>org.telegram.TelegramDesktop.desktop</id>
```
auchri: Where do you got this id from?

lib/xdg/telegramdesktop.appdata.xml
```diff
@@ -0,0 +1,21 @@
...
+    <metadata_license>CC0-1.0</metadata_license>
+    <project_license>GPL-2.0+</project_license>
+    <name>Telegram desktop</name>
+    <summary>Telegram Desktop messenger</summary>
+    <description>
+       <p>Fast and secure desktop app, perfectly synced with your mobile phone.</p>
+    </description>
+    <url type="homepage">https://github.com/telegramdesktop/tdesktop</url>
```
auchri: I think this should be https://desktop.telegram.org/

lib/xdg/telegramdesktop.appdata.xml
```diff
@@ -0,0 +1,21 @@
...
+    <url type="bugtracker">https://github.com/telegramdesktop/tdesktop/issues</url>
+    <screenshots>
+      <screenshot type="default">
+        <image>https://jgrulich.fedorapeople.org/telegram/telegram-screenshot.png</image>
```
auchri: We definitely want some other screenshot here

lib/xdg/telegramdesktop.appdata.xml
grulja: The id is usually in form of org.organization_name.app_name.desktop so I created it that way.

lib/xdg/telegramdesktop.appdata.xml
grulja: Ok, will fix it.

grulja: I used GPL-3.0 as GPL-3.0+ is deprecated and there is no GPL-3.0 with OpenSSL description in https://spdx.org/licenses/.

lib/xdg/telegramdesktop.appdata.xml
auchri: Ok, I would suggest `org.telegram.tdesktop.desktop`.

auchri:
Btw. you have to add the signature to your commits:
> Please read our [contributing policy](https://github.com/telegramdesktop/tdesktop/blob/master/.github/CONTRIBUTING.md). You'll need to make a pull request with the "Signed-off-by:" signature being the last line of your commit message, like it is described in [sign your work](https://github.com/telegramdesktop/tdesktop/blob/master/.github/CONTRIBUTING.md#sign-your-work) section. That will grant your work into the public domain.

grulja:
Commits updated with the signature.

john-preston:
We use `org.telegram.desktop` for macOS app bundle, I think it should fit here as well.

lib/xdg/telegramdesktop.appdata.xml
```diff
@@ -0,0 +1,21 @@
...
+    <name>Telegram Desktop</name>
+    <summary>Telegram Desktop messenger</summary>
+    <description>
+       <p>Fast and secure desktop app, perfectly synced with your mobile phone.</p>
+    </description>
+    <url type="homepage">https://desktop.telegram.org/</url>
+    <url type="bugtracker">https://github.com/telegramdesktop/tdesktop/issues</url>
+    <screenshots>
+      <screenshot type="default">
+        <image>https://jgrulich.fedorapeople.org/telegram/telegram-screenshot.png</image>
```
auchri: Please use `https://raw.githubusercontent.com/telegramdesktop/tdesktop/dev/docs/assets/preview.png` as screenshot

lib/xdg/telegramdesktop.appdata.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<component type="desktop">
+    <id>org.telegram.tdesktop.desktop</id>
```
auchri: Use `org.telegram.desktop` as suggested by @john-preston

grulja: Then it would be against specification, as you can see here [1], the id has to be in form of {tld}.{vendor}.{product}, which in this case is org.telegram.tdesktop and .desktop and the end is used to identify that it's a desktop app.

[1] https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

auchri: > To increase the uniqueness and to distinguish between different pieces of a software suite, it is suggested to append the type name to the component-id in these cases. For example, one can use com.hugski.ColorHug2 for the client tools to control hardware, and com.hugski.ColorHug2.firmware for the runtime firmware files.

So I think `org.telegram.desktop` is not against the specification.

TLD: org
Vendor: Telegram
Product: Desktop

auchri:
Sorry for the many comments, but I think a `category` should also be added.

> This tag can contain one or more <category>> entries, describing the categories this software component is associated with. 

````
<categories>
​    <category>Network</category>
​    <category>InstantMessaging</category>
​</categories>
```

auchri:
@grulja  Please change the target branch to `dev` next time.

PR #3232, opened 2017-04-08
Spellcheсker: WIP

stek29:
## ToDo:
- Update gyp files && Update Instructions -- once finished
- Consider adding some dictionaries (bundle them into executable?) 
 Currently you have to make a `spell` folder in `tdata`, and place aff+dic files there.
 Some dictionaries are available at https://github.com/LibreOffice/dictionaries, but I'm not sure if they're are allowed to use. Aspell.net could be a good source too, but I'm not sure if Aspell's dictionaries are compatible with hunspell.
- Move to enchant spelling lib as it supports a lot more backends, including system native backends
- Allow enabling/disabling particular dictionaries through settings (as well as disabling spellchecker completely)
- Add user dictionary support
- Disable spell check inside of ```` ``` ```` blocks
- It needs a bit smarter regexp, so words like `*'s, *n't` are not marked misspelled
- Don't check words at cursor, so user doesn't get ""Misspelled"" words while typing is in progress

### DIFF
--- a/Telegram/Resources/colors.palette
+++ b/Telegram/Resources/colors.palette
@@ -547,3 +547,6 @@ callBarFg: dialogsNameFgActive; // phone call bar text and icons
 importantTooltipBg: toastBg;
 importantTooltipFg: toastFg;
 importantTooltipFgLink: mediaviewTextLinkFg;
+
+// spellchecker
+spellUnderline: #ff0000 | attentionButtonFg;
--- a/Telegram/SourceFiles/chat_helpers/message_field.cpp
+++ b/Telegram/SourceFiles/chat_helpers/message_field.cpp
@@ -24,10 +24,8 @@ Copyright (c) 2014-2017 John Preston, https://desktop.telegram.org
 #include "base/qthelp_regex.h"
 #include "styles/style_history.h"
 #include "window/window_controller.h"
-#include "ui/widgets/popup_menu.h"
 #include "mainwindow.h"
 #include "auth_session.h"
-#include "chat_helpers/spellcheck.h"
 
 namespace {
 
@@ -97,10 +95,7 @@ TextWithTags::Tags ConvertEntitiesToTextTags(const EntitiesInText &entities) {
 }
 
 MessageField::MessageField(QWidget *parent, gsl::not_null<Window::Controller*> controller, const style::FlatTextarea &st, const QString &ph, const QString &val) : Ui::FlatTextarea(parent, st, ph, val)
-, _controller(controller)
-, _spellHighlighter(this)
-{
-
+, _controller(controller) {
 	setMinHeight(st::historySendSize.height() - 2 * st::historySendPadding);
 	setMaxHeight(st::historyComposeFieldMaxHeight);
 
@@ -154,41 +149,4 @@ void MessageField::insertFromMimeData(const QMimeData *source) {
 void MessageField::focusInEvent(QFocusEvent *e) {
 	FlatTextarea::focusInEvent(e);
 	emit focused();
-}
-
-void MessageField::contextMenuEvent(QContextMenuEvent *e) {
-	if (auto menu = createStandardContextMenu()) {
-		QTextCursor newTextCursor = cursorForPosition(e->pos());
-		newTextCursor.select(QTextCursor::WordUnderCursor);
-		auto codeBlocks = static_cast<ChatHelpers::CodeBlocksData *>(newTextCursor.block().userData());
-		if (codeBlocks != nullptr) {
-			bool inCode = false;
-			auto pos = newTextCursor.positionInBlock();
-			for (auto blk : codeBlocks->codeBlocks) {
-				if (pos - blk.len < blk.pos) {
-					if (pos > blk.pos) inCode = true;
-					break;
-				}
-			}
-			if (!inCode) {
-				QString word = newTextCursor.selectedText();
-
-				if (!ChatHelpers::SpellHelperSet::Instance().isWordCorrect(word)) {
-					menu->addSeparator();
-					for (auto &vec : ChatHelpers::SpellHelperSet::Instance().getSuggestions(word))
-						for (auto &suggestion : vec) {
-							menu->addAction(suggestion, [this, newTextCursor, suggestion]() {
-								QTextCursor oldTextCursor = textCursor();
-								setTextCursor(newTextCursor);
-								textCursor().clearSelection();
-								textCursor().insertText(suggestion);
-								setTextCursor(oldTextCursor);
-							});
-						}
-				}
-			}
-		}
-
-		(new Ui::PopupMenu(nullptr, menu))->popup(e->globalPos());
-	}
 }
--- a/Telegram/SourceFiles/chat_helpers/message_field.cpp
+++ b/Telegram/SourceFiles/chat_helpers/message_field.cpp
@@ -24,6 +24,7 @@ Copyright (c) 2014-2017 John Preston, https://desktop.telegram.org
 #include "base/qthelp_regex.h"
 #include "styles/style_history.h"
 #include "window/window_controller.h"
+#include "ui/widgets/popup_menu.h"
 #include "mainwindow.h"
 #include "auth_session.h"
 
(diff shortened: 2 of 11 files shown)


Telegram/Resources/colors.palette
```diff
@@ -507,3 +507,6 @@ mediaviewTransparentFg: #cccccc; // another transparent filling part
 
 // notification
 notificationBg: windowBg; // custom notification window background
+
+// spellchecker
+spellUnderline: #ff0000;
```
john-preston: All new colors should have a fallback value, so they will get some similar values in all themes that don't support them right now. It is done like `mainMenuCloudBg` key in this commit: https://github.com/telegramdesktop/tdesktop/commit/22d905b39c6d0f535748043c11c0da9b19c388a1

Telegram/SourceFiles/historywidget.cpp
```diff
@@ -2534,7 +2534,38 @@ void HistoryInner::onParentGeometryChanged() {
 MessageField::MessageField(HistoryWidget *history, const style::FlatTextarea &st, const QString &ph, const QString &val) : Ui::FlatTextarea(history, st, ph, val), history(history) {
 	setMinHeight(st::historySendSize.height() - 2 * st::historySendPadding);
 	setMaxHeight(st::historyComposeFieldMaxHeight);
+
+#ifdef TDESKTOP_ENABLE_SPELLCHECK
```
john-preston: I think if we get the spellchecking working so that it could be added to the deployed versions we won't need the ability to turn it off, maybe in Settings. Perhaps it will be required to be able to build without the hunspell, but in that case only it should be #ifndef-ed out and a fallback to some "null" spellchecking backend should be possible.

Telegram/SourceFiles/historywidget.cpp
```diff
@@ -2534,7 +2534,38 @@ void HistoryInner::onParentGeometryChanged() {
...
+	_spellHelperSet.addLanguages({"en_US"});
+	_spellHighlighter = new SpellHighlighter(this, &_spellHelperSet);
```
john-preston: Avoid naked `new`, use std::unique_ptr + std::make_unique / std::shared_ptr + std::make_shared for owned non-QObject derivatives and use object_ptr for owned QObject derivatives.

Telegram/SourceFiles/historywidget.h
```diff
@@ -29,6 +29,10 @@ Copyright (c) 2014-2017 John Preston, https://desktop.telegram.org
 #include "window/section_widget.h"
 #include "core/single_timer.h"
 
+#ifdef TDESKTOP_ENABLE_SPELLCHECK
+#include "spellcheck.h"
```
john-preston: I think spellcheck.cpp/h should be placed in `chat_helpers` subdirectory and `ChatHelpers` namespace. The message field would be moved there in the near future.

Telegram/SourceFiles/spellcheck.cpp
```diff
@@ -0,0 +1,91 @@
...
+ Copyright (c) 2014-2017 John Preston, https://desktop.telegram.org
+ */
+
+#ifdef TDESKTOP_ENABLE_SPELLCHECK
+#include "spellcheck.h"
+
+void SpellHelperSet::addLanguages(QStringList languages) {
+	for (const auto &lang : languages) {
+		if (_helpers.count(lang) == 0) {
+			auto helper = std::unique_ptr<HunspellHelper>(new HunspellHelper(lang.toLatin1()));
```
john-preston: Don't use naked `new`, use std::make_unique.

Telegram/SourceFiles/spellcheck.cpp
```diff
@@ -0,0 +1,91 @@
...
+			if (helper->isOpen()) {
+				_helpers.insert(std::make_pair(lang, std::move(helper)));
```
john-preston: Use `.emplace(lang, std::move(helper))`

Telegram/SourceFiles/spellcheck.cpp
```diff
@@ -0,0 +1,91 @@
...
+}
+
+void SpellHighlighter::highlightBlock(const QString &text) {
+	// If the theme was changed, color would be changed too 
+	_underlineFmt.setUnderlineColor(st::spellUnderline->c);
+	// \b - split the string into an alternating seq of non-word and word tokens
+	bool skip = true;
+	for (auto &ref : text.splitRef(QRegExp("\\b"))) {
+		if (!skip) {
+			if (!_helperSet->isWordCorrect(ref))
```
john-preston: Always use { and } for the if (...) body.

Telegram/SourceFiles/spellcheck.cpp
```diff
@@ -0,0 +1,91 @@
...
+				setFormat(ref.position(), ref.size(), _underlineFmt);
+				}
```
john-preston: Something went wrong with the alignment.

Telegram/SourceFiles/spellcheck.h
```diff
@@ -0,0 +1,98 @@
...
+#ifdef TDESKTOP_ENABLE_SPELLCHECK
+#include <hunspell/hunspell.hxx>
+
+class HunspellHelper {
+public:
+	HunspellHelper(const QByteArray &lang)
+	: _hunspell(nullptr)
+	, _codec(nullptr)
+	{
+		const QByteArray basePath = (cWorkingDir() + qsl("tdata/spell/") + lang).toUtf8();
```
john-preston: Prefer to use `auto` where it is possible in any place you declare a variable.

Telegram/SourceFiles/spellcheck.h
```diff
@@ -0,0 +1,98 @@
...
+		const QByteArray affPath = basePath + ".aff";
+		const QByteArray dicPath = basePath + ".dic";
+		if (QFileInfo(affPath).isFile() && QFileInfo(dicPath).isFile()) {
+			_hunspell.reset(new Hunspell(affPath, dicPath));
```
john-preston: Avoid naked `new`, use `= std::make_unique` instead.

Telegram/SourceFiles/spellcheck.h
```diff
@@ -0,0 +1,98 @@
...
+		for (auto &suggestion : suggestions) {
+			result.append(_codec->toUnicode(suggestion.data(), suggestion.length()));
+		}
+
+		return result;
+	}
+
+private:
+	std::unique_ptr<Hunspell> _hunspell;
+	QTextCodec *_codec;
```
john-preston: Add a newline after each block (public:, protected:, private:).

Telegram/SourceFiles/spellcheck.h
```diff
@@ -0,0 +1,98 @@
...
+class SpellHighlighter : public QSyntaxHighlighter {
+public:
+	SpellHighlighter(QTextEdit *textEdit, SpellHelperSet *helperSet)
+	: QSyntaxHighlighter(textEdit->document()) {
+		_helperSet = helperSet;
+		_underlineFmt.setFontUnderline(true);
+	}
+
+private:
+	void highlightBlock(const QString &text) final;
```
john-preston: Add a newline between methods and member variables in a class definition.

xvitaly:
Why not just use [Qtspell](https://github.com/manisandro/qtspell)?

stek29:
@xvitaly Qtspell is a wrapper for enchant spell checking library, and that wrapper won't fit to TDesktop.
But enchant itself looks interesting, thanks for sharing :)

superusercode:
You mean spellchecker, right?

stek29:
I'll try to get back to it after exams. Maybe it's better to close this for now.

clementhk:
It looks like Sonnet supports MacOS system spell check and that would mean a lot of Mac users.
Thanks for spending your time on this.

eirnym:
Why not to use native spell checking?

Aldo-f:
So a WIP is closed? 
Seems a bit weird.



> I'll try to get back to it after exams. Maybe it's better to close this for now.

Exams take a long while ;)

rx65m:
> Why not to use native spell checking?

How can we do that?

enzogian:
Hi, what happens to desktop version that does not allow spell check ??
please this is an extremely useful feature. tthanks

Issue #3192, opened 2017-04-01
Selected messages to forward are ignored when clicking on my profile picture


srazi:
### Steps to reproduce
1. Select some messages
2. Click on forward button to forward them
3. Click on sidebar icon
4. Now click on your profile picture

### Expected behaviour
By clicking on send button you should be able to forward messages to yourself.

### Actual behaviour
Send button and forward message status are disappeared an so you cant forward to yourself in this way.

### Configuration
**Operating system:**
Windows 10 x64
**Version of Telegram Desktop:**
1.0.27

auchri:
@srazi Just write yourself some message so that you appear in your contact list. Then you are able to forward you messages.

srazi:
@auchri Sorry for my bad English. It seems you didn't get what I mean. I try to explain more:
I can forward messages to myself in several ways, for example:
 1. Select messages -> click forward button -> search my name or username -> select my contact item from search result -> click on send button.
 2. Select messages -> click forward button -> go to a group chat info -> select my contact item from chat member list -> click on send button.
 3. First pin my contact item -> select messages -> click forward button -> select my contact item that is pinned -> click on send button.
 4. (the way that I suggested here) Select messages -> click forward button -> click on three parallel lines at top-left -> click on profile picture -> click on send button.

P.S.1: As you can see both way 1 and way 2 needs more mouse or keyboard interaction than the way that I suggested in this issue.

P.S.2: The way 3 is good but one of 5 pin locations is used by my contact item.

stek29:
@auchri That's a workaround.

Being unable to select chat with yourself when forwarding, but being able to do it in normal state is very counter-intuitive, and I'd call it a bug.

PR #3248, opened 2017-04-10
Fix canceling forward when clicking on _userpicButton or _cloudButton

srazi:
To be honest, I am not very familiar with TDesktop code and I didn't know how to check if user is forwarding some messages. But the patch works correctly without that check.

### DIFF
--- a/Telegram/SourceFiles/window/window_main_menu.cpp
+++ b/Telegram/SourceFiles/window/window_main_menu.cpp
@@ -32,6 +32,7 @@ Copyright (c) 2014-2017 John Preston, https://desktop.telegram.org
 #include "lang.h"
 #include "core/click_handler_types.h"
 #include "auth_session.h"
+#include "mainwidget.h"
 
 namespace Window {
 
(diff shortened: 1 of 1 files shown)


stek29:
#3192 for anyone wondering

Telegram/SourceFiles/window/window_main_menu.cpp
```diff
@@ -81,7 +83,7 @@ void MainMenu::checkSelf() {
 	if (auto self = App::self()) {
 		auto showSelfChat = [] {
 			if (auto self = App::self()) {
-				Ui::showPeerHistory(App::history(self), ShowAtUnreadMsgId);
+				App::wnd()->mainWidget()->onForward(self->id, ForwardSelectedMessages);
```
john-preston: 1. I'm afraid this won't work with some other "choose recipient" cases, like "Share contact".

2. This code now is unclear. I think there should be a method in App::main() that does exactly what happens when you click on a user in chats list: shows chat or chooses user for forwarding messages or chooses user for sharing contact etc. This method should be called both here and in click handler in dialogs widget.

srazi:
@john-preston Thanks for the hint. PR was updated.

john-preston:
Thank you, that works great.

PR #3250, opened 2017-04-10
Make theme preview more realistic

mymedia2:
Time on theme preview is not consistent. There is a difference between time in dialog list and time of last message in current dialog. Besides, I doubt that an interlocutor, Eva Summer, can answer immediately, without any delay.

tdesktop theme preview

This commit fixes these contradictions.


stek29:
But isn't that a feature?

auchri:
@stek29 

> Btw, it makes sense to move those strings to localization files

But not in this PR.

john-preston:
@stek29 

I don't think so. The preview was designed with those specific strings in mind, if you will translate it the line count won't be the same in some places. And you don't need to understand the phrases perfectly to decide whether you want to apply a theme or not :)

john-preston:
Thanks for the fix.

PR #3261, opened 2017-04-11
Clean up regular expressions

stek29:
New expressions are the same[1], but a big shared chunk was taken into
macro to simpliy understanding
<pre>
[1] -- the only difference is '`' is not an allowed "divider" for
Code&Pre anymore, because it makes no sence for it to be.
You can test it with this expression: \`\`\`fo\`\`\`o:
- Before: gets parsed as <code>``fo</code>``o
- Now: ```fo```o (stays unchanged)
</pre>

Also, I can't understand why is `?` allowed for code/pre but not for
hashtags/mentions/bot commands... But I decided to keep it.


john-preston:
I don't like an idea of macro there. This is legacy code and if you're improving it, it is better to convert them to functions that return an internal static QRegularExpression variable. You can use simple string concatenation to compose the patterns when you initialize those static variables.

john-preston:
Something like that:
```
QString Dividers() {
  return qsl("\\s\\.,:;<>|'\"\\[\\]\\{\\}\\~\\!\\%\\^\\*\\(\\)\\-\\+=\\x10");
}

const QRegularExpression &reHashtag() {
  static const auto result = QRegularExpression(qsl("(^|[") + Dividers() + qsl("`])#[\\w]{2,64}([\\W]|$)"), QRegularExpression::UseUnicodePropertiesOption);
  return result;
}
...
```

john-preston:
Also rename them to `GetHashtagRegExp()`, `GetMentionRegExp()` etc. and make all of them visible outside that module (like there is `const QRegularExpression &reDomain()` already), leaving just the `Dividers()` function in the anonymous namespace.

stek29:
@john-preston What about putting all of them into one namespace (`GlobalRegExp` or something)?

john-preston:
I'd leave like that for now, they're part of text entity parsing.

stek29:
Sure.

PR #3346, opened 2017-05-02
More accurate phrasing when copying text

lindhe:
If I select a message and right click it says "Copy Selected Text",
while in fact it puts more information than just the text as content.

If I select the message "Test" and choose "Copy Selected Text", it
copies the following:
"Andreas Lindhé, [02.05.17 18:12]
Test"

By this choice of words, it is more clear to the user that it's not only
the text that gets copied.

I would have preferred "Copy Selected Messages", but that would
necessitate extending the functionality to check if more than one
message was selected (in order to look nice and be grammatically
correct). Another option would have been "Copy Selected Content", but
that might be perceived as copying multimedia content as well.

Signed-off-by: Andreas Lindhé <andreas@lindhe.io> (github: lindhe)

### DIFF
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -962,7 +962,7 @@ Copyright (c) 2014-2017 John Preston, https://desktop.telegram.org
 "lng_context_pin_msg" = "Pin Message";
 "lng_context_unpin_msg" = "Unpin Message";
 "lng_context_cancel_upload" = "Cancel Upload";
-"lng_context_copy_selected" = "Copy Selected Text";
+"lng_context_copy_selected" = "Copy Selected as Text";
 "lng_context_forward_selected" = "Forward Selected";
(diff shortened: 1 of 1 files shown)


john-preston:
Thank you for your contribution.

I see your point, but this is not exactly what I would do in this case. The current phrase is perfect for the selected part of text, but is not that good for one or more selected messages.

So the phrase should be kept as it is for the selected text but changed for the case when we select one or more messages, in which case it could be changed to the one you propose.

As you wish - I can implement it by myself (I think it would be a bit faster since I know the code) or I could leave it here so you can improve your pull request to meet those requirements.

lindhe:
I agree.

Thanks for asking, but it is not important for me to be in the cool club just for the sake of it. I'm happy if I contributed in this way too, and it's of no use for anyone to go the harder route. Please feel free to implement this yourself.

Cheers!

PR #3390, opened 2017-05-13
Fix for string

ctl:
It's looks weird , lol


john-preston:
Thank you for the fix. But I'll change to another string and in all such phrases.

Issue #3379, opened 2017-05-10
Build failure against Qt 5.9.0


xvitaly:
### Steps to reproduce
1. Build using systemwide Qt 5.9.0.
2. 
3. 

### Expected behaviour
Normal build.

### Actual behaviour
Build failure:
```c++
[ 63%] Building CXX object CMakeFiles/Telegram.dir/builddir/build/BUILD/tdesktop-1.0.37/Telegram/SourceFiles/storage/localstorage.o
In file included from /usr/include/qt5/QtCore/QtCore:4:0,
                 from /builddir/build/BUILD/tdesktop-1.0.37/out/Release/Telegram_pch/stdafx.h:42:
/usr/include/qt5/QtCore/qflags.h: In instantiation of 'class QFlags<QFlags<MTPDstickerSet::Flag> >':
/builddir/build/BUILD/tdesktop-1.0.37/Telegram/SourceFiles/storage/localstorage.cpp:3274:94:   required from here
/usr/include/qt5/QtCore/qglobal.h:733:47: error: static assertion failed: QFlags is only usable on enumeration types.
 #define Q_STATIC_ASSERT_X(Condition, Message) static_assert(bool(Condition), Message)
                                               ^
/usr/include/qt5/QtCore/qflags.h:96:5: note: in expansion of macro 'Q_STATIC_ASSERT_X'
     Q_STATIC_ASSERT_X((std::is_enum<Enum>::value), "QFlags is only usable on enumeration types.");
     ^~~~~~~~~~~~~~~~~
In file included from /usr/include/qt5/QtCore/qglobal.h:45:0,
                 from /usr/include/qt5/QtCore/QtCore:4,
                 from /builddir/build/BUILD/tdesktop-1.0.37/out/Release/Telegram_pch/stdafx.h:42:
/usr/include/c++/7/type_traits: In instantiation of 'struct std::underlying_type<QFlags<MTPDstickerSet::Flag> >':
/usr/include/qt5/QtCore/qflags.h:112:17:   required from 'class QFlags<QFlags<MTPDstickerSet::Flag> >'
/builddir/build/BUILD/tdesktop-1.0.37/Telegram/SourceFiles/storage/localstorage.cpp:3274:94:   required from here
/usr/include/c++/7/type_traits:2245:38: error: 'QFlags<MTPDstickerSet::Flag>' is not an enumeration type
       typedef __underlying_type(_Tp) type;
                                      ^~~~
[...]

PR #3400, opened 2017-05-15
localstorage: qFlags is working only on enumeration types

ignatenkobrain:
Closes: https://github.com/telegramdesktop/tdesktop/issues/3379
Signed-off-by: Igor Gnatenko <ignatenko@redhat.com>


ignatenkobrain:
P.S. I didn't test it yet

ignatenkobrain:
(at least it allows to build under Qt5.9+), and other `|` are used in the same way, so I think it is correct fix

john-preston:
Thank you!

PR #3479, opened 2017-06-02
Fix inconsistency of max width for message content

Mendor:
Full-width text-only messages are slightly wider than images.


stek29:
Signoff is missing :c

Mendor:
@stek29 fixed it

john-preston:
Thank you for this improvement!

Issue #3200, opened 2017-04-04
https instead of git for Qt repo link


stek29:
<h3>This comment is not relevant anymore, please scroll down a bit</h3>
`init-repository` fails if `git` is used for origin remote.
<details><summary>What happens if I try to init repo:</summary>
<pre>
=> perl ./init-repository --module-subset=qtbase,qtimageformats
+ git submodule init qtbase qtimageformats
Submodule 'qtbase' (git://code.qt.io/qt/qtbase.git) registered for path 'qtbase'
Submodule 'qtimageformats' (git://code.qt.io/qt/qtimageformats.git) registered for path 'qtimageformats'
+ git config commit.template /Users/stek/qt5/.commit-template
+ git clone --no-checkout git://code.qt.io/qt/qt5/../qtbase.git qtbase
Cloning into 'qtbase'...
fatal: remote error: no such repository: /qt/qt5/../qtbase.git
git clone --no-checkout git://code.qt.io/qt/qt5/../qtbase.git qtbase exited with status 32768 at ./init-repository line 198.
	Qt::InitRepository::exe(Qt::InitRepository=HASH(0x7f97af02bf40), "git", "clone", "--no-checkout", "git://code.qt.io/qt/qt5/../qtbase.git", "qtbase") called at ./init-repository line 534
	Qt::InitRepository::git_clone_one_submodule(Qt::InitRepository=HASH(0x7f97af02bf40), "qtbase", "qt5/../qtbase.git", 0) called at ./init-repository line 408
	Qt::InitRepository::git_clone_all_submodules(Qt::InitRepository=HASH(0x7f97af02bf40), "qt5", 0, "qtbase", "qtimageformats") called at ./init-repository line 644
	Qt::InitRepository::run(Qt::InitRepository=HASH(0x7f97af02bf40)) called at ./init-repository line 655
</pre>
</details>

However, if I clone repo via `https://code.qt.io/qt/qt5.git` / change `origin` remote to it, everything works perfectly fine.
I'm not so confident, but I think `..` isn't supported for `git://`

auchri:
@stek29 As stated in the build instructions, the correct command for cloning qt is git clone git://code.qt.io/qt/qt5.git qt5_6_2. I don't know where you got that .. from.

stek29:
@auchri As I said, `init-repository` fails, not the initial clone.

> where you got that `..` from

[...]

PR #3552, opened 2017-06-16
Fix errors/bugs in VS2015 build instructions

grasmanek94:


docs/building-msvc.md
```diff
@@ -196,7 +197,7 @@ While still running the VS2015 x86 Native Tools Command Prompt, go to **D:\\msys
 
 <!-- -->
 
-    git clone git://code.qt.io/qt/qt5.git qt5_6_2
+    git clone https://code.qt.io/qt/qt5.git qt5_6_2
```
stek29: See #3200

john-preston: I got this git: link from some Qt manual. Why is that important?

grasmanek94: because the git link is outdated and doesn't work, http too.

stek29: @grasmanek94 It is outdated and it's insecure, but it does work.

grasmanek94: Hmm okay it works again, just tested it now, but it didn't work yesterday for some unknown reason.

Issue #687, opened 2015-04-30
Option to disable the link preview


Akamaru:
Pls add an option to disable link previews.
![2015-04-30_20-04-52](https://cloud.githubusercontent.com/assets/8063440/7419572/2927004a-ef76-11e4-887e-cfb20c813a1b.png)

Brawl345:
Yeah, they're really big and I think that your PC connects to the services like youtube? Not really a good feature for a privacy messenger.

ppparadox:
+1

OliverWillemse:
I would also very much like this feature!

yuergen:
I concur. The link preview is annoying.

Sanva:
+1 —I'm also really annoyed by this... it's really disturbing to be writing with someone talking about some web application we are developing and having that preview every time we mention the app by its domain name.

auchri:
@Brawl345 your pc is not connecting to the services, instead the telegram servers are connecting to the different websites

yuergen:
@auchri if the URI includes tracking information, information will leak to the site's operator.

halcion:
please, i cannot stress enough how much i dislike the link preview, please remove it.

Aokromes:
How about a dropdown with the following options:

"Preview on message" (current mode)
"Preview on mouseover"
"Preview on click"
"Preview disabled"

brantje:
+1, i expected an option to turn it off when they introduced link previews...
Also, as @Brawl345 mentioned, it also gets youtube / facebook links, not good for your privacy

animalillo:
+1000

Aokromes:
Maybe the setting can he also be done per group or contact overwriting general settings.

tech-no-logical:
+1 this should really be optional

ezdev128:
Please disable this [censored] preview feature!!!
Today I had an issue when trying to send auth link (to recover password to one of my online services) and I  was unable to do it because your built-in crawler already gets this link, therefore remote side think you have already checked link and doesn't allows to change password. My account has been blocked for fraud attempts after 3rd unsuccessful try.
This is awful.

@telegramdesktop how much signatures needs from unhappy persons to disable this [censored] feature? 10? 20? 100? 10000??? Don't you have an eyes or don't you read the conversation??? This question related to https://github.com/telegramdesktop/tdesktop/issues/807

ralesk:
:+1: would love to have a setting to disable images sent/received and link previews

[...]

PR #3559, opened 2017-06-17
Done #687: options to disable link preview

vdmit:
I've implemented a feature request #687 that was published over 2 years ago and is wanted by many users: couple of options to control web pages preview. Now web pages and documents preview could be disabled in Chat options.

no-preview-checkboxes

Rationale: ability to have compact history, when lots of links are used (web dev or design collaboration, etc.)

### DIFF
--- a/Telegram/Resources/langs/lang_es.strings
+++ b/Telegram/Resources/langs/lang_es.strings
@@ -288,6 +288,8 @@ Copyright (c) 2014-2017 John Preston, https://desktop.telegram.org
 "lng_settings_section_chat_settings" = "Ajustes de chat";
 "lng_settings_replace_emojis" = "Reemplazar emojis";
 "lng_settings_view_emojis" = "Ver lista";
+"lng_settings_no_web_page_preview" = "Desactivar la vista previa de la página web";
+"lng_settings_no_document_preview" = "Desactivar la vista previa de los documentos";
 "lng_settings_emoji_list" = "Emojis soportados";
 "lng_settings_send_enter" = "Enviar con Intro";
 "lng_settings_send_ctrlenter" = "Enviar con Ctrl+Intro";


Telegram/SourceFiles/storage/localstorage.cpp
```diff
@@ -1720,7 +1738,9 @@ void _writeUserSettings() {
 		return Window::Controller::kDefaultDialogsWidthRatio;
 	};
 
-	uint32 size = 21 * (sizeof(quint32) + sizeof(qint32));
+	// NOTE: number below is a number of single qint32 options that are serialized right after 'data' object definition.
+	// don't forget to update it after adding of new option.
+	uint32 size = 26 * (sizeof(quint32) + sizeof(qint32));
```
stek29: I thought 21+2 is 23 and not 26 XD

stek29:
1. Please squash the commits
2. Don't edit langs except English -- they'd be handled on Transifex
3. And don't forget about the signoff 😉

stek29:
Also, I guess you've tested what happens if you change the setting when previews are already rendered and handled the situation properly (redraw dialog), right?

Telegram/SourceFiles/storage/localstorage.cpp
vdmit: Sure, but it seems that this value was not incremented earlier - please count serialized values thoroughly.

stek29:
@vdmit Can't you just rebase on top of dev instead of making a merge commit?

vdmit:
Well, I'm not a git guru today. I've tried to do git reset --soft, but it does not work as expected and still add new commits. Now i've done a fresh repo and create another PR, seems that it's shorter.

About redraw, etc: of course, I tested option enable/disable, and it works adeqately: If you disable it, new links are not previewed. If you enable it again - new links are previewed.

stek29:
@vdmit But maybe it should affect old links too?

vdmit:
In fact, links change their view exactly on first draw or redraw, and I think that it's OK.
For example, if you switch between chats or scroll - they are NOT redrawed and therefore does not change , even if option was already switched. But, if you restart TD completely - all history elements are redrawed in their 'new' flavour, matching current settings. Maybe there other events that trigger redraw, but still, it looks pretty good. And of course, those who want to disable it - they'll turn off previews once for ages, so there won't be any surprises.

stek29:
@vdmit IMO redraw should be forced for such messages on option switch, but it's just my opinion.

Issue #3330, opened 2017-04-26
Toggle/Collapse downloaded images/media in chat


FabianMeul:
### Steps to reproduce
1. Open a chat where the last message is e.g. an @gif entry
2. Tap a toggle icon/button that collapses the image
3. Profit

### Expected behaviour
The behaviour is requested as above. ;-)

### Actual behaviour
The behaviour is requested as above

### Configuration
**Operating system:** MacOS 12.0.4

**Version of Telegram Desktop:** 1.0.29

**Used theme**: Default

I think this could be a great addition to Telegram, as the app makes it super easy to spam "gifs" (which is not necessarily a bad thing!) but gives no option to hide those gifs, besides removing them completely.

FabianMeul:
#1535

stek29:
Does any other Telegram app have it? Could you provide better explanation (maybe something visual)

FabianMeul:
No, I think we'd be the first to implement this. Slack has a comparable functionality, where you can toggle the visibility of media files.

I'll try to whip up a visual proposal tonight. :-)

FabianMeul:
I hope these images explain the concept a little better. :-)

Toggle Open
![media-open](https://cloud.githubusercontent.com/assets/756903/25454011/5a1a717c-2acb-11e7-9afa-9064d89461ca.png)

Toggle Closed
![media-closed](https://cloud.githubusercontent.com/assets/756903/25454016/611b087e-2acb-11e7-8745-41dd0d9b3bf1.png)


PS: Thx to Dmitriy Markov for providing a Sketch design file!
(https://dribbble.com/shots/3289878-Telegram-desktop-messenger-Sketch-included)

stek29:
#687, but this one is much cleaner

FabianMeul:
If this ticket is under serious consideration, I'm willing to spend some extra time to visually refine the proposal :-)

Are there design guidelines for interface proposals?

john-preston:
@FabianMeul I'm afraid it is not under serious consideration for now, sorry.

charlesmims:
I'd like to piggy back on this proposal and ask for slack-like "/" commands    /collapse in this case

PR #3560, opened 2017-06-18
Done #687: options to disable link preview [re-created PR]

vdmit:
I've implemented a feature request #687 that was published over 2 years ago and is wanted by many users: couple of options to control web pages preview. Now web pages and documents preview could be disabled in Chat options.

no-preview-checkboxes

Rationale: ability to have compact history, when lots of links are used (web dev or design collaboration, etc.)

### DIFF
--- a/Telegram/SourceFiles/storage/localstorage.cpp
+++ b/Telegram/SourceFiles/storage/localstorage.cpp
@@ -1738,21 +1738,34 @@ void _writeUserSettings() {
 		return Window::Controller::kDefaultDialogsWidthRatio;
 	};
 
-	// NOTE: number below is a number of single qint32 options that are serialized right after 'data' object definition.
+	// NOTE: this is a number of single qint32 options that are serialized below.
 	// don't forget to update it after adding of new option.
-	uint32 size = 26 * (sizeof(quint32) + sizeof(qint32));
+	uint32 size = 22 * (sizeof(quint32) + sizeof(qint32));
+
+	// for dbiDownloadPath
 	size += sizeof(quint32) + Serialize::stringSize(Global::AskDownloadPath() ? QString() : Global::DownloadPath()) + Serialize::bytearraySize(Global::AskDownloadPath() ? QByteArray() : Global::DownloadPathBookmark());
 
+	// dbiRecentEmoji
 	size += sizeof(quint32) + sizeof(qint32);
 	for (auto &item : recentEmojiPreloadData) {
 		size += Serialize::stringSize(item.first) + sizeof(item.second);
 	}
 
+	// for dbiEmojiVariants
 	size += sizeof(quint32) + sizeof(qint32) + cEmojiVariants().size() * (sizeof(uint32) + sizeof(uint64));
+
+	// for dbiRecentStickers
 	size += sizeof(quint32) + sizeof(qint32) + (cRecentStickersPreload().isEmpty() ? cGetRecentStickers().size() : cRecentStickersPreload().size()) * (sizeof(uint64) + sizeof(ushort));
+
+	// for dbiDialogLastPath
 	size += sizeof(quint32) + Serialize::stringSize(cDialogLastPath());
+
+	// for dbiAutoDownload
 	size += sizeof(quint32) + 3 * sizeof(qint32);
+
+	// for dbiDialogsMode
 	size += sizeof(quint32) + 2 * sizeof(qint32);
+
 	if (!Global::HiddenPinnedMessages().isEmpty()) {
 		size += sizeof(quint32) + sizeof(qint32) + Global::HiddenPinnedMessages().size() * (sizeof(PeerId) + sizeof(MsgId));
 	}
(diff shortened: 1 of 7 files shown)


Telegram/SourceFiles/storage/localstorage.cpp
```diff
@@ -1720,7 +1738,9 @@ void _writeUserSettings() {
 		return Window::Controller::kDefaultDialogsWidthRatio;
 	};
 
-	uint32 size = 21 * (sizeof(quint32) + sizeof(qint32));
+	// NOTE: number below is a number of single qint32 options that are serialized right after 'data' object definition.
+	// don't forget to update it after adding of new option.
+	uint32 size = 26 * (sizeof(quint32) + sizeof(qint32));
```
stek29: This should be 22 and not 26.

stek29:
Why it should be 22 and not 26?
Well...


dbiDownloadPath and dbiDialogLastPath should not be included, they're calculated in other place.
dbiAutoDownload has 3 ints instead of 1, so it's handled 11 lines below (`size += sizeof(quint32) + 3 * sizeof(qint32);`)
`size += sizeof(quint32) + 2 * sizeof(qint32);` 12 lines below was added along with `dbiDialogsMode` which has 2 ints.

A little (4 hour) dive into blame and logs showed this:

Commit|Calculated size|Actual size
---|---|---
This PR|26|22
1540f6f528a6cfe0dc8dd9e0f0ee112d5d1eded9 | 20 | 21
75462452132f9af8e42a5fcab236b9f549e57c74 | 22 | 21
620663da21792a75b102923c15dbab02d14d4fb5 |21 | 21
fc9d75b4f01fed32f3b4d5827ac47c5724ea5787 | 19 | 19
3cff50009ca4c7a18bf711ae3d14bb6d4cafac76 | 19 | 20
16ce28f4d25e43cb8fe2ec519316940080dd2bda|20|20
8da39356dc33bb75caf534eac1e9b5b275c361ed|18|18
cbde6e1ee9b20b06d3f09a7abd674858890e7cce|17|17
fc9d75b4f01fed32f3b4d5827ac47c5724ea5787|16|16
9a1798f043acd6daca3d9df140017b5a44203c6e|16|15
ba0825c892870e33162322c31bb90cc44c1cb6fd|14|13
9a51e2c7e3c51edbc4bfc8a477ec12ae3261891a|14|14
cf9737e762e161d788986e6d24c9b4252ebd53dc|13|13
12687745172f78b8dff98c39753ffdb86b79d849|12|12
ad04bf77d1c95a9f53dc9dbf7d12b8e12746e759|11|11


- Also, b35c99cb0c6fa0747aafc74ab599ee91cb91a975 [suggests that](https://github.com/telegramdesktop/tdesktop/commit/b35c99cb0c6fa0747aafc74ab599ee91cb91a975#diff-82a0112076431fa56e0d6de9b94ab920R2187) there's one more error like this in the file. [6 should be changed to 7](https://github.com/telegramdesktop/tdesktop/blob/8b54910538318ba7ad6f4be14ddbd4a9137194d5/Telegram/SourceFiles/storage/localstorage.cpp#L2306) and [12 to 13](https://github.com/telegramdesktop/tdesktop/blob/8b54910538318ba7ad6f4be14ddbd4a9137194d5/Telegram/SourceFiles/storage/localstorage.cpp#L2306)
- As 2940561ee211850fdade1a00f6c08eed19ed894d shows, dbiTryIPv6 is added to that count too, so it has to be not 12 and not 13 but 14!
- As af9edc17d2ec685257309d07439fbb1f249dd126 suggests, [`2 *` should be removed](https://github.com/telegramdesktop/tdesktop/blob/8b54910538318ba7ad6f4be14ddbd4a9137194d5/Telegram/SourceFiles/storage/localstorage.cpp#L2304)
- And one more (4ec579112c4c1505cbebf3168899323c8eb4bf3d) -- 3 should be changed to 6 [here](https://github.com/telegramdesktop/tdesktop/blob/8b54910538318ba7ad6f4be14ddbd4a9137194d5/Telegram/SourceFiles/storage/localstorage.cpp#L2578), [see this line in diff](https://github.com/telegramdesktop/tdesktop/commit/4ec579112c4c1505cbebf3168899323c8eb4bf3d#diff-82a0112076431fa56e0d6de9b94ab920R2381)

So, size calculations are completely fucked up.

vdmit:
@stek29 thanks a lot, i've fucked up here too :) Very error-prone section indeed.
Now I've annotated all this serialization to make it easier to change later.
Don't realize how it could be refactored even better, but now each magic number has its meaning.

stek29:
@vdmit No problem :)
It's much better now imo, and maybe someone should make a PR fixing other size calculations.

vdmit:
Mmm.. in general, you're right, features should not be mixed with refactoring. But in this case I think it should be in one single commit: we add **two** options, but size increment seems inappropriate at a glance. That's why these changes should be together, and the "extra" diff is not very huge to understand.

Returning to forcing of redraw after option value change - I'll try it later. Ultimately, the most of users' hate about this "feature" is thrown away now. And if there are perfectionists - they could improve it by their own. 

Let's ship it and have fun!

vdmit:
While this PR is in progress, anyone who is interested in this feature, could obtain it in 2 ways:

1. Get tdesktop sources from fork https://github.com/vdmit/tdesktop and build it in your own way 
2. If you're using Ubuntu Linux, you are lucky: you can use new cool all-in-one build script from https://github.com/vdmit/autogram
Just change TD repo path (beginning of the script) to https://github.com/vdmit/tdesktop and run it.

john-preston:
Thanks for your work, but I'm afraid I won't merge this request (requests that add new UI elements almost never get merged, see [contributing guidelines](https://github.com/telegramdesktop/tdesktop/blob/dev/.github/CONTRIBUTING.md)).

Also, I can see some sense in the disabling web links preview, but the documents preview checkbox just leaves me puzzled: you simply are not initing media for a message that contains only media, specifically - the file that someone have sent you. So when someone sends you a file (any file at all), you'll just won't receive it at all, you'll get some bugged empty message instead. I don't see the point.

vdmit:
Yeah, thanks for your notice about documents checkbox. I've realized a bit later (after I've used my version for a while) that this checkbox is doubtful.
Okay, besides all this, if we don't want to add any new checkboxes, it there any way to make all these preview-like frames **a bit more compact**? For instance, when I open some chats after some "absence", I'm tired of scrolling unread messages down - just because they are large  (I mean px height, not texts). I'm able to read the texts fast, but there is no option to observe all these previews at a glance.

auchri:
@vdmit https://github.com/telegramdesktop/tdesktop/issues/3330

PR #3561, opened 2017-06-18
Minor build improvements (macOS mostly)

stek29:
- Fix TDESKTOP_DISABLE_CRASH_REPORTS define on macOS
- Add TDESKTOP_BUILD_DEFINES to `refresh.sh`, since they're in `refresh.bat` already.
- Update xcode and cmake docs

### DIFF
--- a/Telegram/gyp/telegram_mac.gypi
+++ b/Telegram/gyp/telegram_mac.gypi
@@ -28,8 +28,6 @@
       'OTHER_LDFLAGS': [
         '-lbsm',
         '-lm',
-        '-lssl',
-        '-lcrypto',
         '/usr/local/lib/liblzma.a',
       ],
     },
@@ -78,6 +76,8 @@
         '/usr/local/macold/lib/libexif.a',
         '/usr/local/macold/lib/libc++.a',
         '/usr/local/macold/lib/libc++abi.a',
+        '<(libs_loc)/macold/openssl-1.0.1h/libssl.a',
+        '<(libs_loc)/macold/openssl-1.0.1h/libcrypto.a',
       ],
     },
     'include_dirs': [
@@ -86,9 +86,6 @@
       '<(libs_loc)/macold/openssl-1.0.1h/include',
       '<(libs_loc)/macold/crashpad',
       '<(libs_loc)/macold/crashpad/third_party/mini_chromium/mini_chromium',
-    ],
-    'library_dirs': [
-      '<(libs_loc)/macold/openssl-1.0.1h',
     ],
     'configurations': {
       'Debug': {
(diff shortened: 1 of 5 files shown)


stek29:
Huh, appveyour failed to detect signature...

stek29:
@john-preston I've also changed openssl ldflags, but I can't check them for macold (and for openssl-xcode too since I'm building TDesktop in slightly different way)

Telegram/gyp/telegram_mac.gypi
```diff
@@ -122,9 +119,9 @@
     }, {
       'postbuild_name': 'Copy crashpad_handler to Helpers',
       'action': [
-        'cp',
-        '<(libs_loc)/macold/crashpad/out/${CONFIGURATION}/crashpad_handler',
-        '${BUILT_PRODUCTS_DIR}/Telegram.app/Contents/Helpers/',
+        'sh', '-c',
+        'cp "<(libs_loc)/crashpad/out/${CONFIGURATION}/crashpad_handler"\
+        "${BUILT_PRODUCTS_DIR}/Telegram.app/Contents/Helpers/" && exit 0'
```
john-preston: If this is used to continue with build no matter was the crashpad copied or not - this is not a good solution, because in case of not disabled crash reports I need to get a build error if it was not copied.

Also you've removed "/macold/" for some reason from crashpad path here.

Telegram/gyp/telegram_mac.gypi
```diff
@@ -78,6 +76,8 @@
         '/usr/local/macold/lib/libexif.a',
         '/usr/local/macold/lib/libc++.a',
         '/usr/local/macold/lib/libc++abi.a',
+        '<(libs_loc)/macold/openssl-1.0.1h/libssl.a',
+        '<(libs_loc)/macold/openssl-1.0.1h/libcrypto.a',
```
john-preston: Why is this better than "-lssl -lcrypto" with correct library_dir?

Telegram/gyp/telegram_mac.gypi
stek29: > Also you've removed "/macold/" for some reason from crashpad path here.

Copypasting :(

> If this is used to continue with build no matter was the crashpad copied or not - this is not a good solution, because in case of not disabled crash reports I need to get a build error if it was not copied.

I know, but I wasn't able to find a better solution :(

Telegram/gyp/telegram_mac.gypi
stek29: @john-preston Because if I have libssl.dylib and libcrypto.dylib in library dir (which will be in /usr/local/lib with a very high chance) linker seems to prefer them to static archives. And I get an executable which links to everything statically except SSL libs, and it crashes on other system which doesn't have those dylibs at the same path, or when homebrew updates openssl)

It doesn't happen with Qt since having dylibs of Qt in library search path is very unlikely -- much more unlikely than having openssl libs. (In fact, homebrew's Qt5 is "keg only formula", so it doesn't link to /usr/local/lib unless user forces it)

stek29:
I'll spit it into different PRs -- one for OpenSSL (#3635 suggests I'm not the only one) and one for build defines.
And I still don't have a clue on how to make conditional Xcode action in gyp :\

stek29:
Can someone who knows gyp help? :\

john-preston:
@stek29 Some custom script with build env as params and condition inside should do the trick.

mymedia2:
> And I still don't have a clue on how to make conditional Xcode action in gyp :\

@stek29 you can determine that gyp produces Xcode project, using `GENERATOR` undocumented variable. Like this:

```python
'conditions': [
  ['GENERATOR == "xcode"', {
    'actions': [
      {
        'action_name': 'blablabla',
        # ...
      }
    ]
  }]
]
```

PR #3565, opened 2017-06-20
Create fa.strings

sssgnu:
created persian language for tdesktop


stek29:
Translations are managed on Transifex.

auchri:
Yes, also the signature is missing (for the next time).

Issue #2730, opened 2016-12-22
[Feature Request] Show "original sent" timestamp for forwarded messages


MasterGroosha:
(not using template because it's useless here)

When you hover mouse over edited message, is shows you "first sent" timestamp and "last edited" one.
![image](https://cloud.githubusercontent.com/assets/2780044/21421607/d11e77d0-c844-11e6-9026-7cec1b33bf67.png)
Since "edited" is not shown for forwarded messages, it will be very nice to see timestamp of original message, which was forwarded.

stek29:
But forwarded messages can't be edited, right?

MasterGroosha:
@stek29 That's why "original sent" timestamp can be placed instead of "last edited" one.

stek29:
Isn't this an API issue?

MasterGroosha:
@stek29 Since TDesktop has "original sent" timestamp for messages - no, this is not API issue

stek29:
@MasterGroosha Where does it?

MasterGroosha:
Look at my screenshot above: message has "original sent" and "edited" timestamps. 
What I suggest is to replace "edited" with "original sent" for **forwarded** messages since they can't be edited.

stek29:
@MasterGroosha What do you mean by "Original send"? Do you want it to show time when message was forwarded and time when original message which is being forwarded was sent?

MasterGroosha:
Exactly

stek29:
@MasterGroosha Oh, I see. Yup, that's not an API issue indeed, messageFwdHeader has Date.

auchri:
Fixed by #3568

Issue #2106, opened 2016-06-01
Show original timestamp of forwarded messages


DarthJahus:
Forwarded messages, actually, have the timestamp of when they got forwarded. It would be good, it availabled on the API, to additionally show the timestamp of the original message; that way, it would make more sense when we forward conversation, and avoid forwarding messages in a custom order that would create a mess.

stek29:
And timestamp of last edition too?

DarthJahus:
@stek29, it would be great to see edit timestamp on tooltip for every edited message, like reddit does it.

stek29:
Well, it's an API issue anyway.
So you should use "Ask a question" button in settings, or ping telegram on twitter

tdesktop-closer:
Hey there!

We're automatically closing this issue since there was no activity in this issue since 379 days ago. We therefore assume that the user has lost interest or resolved the problem on their own. Closed issues that remain inactive for a long period may get automatically locked.

Don't worry though; if this is in error, let us know with a comment and we'll be happy to reopen the issue.

Thanks!

<sub>(Please note that this is an [automated](https://github.com/auchri/github_auto_closer) comment.)</sub>

stek29:
Sorry for that misinformation. It is not an API issue.
However, I'm sure I've seen a more recent request for this with more activity.

DarthJahus:
I believe it's an API issue since the "last edit" data has to be stored on Telegram servers and not just locally.
Or is it already stored there?

stek29:
Yes, there's date of original message.

PR #3568, opened 2017-06-20
Show date from fwdHeader on tooltip

stek29:
This is how it looks like currently:
image

I'm still not sure about the wording.


stek29:
Oh, and forgot to say
Fixes #2730 #2106

john-preston:
Thanks for your work!

PR #3578, opened 2017-06-26
Always use UTF-8 to open files in list_sources.py

eduardosm:
This avoids problems with some locale settings.


john-preston:
Thanks for your pull request!

PR #3588, opened 2017-06-28
Create fa.strings

sssgnu:
Persian language added!


stek29:
Please look what I and @auchri have said in #3565.

PR #3614, opened 2017-07-01
Fix remove option in normal groups

stek29:
If user invites X to group and then X becomes an admin, user won't be able to remove X anymore, so remove option shouldn't be shown.

Currently remove is always shown if X was invited by user.
image


john-preston:
Indeed, thanks.

PR #3632, opened 2017-07-06
Fix a couple of typos

mymedia2:


stek29:
@mymedia I remember some "succed"s instead of "succeeded", you can look for them too (they were in mtproto/connection* iirc)

john-preston:
Thanks for your fixes.

PR #3639, opened 2017-07-09
Some optimization in MediaView::keyPressEvent.

23rd:
— Multiple calls of key() are replaced with variable.
— Last "else if" had excessive conditions.


stek29:
@23rd iirc `e->key()` is used almost everywhere.
Also, `QKeyEvent::modifiers()` and `QKeyEvent::key()` are `const`, so compiler would optimize that anyway I guess.

Also, signature is missing.

But last else if "optimization" looks great -- code is finally readable, yay.

john-preston:
There is still no signature and I'm not sure this all is really important.

Issue #3635, opened 2017-07-08
library not found for -lssl


OmidMahmoudi:
![screen shot - - at](https://user-images.githubusercontent.com/19889603/27980862-6a77e58a-6398-11e7-9cb0-0f762cf488fa.png)

stek29:
What if you apply my patch from #3561?
stek29/tdesktop@9dd2b43075e6db9b6e9bafa525f7b7bce9837ca6

OmidMahmoudi:
i apply your patch, and successfully build it
thank you

PR #3644, opened 2017-07-10
Fix openssl ldflags on macOS

stek29:
If ld finds ssl/crypto dylibs it prefers them to static archived objects
And because of that produced .app links dynamically to openssl, so it
won't launch unless users install openssl.
There are two solutions: either uninstall openssl completely from build
machine on each build, or use paths to .a files.
Obviously latter is preferred.

See #3561 and #3635


john-preston:
Yeah, thanks.

PR #3712, opened 2017-07-26
Improve multiple people typing/playing message

nyuszika7h:
This is consistent with the Android version, and is more clear.

Also, in languages that use singular for all numbers, it avoids confusion between "3 people are typing" and "a person named '3' is typing".


john-preston:
Yeah, thanks.

Issue #2411, opened 2016-09-07
GTK3 FileChooser is the worst


KAMiKAZOW:
The GTK3 FileChooser has the worst usability of all FileChoosers, incl. Qt's built-in one.
I get that masochists who like Gnome applications like it but why does TDesktop force that thing under proper DEs like KDE Plasma, LXQt, etc.?

chrjguill:
there's a lot I could say, but QT's built-in file picker that used to be used in the linux builds was actually far less usable (ugly as fuck, as well as not being able to send files that were missing extensions, among a few other 'nice' quirks) than GTK's. 

Look at [this](https://github.com/telegramdesktop/tdesktop/issues/987) issue for reference, but until an unpatched QT framework is used, you're kinda stuck.

You're in the vast minority of users that complain about the gtk filepicker over the old, shitty, broken filepicker.

KAMiKAZOW:
I want the KDE's native, far superior file picker which, among other things, displays thumbnails for images.

Users of GTK desktops should get a GTK version of Telegram and not force GTK practices onto a Qt application, just because they happened to shout loudest first.

john-preston:
@KAMiKAZOW Does Qt apps show KDE native file picker? (as I see it is different from custom Qt file picker).

I saw code working with it in google chrome codebase, but not in Qt codebase.

KAMiKAZOW:
Yes, all other Qt5 applications just use the KDE file picker, like QupZilla here:

![screenshot_20160921_165025](https://cloud.githubusercontent.com/assets/3259076/18716089/cdd41df8-801b-11e6-887a-f4a5d05ebbde.png)

AndydeCleyre:
I can confirm that other Qt5 cross platform programs use the KDE file picker, such as [qbittorrent](https://github.com/qbittorrent/qBittorrent).

Aokromes:
#987 ?

c-cesar:
Since 2008 Qt is aware of the DE, and open a GTK file dialog under gnome and a KF5 file dialog under Plasma. So, the best is letting Qt handle this. 
But as @chrjguill pointed, maybe using a bundled Qt lib broke this feature.  
http://blog.qt.io/blog/2008/10/01/native-file-dialogs-in-gnome/

john-preston:
[...]

PR #3778, opened 2017-08-09
New build option: without GTK and appindicator

eegorov:
Following #2411 , I added the ability to build a project without the need for GTK in the system. However, I have not worked with GYP yet, so I do not know how to disable the dependence on gtk+ and appindicator in the cmake step. Maybe someone will help in this. In any case, I think someone will find this interesting.


auchri:
> However, I have not worked with GYP yet, so I do not know how to disable the dependence on gtk+ and appindicator in the cmake step. 

This one: 

```
'cflags_cc': [
      ...
      '<!(pkg-config 2> /dev/null --cflags gtk+-2.0)',
      ...
    ],
```
([telegram_linux.gypi, L78](https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/gyp/telegram_linux.gypi#L78))?

eegorov:
Yes. how to exclude this two lines when TDESKTOP_DISABLE_GTK_INTEGRATION is set?

auchri:
It should work somehow like this: https://stackoverflow.com/a/43618094/4752807

auchri:
@eegorov Did it work? :)

eegorov:
Yes.

Telegram/SourceFiles/platform/linux/main_window_linux.cpp
```diff
@@ -321,6 +343,7 @@ void MainWindow::updateIconCounters() {
 
 	auto counter = App::histories().unreadBadge();
 
+#ifndef TDESKTOP_DISABLE_GTK_INTEGRATION
 #ifndef TDESKTOP_DISABLE_UNITY_INTEGRATION
```
mymedia2: Why do not combine these conditions through the AND operator? Like this

    #if !defined (TDESKTOP_DISABLE_GTK_INTEGRATION) && !defined(TDESKTOP_DISABLE_UNITY_INTEGRATION)
    ...
    #endif

Telegram/SourceFiles/platform/linux/main_window_linux.cpp
```diff
@@ -516,11 +547,13 @@ void MainWindow::psCreateTrayIcon() {
...
 		workmodeUpdated(Global::WorkMode().value());
 	}
+#endif // !TDESKTOP_DISABLE_GTK_INTEGRATION
 }
 
 void MainWindow::psFirstShow() {
 	psCreateTrayIcon();
 
+#ifndef TDESKTOP_DISABLE_GTK_INTEGRATION
 #ifndef TDESKTOP_DISABLE_UNITY_INTEGRATION
```
mymedia2: the same here

eegorov: Logically, this is the same, I agree, but I do not think that this is a principle. In addition, the readability of pull-request is better in my case. IMHO.
In any case, there is still something to improve. Now it is important to agree in principle with this approach.

stek29: @eegorov Readability of pull request is better, but readability of code is worse
Also, why is readability of PR better? Because it shows endif? Just add comment to endif to improve both code and PR readability.

eegorov:
Fix prev. comment (signed-off)+squash commits to one.

john-preston:
Thanks for your work.

PR #3817, opened 2017-08-22
Packed resources

mymedia2:
This introduces a new build option for use packed resources which are folded in a separated file. The resources will be loaded from this file on startup. On Linux, the program tries to search the resources in the next files and uses the first found one.

 * in a directory of the binary (only in debug mode),
 * `~/.local/share/TelegramDesktop/tresources.rcc`,
 * `/usr/local/share/TelegramDesktop/tresources.rcc`,
 * `/usr/share/TelegramDesktop/tresources.rcc`.

I assume somebody will write something like that while configuring.

    gyp -Duse_packed_resources ...

This feature doesn't break the default behaviour.

I also removed duplicated or unused resources:

 * Sounds from the `telegram.qrc` file because these files are already listed in a `telegram_sounds.qrc` file. It eliminates warnings of the resource compiler and reduces total size of the binary by 304 kB.
 * The `freedesktop.org.xml` file because it seems Telegram Desktop doesn't use it at all. If I'm wrong, I'll put it back.

### DIFF
--- a/Telegram/Resources/qrc/telegram.qrc
+++ b/Telegram/Resources/qrc/telegram.qrc
@@ -10,17 +10,6 @@
     <file alias="art/sunrise.jpg">../art/sunrise.jpg</file>
     <file alias="night.tdesktop-theme">../night.tdesktop-theme</file>
   </qresource>
-  <qresource prefix="/sounds">
-    <file alias="msg_incoming.mp3">../sounds/msg_incoming.mp3</file>
-    <file alias="call_incoming.mp3">../sounds/call_incoming.mp3</file>
-    <file alias="call_outgoing.mp3">../sounds/call_outgoing.mp3</file>
-    <file alias="call_busy.mp3">../sounds/call_busy.mp3</file>
-    <file alias="call_connect.mp3">../sounds/call_connect.mp3</file>
-    <file alias="call_end.mp3">../sounds/call_end.mp3</file>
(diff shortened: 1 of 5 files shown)


john-preston:
@mymedia2 Thanks for your work.

About "freedesktop.org.xml" - Qt uses it for MIME library data, this is the default resource path for Qt to load this MIME type database, see:

https://github.com/qt/qtbase/blob/5.9/src/corelib/mimetypes/qmimeprovider.cpp#L779

mymedia2:
Okay, freedesktop.org.xml seems to be used, but how? I can't find any 
indications. And I don't see differences in the appearance or the 
behaviour of the application if there is no this file in the resources 
pack. Perhaps, special conditions are necessary--

john-preston:
Looks like Qt 5.6.2 has its own "freedesktop.org.xml" with mime types right here:

qtbase/src/corelib/mimetypes/mime/packages

and it is linked already. It is not exactly the same as the one bundled with tdesktop (it is less for about 120KB), but maybe it is good enough.

auchri:
What's the status of this PR?

auchri:
Closing for now 🤷‍♂️

mymedia2:
@john-preston, so, what's the suggestion? If Qt already provides a freedesktop.org.xml file, can we remove it from resources?

PR #3842, opened 2017-08-27
Fix CVE-2016-10351: Insecure cWorkingDir permissions.

### LINKED ISSUE #2666: Set better permission on .TelegramDesktop (CVE-2016-10351)
Hello,

by default the .TelegramDesktop directory has 755:

```
drwxr-xr-x  4 ago  ago      4096 nov 23 14:30 .TelegramDesktop
```

I think would be great set it to 700 or 770.

While the main homedir has 755 (and it happens by default here), a local user can obtain the contents of .TelegramDesktop.

I did not investigate what a local user which obtain those files can eventually steal.
Thanks.

gerasiov:
Set 700 permisson on dir on every start.

Should fix #2666


john-preston:
Thanks, that indeed should be better.

PR #3868, opened 2017-09-07
Fix IME being interrupted after reconnection

Yonsh:
ime_issue
Notice how the words were duplicated right after the "Connecting..." message disappeared.

I think this bug is somehow related to the deletion of `ConnectingWidget` (or `QWidget`).
I was able to fix it by only hiding it instead.

**Operating system:** Windows 10 x64


john-preston:
Thanks for your work!

sujunmin:
Could you build a new release for this fix?
Thanks.

john-preston:
@sujunmin I hope a version will be ready within a week. But I’m not sure :(

PR #3878, opened 2017-09-11
Fix build instructions for MSVC

monote:
Fixed 'change directory' commands for following build process step-by-step.

Re-opened as new to cull a number of previous erroneous commits in proposed branch.


stek29:
Thanks :)

PR #3899, opened 2017-09-17
Fix small issues pointed out by static analysis

visuve:
- Probed the code with Cppcheck 1.80

### DIFF
--- a/Telegram/SourceFiles/platform/win/main_window_win.cpp
+++ b/Telegram/SourceFiles/platform/win/main_window_win.cpp
@@ -113,7 +113,7 @@ public:
 	using Change = MainWindow::ShadowsChange;
 	using Changes = MainWindow::ShadowsChanges;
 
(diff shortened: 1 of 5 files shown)


john-preston:
Thanks.

Telegram/SourceFiles/platform/win/main_window_win.cpp
```diff
@@ -113,7 +113,7 @@ class _PsShadowWindows {
 	using Change = MainWindow::ShadowsChange;
 	using Changes = MainWindow::ShadowsChanges;
 
-	_PsShadowWindows() : screenDC(0), max_w(0), max_h(0), _x(0), _y(0), _w(0), _h(0), hidden(true), r(0), g(0), b(0), noKeyColor(RGB(255, 255, 255)) {
+	_PsShadowWindows() : screenDC(0), noKeyColor(RGB(255, 255, 255)) {
```
john-preston: I wonder why screenDC was left here.

visuve: So that someone would ask. I was unsure whether to use nullptr or 0.

PR #3920, opened 2017-09-24
allow a customized BUILD_DIR in install.bat

mrexodia:
This is a very simple change that allow people to build Telegram like AppVeyor does it, but not in C:\TBuild (often C is "System Reserved" and cannot be used). Instead you can build it anywhere (from the visual studio 2017 x86 command prompt) like this:

    >set BUILD_DIR=%CD%
    >git clone https://github.com/telegramdesktop/tdesktop.git
    >cd tdesktop
    >.\.appveyor\install.bat
    >msbuild Telegram\Telegram.sln /property:Configuration=Debug /p:Platform=Win32

Some notes:

- I replaced all C:\TBuild with %BUILD_DIR%, even if it is decided not to merge this, keep that.
- %BUILD_DIR%\Libraries\prepare.bat might fail to extract the 7z Qt libraries for some reason (even if 7z.exe is in PATH) this has not been fixed

### DIFF
--- a/.appveyor/install.bat
+++ b/.appveyor/install.bat
@@ -1,6 +1,6 @@
 @echo off
 
-SET BUILD_DIR=C:\TBuild
+IF "%BUILD_DIR%"=="" SET BUILD_DIR=C:\TBuild
 SET LIB_DIR=%BUILD_DIR%\Libraries
 SET SRC_DIR=%BUILD_DIR%\tdesktop
 SET QT_VERSION=5_6_2
@@ -45,7 +45,7 @@ GOTO:EOF
     git clone https://chromium.googlesource.com/external/gyp
     cd gyp
     git checkout a478c1ab51
-    SET PATH=%PATH%;C:\TBuild\Libraries\gyp;C:\TBuild\Libraries\ninja;
+    SET PATH=%PATH%;%BUILD_DIR%\Libraries\gyp;%BUILD_DIR%\Libraries\ninja;
     cd %SRC_DIR%
     git submodule init
     git submodule update


stek29:
What was wrong with #3919?
And what about this?

> Previously you needed VS2017 Professional, %VS2017INSTALLDIR% points to the relevant path for your installation so it also works for Community/Enterprise

mrexodia:
Appveyor doesn't build those because reasons... Basically the architecture "x86" is named "Win32" and if you call vcvarsall it will not compile the solution.

auchri:
@mrexodia We can use the `APPVEYOR` environment variable to check if the build runs inside appveyor.

mrexodia:
I will update this pull request with a fix for the solution platform later... Basically you don't need to call vcvarsall.bat because Appveyor already runs from the visual studio command line. For some reason it didn't build yesterday, but can easily be fixed by manually calling msbuild

auchri:
Thank you for your efforts!

auchri:
@mrexodia I'll merge this one, please submit an follow up PR with the fix for the solution platform :)

Thank you for your first-time contribution! 👑

auchri:
@mrexodia For me the environment variable is not set, instead this registry key points to the install location: `HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\VisualStudio\SxS\VS7` (value 15.0).

Maybe someone can post the output of the following script to verify it is working on different setups?

```cmd
@ECHO OFF
setlocal ENABLEEXTENSIONS

set KEY_NAME="HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\VisualStudio\SxS\VS7"
set VALUE_NAME=15.0

FOR /F "usebackq skip=2 tokens=1-2*" %%A IN (`REG QUERY %KEY_NAME% /v %VALUE_NAME% 2^>nul`) DO (
    set ValueValue=%%C
)

if defined ValueValue (
    @echo Install path = "%ValueValue%"
) else (
    @echo %KEY_NAME%\%VALUE_NAME% not found.
)
```

mrexodia:
The point I was trying to make is that vcvarsall.bat shouldn't be called
because then the build breaks :)

I tested this in my earlier pull request, and it can be fixed by manually
calling msbuild for the "Win32" platform (because Appveyor only knows about
"x86"). I contacted Appveyor about this, but they don't seem interested in
fixing this.

PR #3930, opened 2017-09-27
Prefer AppIndicator if DBus interface is available

jhasse:
Starting with Ubuntu 17.10 the interface will be provided by an
extension (shipped by default):

https://github.com/ubuntu/gnome-shell-extension-appindicator

Legacy tray icons have been completely removed in GNOME 3.26. By
checking the interface, this will allow users of other distributions
with GNOME to also use Telegram's indicator with the extension.

I have tested the Qt code snippet (it correctly detects if the extension is enabled), but couldn't test it with Telegram, as setting up the build environment seems to require Ubuntu 12.04 (and I'm running Fedora).

Previous PR: #3928

### DIFF
--- a/Telegram/SourceFiles/platform/linux/linux_desktop_environment.cpp
+++ b/Telegram/SourceFiles/platform/linux/linux_desktop_environment.cpp
@@ -20,6 +20,8 @@ Copyright (c) 2014-2017 John Preston, https://desktop.telegram.org
 */
 #include "platform/linux/linux_desktop_environment.h"
 
+#include <QDBusInterface>
+
 namespace Platform {
 namespace DesktopEnvironment {
 namespace {
@@ -120,7 +122,8 @@ bool TryQtTrayIcon() {
 }
 
 bool PreferAppIndicatorTrayIcon() {
-	return IsXFCE() || IsUnity();
+	return IsXFCE() || IsUnity() ||
+	       (IsGnome() && QDBusInterface("org.kde.StatusNotifierWatcher", "/").isValid());
 }
 
 bool TryUnityCounter() {
--- a/Telegram/gyp/qt.gypi
+++ b/Telegram/gyp/qt.gypi
@@ -196,6 +196,7 @@
     '<(qt_loc)/include',
     '<(qt_loc)/include/QtCore',
     '<(qt_loc)/include/QtGui',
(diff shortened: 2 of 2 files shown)


xvitaly:
You shouldn't use AppIndicator in KDE. KDE should use Qt tray icon.

makruiten:
That’s why it checks if it’s running on Gnome, right?

xvitaly:
> That’s why it checks if it’s running on Gnome, right?

Oh, @jhasse has changed his commit. Now everything seems to be OK.

xvitaly:
> That’s why it checks if it’s running on Gnome, right?

Oh, @jhasse has changed his commit. Now everything seems to be OK.

john-preston:
Thanks!

Tray icons are a mess in Linux, it is a usual thing when changing something regarding them to get "Thanks, now it work!" and "WTF, it stopped working!" feedback at the same time.

I hope this one won't hit hard anyone 🙈

jhasse:
Yeah, hopefully these things can be fixed upstream in Qt in the future, so that Telegram could only use a Qt tray icon which does all the trickery involved. At the moment this is blocked by https://bugreports.qt.io/browse/QTBUG-63065 though.

Telegram/SourceFiles/platform/linux/linux_desktop_environment.cpp
```diff
@@ -120,7 +122,8 @@ bool TryQtTrayIcon() {
 }
 
 bool PreferAppIndicatorTrayIcon() {
-	return IsXFCE() || IsUnity();
+	return IsXFCE() || IsUnity() ||
+	       (IsGnome() && QDBusInterface("org.kde.StatusNotifierWatcher", "/").isValid());
```
3v1n0: This might still return null when Telegram is confined (such as when in flatpack or snap packages), thus I'd suggest to instead call `org.kde.StatusNotifierWatcher.IsNotificationHostRegistered()` or `org.kde.StatusNotifierWatcher.ProtocolVersion()`

jhasse: It "might" or "does"?

3v1n0: It does.... As this is just doing intrsopection and that's blocked when confined.

PR #4061, opened 2017-11-11
Fix build against OpenSSL 1.1

mymedia2:
Please merge this PR to compile against the latest OpenSSL 1.1

### DIFF
--- a/Telegram/SourceFiles/base/openssl_help.h
+++ b/Telegram/SourceFiles/base/openssl_help.h
@@ -55,8 +55,7 @@ private:
 
 class BigNum {
 public:
-	BigNum() {
-		BN_init(raw());
+	BigNum() : _data(BN_new()) {
 	}
 	BigNum(const BigNum &other) : BigNum() {
 		*this = other;
@@ -176,10 +175,10 @@ public:
 	}
 
 	BIGNUM *raw() {
-		return &_data;
+		return _data;
 	}
 	const BIGNUM *raw() const {
-		return &_data;
+		return _data;
 	}
 
 	bool failed() const {
@@ -193,7 +192,7 @@ public:
 	}
 
 private:
-	BIGNUM _data;
+	BIGNUM *_data = nullptr;
 	mutable bool _failed = false;
 
 };
--- a/Telegram/SourceFiles/mtproto/rsa_public_key.cpp
+++ b/Telegram/SourceFiles/mtproto/rsa_public_key.cpp
@@ -79,8 +79,8 @@ public:
 	}
 	Private(base::const_byte_span nBytes, base::const_byte_span eBytes) : _rsa(RSA_new()) {
 		if (_rsa) {
-			BIGNUM *n = BN_dup(openssl::BigNum(nBytes).raw());
-			BIGNUM *e = BN_dup(openssl::BigNum(eBytes).raw());
+			BIGNUM *n = openssl::BigNum(nBytes).raw();
+			BIGNUM *e = openssl::BigNum(eBytes).raw();
 			RSA_set0_key(_rsa, n, e, nullptr);
 			if (!n || !e) {
 				RSA_free(base::take(_rsa));
--- a/Telegram/SourceFiles/mtproto/rsa_public_key.cpp
+++ b/Telegram/SourceFiles/mtproto/rsa_public_key.cpp
@@ -29,6 +29,45 @@ Copyright (c) 2014-2017 John Preston, https://desktop.telegram.org
 using std::string;
 
 namespace MTP {
+namespace {
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+
+// This is a key setter for compatibility with OpenSSL 1.0
+int RSA_set0_key(RSA *r, BIGNUM *n, BIGNUM *e, BIGNUM *d) {
+	if ((r->n == nullptr && n == nullptr) || (r->e == nullptr && e == nullptr)) {
+		return false;
+	}
+	if (n != nullptr) {
+		BN_free(r->n);
+		r->n = n;
+	}
+	if (e != nullptr) {
+		BN_free(r->e);
+		r->e = e;
+	}
+	if (d != nullptr) {
+		BN_free(r->d);
+		r->d = d;
+	}
+	return true;
+}
+
+// This is a key getter for compatibility with OpenSSL 1.0
+void RSA_get0_key(const RSA *r, const BIGNUM **n, const BIGNUM **e, const BIGNUM **d) {
+	if (n != nullptr) {
+		*n = r->n;
+	}
+	if (e != nullptr) {
+		*e = r->e;
+	}
+	if (d != nullptr) {
+		*d = r->d;
+	}
+}
+
+#endif
+}
+
 namespace internal {
 
 class RSAPublicKey::Private {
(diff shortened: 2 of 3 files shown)


auchri:
You have to fix the failing tests to get this merged.

mymedia2:
Which tests do you have in mind? Travis? To be honest, I don't know what that error means and how to fix it ._.

> Qt was built, please restart the job :(

auchri:
@mymedia2 I've restarted the buils, but there are build errors.

mymedia2:
Well, those errors are related to ffmpeg, and I have no idea why they appeared just now, since I didn't modify that file. Just in case, I quote the arisen errors, perhaps somebody will help to fix them.

    [ 57%] Building CXX object CMakeFiles/Telegram.dir/home/travis/build/telegramdesktop/tdesktop/upstream/Telegram/SourceFiles/media/media_audio_capture.o
    /home/travis/build/telegramdesktop/tdesktop/upstream/Telegram/SourceFiles/media/media_audio_capture.cpp: In member function ‘void Media::Capture::Instance::Inner::onStart()’:
    /home/travis/build/telegramdesktop/tdesktop/upstream/Telegram/SourceFiles/media/media_audio_capture.cpp:266:29: error: ‘CODEC_FLAG_GLOBAL_HEADER’ was not declared in this scope
       d->codecContext->flags |= CODEC_FLAG_GLOBAL_HEADER;
                                 ^~~~~~~~~~~~~~~~~~~~~~~~
    /home/travis/build/telegramdesktop/tdesktop/upstream/Telegram/SourceFiles/media/media_audio_capture.cpp:279:58: error: ‘CODEC_CAP_VARIABLE_FRAME_SIZE’ was not declared in this scope
      d->srcSamples = (d->codecContext->codec->capabilities & CODEC_CAP_VARIABLE_FRAME_SIZE) ? 10000 : d->codecContext->frame_size;
                                                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~

john-preston:
Yeah, it just broke :( Even on the current dev branch.

john-preston:
Travis build isn't bound to specific commit of ffmpeg repository, so it breaks when ffmpeg removes or deprecates something. I think in the near future I'll update ffmpeg version together with fixing the version in the travis script.

john-preston:
I've deployed a fix for travis build. Please rebase your changes and see if the checks pass.

auchri:
```
Linking CXX executable Telegram
CMakeFiles/Telegram.dir/home/travis/build/telegramdesktop/tdesktop/upstream/Telegram/SourceFiles/mtproto/auth_key.o: In function `MTP::aesCtrEncrypt(void*, unsigned int, void const*, MTP::CTRState*)':
/home/travis/build/telegramdesktop/tdesktop/upstream/Telegram/SourceFiles/mtproto/auth_key.cpp:113: undefined reference to `CRYPTO_ctr128_encrypt(unsigned char const*, unsigned char*, unsigned long, void const*, unsigned char*, unsigned char*, unsigned int*, void (*)(unsigned char const*, unsigned char*, void const*))'
collect2: error: ld returned 1 exit status
make[2]: *** [Telegram] Error 1
make[1]: *** [CMakeFiles/Telegram.dir/all] Error 2
make: *** [all] Error 2
```

auchri:
```
/home/travis/build/telegramdesktop/tdesktop/upstream/Telegram/SourceFiles/mtproto/rsa_public_key.cpp: In function ‘int MTP::{anonymous}::RSA_set0_key(RSA*, BIGNUM*, BIGNUM*, BIGNUM*)’:
/home/travis/build/telegramdesktop/tdesktop/upstream/Telegram/SourceFiles/mtproto/rsa_public_key.cpp:37:22: error: suggest parentheses around ‘&&’ within ‘||’ [-Werror=parentheses]
  if (r->n == nullptr && n == nullptr || r->e == nullptr && e == nullptr) {
      ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
```

mymedia2:
Yay, all checks passed! Thank you, @john-preston and @auchri, for the help! Now the pull request can be merged.

auchri:
Please rebase to the dev branch.

mymedia2:
@auchri, please restart the Travis check. It rebuilt Qt again...

john-preston:
I've restarted it, but I guess it will fail like the dev branch travis build fails right now :/ But I'll check and merge that if everything is fine. Thanks.

Telegram/SourceFiles/base/openssl_help.h
```diff
@@ -55,8 +55,7 @@ class Context {
 
 class BigNum {
 public:
-	BigNum() {
-		BN_init(raw());
+	BigNum() : _data(BN_new()) {
```
john-preston: You're now allocating BIGNUM, but I don't see any BN_free() calls. Why is it better to allocate dynamically? If it is needed for some reason, then BigNum class should be made self-consistent. Right now if I create a BigNum variable I'll leak memory.

Telegram/SourceFiles/mtproto/rsa_public_key.cpp
```diff
@@ -40,9 +79,10 @@ class RSAPublicKey::Private {
 	}
 	Private(base::const_byte_span nBytes, base::const_byte_span eBytes) : _rsa(RSA_new()) {
 		if (_rsa) {
-			_rsa->n = BN_dup(openssl::BigNum(nBytes).raw());
-			_rsa->e = BN_dup(openssl::BigNum(eBytes).raw());
-			if (!_rsa->n || !_rsa->e) {
+			BIGNUM *n = BN_dup(openssl::BigNum(nBytes).raw());
```
john-preston: If I don't miss anything we leak two BIGNUMs here by creating a temporary BigNum, duplicating its data and then just forgetting it.

Telegram/SourceFiles/base/openssl_help.h
mymedia2: BN_clear_free is already called a few lines below in the destructor. It properly frees memory for the pointer to BIGNUM allocated by BN_new. I have to call BN_new because BN_init became a private function in OpenSSL 1.1. So now we should work with pointers.

Telegram/SourceFiles/mtproto/rsa_public_key.cpp
mymedia2: Well, it seems there was redundant duplicating. I think it'll work without BN_dup, but I tried to leave the behavior untouched.

Telegram/SourceFiles/base/openssl_help.h
mymedia2: The BN_clear_free function correctly realizes the memory allocated for a pointer to the BIGNUM structure. Look at an implementation https://github.com/openssl/openssl/blob/OpenSSL_1_0_1-stable/crypto/bn/bn_lib.c#L242
So there's no memory leaks.

john-preston:
@mymedia2 About the leaks - sorry, I didn't see the destructor in the diff (because it wasn't changed), but it does indeed free everything. My mistake.

But why do you need to move from BN_init() to BN_new()? Because the BN_init() way isn't supported in OpenSSL 1.1 anymore?

About the key initialization - it seems that the way you done it in the last commit will crash, because it frees memory twice - once in BigNum destructor and once in RSA_set0_key(), because it takes ownership of the passed BIGNUM.

The line `BIGNUM *n = openssl::BigNum(nBytes).raw();` gets a dangling pointer `n`, because you create BigNum, get its BIGNUM pointer and then destroy this BigNum, invalidating the pointer.

It will work without BN_dup() if you add a "takeRaw()" method to BigNum, that will return the pointer to BIGNUM and forget it inside BigNum itself, so that it doesn't free anything in destructor.

john-preston:
Yes, I've checked, the new approach does crash :(

The proposed `takeRaw()` method should be something like `BIGNUM *takeRaw() { return base::take(_data); }`

This will return the contained BIGNUM* and set `_data` to nullptr, so destructor will not free anything and the ownership will be transferred to the caller of `takeRaw`, exactly what we need for `RSA_set0_key`. Like `unique_ptr::release` does.

mymedia2:
> But why do you need to move from BN_init() to BN_new()? Because the BN_init() way isn't supported in OpenSSL 1.1 anymore?

Exactly so. OpenSSL 1.1 doesn't support this function.

> ...
> The proposed takeRaw() method should be something like BIGNUM *takeRaw() { return base::take(_data); }

Thank you for the explanation and fixing the patch.

PR #4124, opened 2017-12-07
Avoid codegen_style generate equals forward declarations in an header file

chaplin89:
This is just a little edit in order to make sure generated files does not contains multiple equals forward declarations, in order to improve a little bit the readability of the generated files.

### DIFF
--- a/Telegram/SourceFiles/codegen/style/generator.cpp
+++ b/Telegram/SourceFiles/codegen/style/generator.cpp
@@ -22,6 +22,8 @@ Copyright (c) 2014-2017 John Preston, https://desktop.telegram.org
 
 #include <memory>
 #include <functional>
+#include <list>
+#include <algorithm>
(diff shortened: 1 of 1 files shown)


Telegram/SourceFiles/codegen/style/generator.cpp
```diff
@@ -609,10 +611,16 @@ bool Generator::writeStructsForwardDeclarations() {
 	}
 
 	header_->newline();
-	bool result = module_.enumVariables([this](const Variable &value) -> bool {
+
+	std::list<QString> alreadyDeclaredTypes;
```
stek29: Why not std::set or something else with log time?

chaplin89: You're right. I'll close this and open another PR because there were some issues with my git config.

PR #4130, opened 2017-12-08
Avoid generating multiple time the forward declarations in the headers generated by codegen_style.

chaplin89:
As the title say, this is just a little edit in order to improve the readability of headers generated by codegen_style.


john-preston:
Thanks.

PR #4218, opened 2017-12-21
Add reply shortcut

### LINKED ISSUE #4208: Feature request: Reply shortcut
Something like ctrl+up/down would be really useful.
I made a simple patch (https://github.com/0e4ef622/tdesktop/commit/02221b9e64279686c368e6b46c3ae596d68dda28), and me and my friends find it very useful, but I read that you don't accept pull requests with new features, so it's there as a reference to what it should do.

0e4ef622:
Fixes #4208


john-preston:
Thanks! I'll add this together with some sort of fast replies in groups.

PR #4230, opened 2017-12-23
Added Release information to the AppStream file

Pointedstick:
Added Release information to the AppStream file, pre-populated with release notes from the latest release (1.2.3 alpha).

This brings the following benefits for Telegram developers:
- Greater control over release messaging and communication
- Users get changelogs without downstream packagers having to manually parse your changelog.txt file (which is extra work so they might not do it)
- Software Center apps display version numbers without downstream packagers having to manually find out what it is (which is extra work so they might not do it)

It also requires some work going forward: every new release will require a new `<release>` tag in the AppStream file.

Relevant documentation for this part of the AppStream spec: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-description

### DIFF
--- a/lib/xdg/telegramdesktop.appdata.xml
+++ b/lib/xdg/telegramdesktop.appdata.xml
@@ -22,4 +22,13 @@
     <provides>
         <binary>telegram-desktop</binary>
     </provides>
+    <releases>
+        <release version="1.2.3 alpha" date="2017-12-17">
+            <description>
+                <ul>
+                    <li>Several crash fixes</li>
+                </ul>
+            </description>
+        </release>
+    </releases>
 </component>


Pointedstick:
I'll mention that If folks don't like the idea of release notes, the `<description>` tag is optional. But marking the versions is really nice for downstream packagers.

stek29:
I totally agree with [Eduard Braun from Inkscape](https://bugs.launchpad.net/inkscape/+bug/1739898/comments/2).

- Hardcoding version numbers in a file is not a good idea
- Adding yet another place to maintain releases isn't worth it imo: there are git tags, website, in-app updates&changelogs, stores 
- TDesktop has no official packages or any way of distribution except a binary on site/github releases (for Linux)
- Whoever makes them is expected to parse version from git tags at least
- The AppStream file was initially submitted in a PR

Pointedstick:
The idea behind AppStream is to make life easier for those people who decide to package Telegram. But it's not just for them: when you make their lives easier, you also increase the consistency and quality with which your app is presented to users.

Without this kind of data, each downstream Linux source has to piece together release notes, app descriptions, and screenshots themselves, by hand. Some might do a better job than others. It dilutes your branding, and you lose a measure of control over how your app is presented to users. For example KDE Neon packaging doesn't even have any screenshots at all.

It's true that hardcoding rapidly-changing information in the file probably isn't ideal. Generating it as part of your build system would make life easy--but of course that's tooling you would have to write and maintain, and I totally understand why that might not seem like the best way to spend your time. :slightly_smiling_face: I'm not an AppStream developer myself, and I'm not sure if they have anything to make this process easier. Paging Dr. @ximion! :smile:

Pointedstick:
Here's another "carrot" reason, from https://bugzilla.gnome.org/show_bug.cgi?id=779839#c5, it seems that the GNOME Software app rewards apps that have added release information to their AppStream files by featuring them for 2 months after each release. I'm going to propose adding something similar to KDE Discover, too. So GNOME users would see Telegram more prominently if you add this information.

xvitaly:
@Pointedstick And who will update this information in appstream file? You, before each release?

Pointedstick:
Like I said, ideally it would be automatically populated by a script in your build system, probably as a part of your release process. AppStream is primarily @ximion's project, so I'm hoping he can chime in with some good options here.

TingPing:
How some other projects operate is appstream is the canonical location for user facing changelogs and then any other changelogs (websites, NEWS file, etc) are generated *from* that information. The opposite could be done ofc, it is fairly basic xml in either direction.

ximion:
@Pointedstick I wanted to add a feature to appstreamcli to convert changelogs from a YAML-based format or other formats into AppStream XML for a while now, but the main reason for adding that so far was that writing XML is a bit annoying for long changelogs.
Sadly, I haven't found time for that feature yet, especially since the question of how to translate the information in the changelog is still open (Metainfo files - including the changelog - can be fully translated, and if I add such a converter, I want to preserve that feature).

So yeah, I hope we can improve the situation for people using release information in their metainfo files soon, both for writing it as well as for maintaining it (because people rarely want to maintain release information in many places, Telegram isn't an exception here - the projects that adopted the release tag switched to writing release information to the Metainfo file first and generate all the other files from that).

Me not having an "official" way to convert changelogs and merge information yet doesn't mean that one couldn't write a simple project-specific script to do that though ;-)

john-preston:
@Pointedstick How is it done usually? Do you provide all changelogs for all releases (298 at this point) in this file? Can it replace [changelog.txt](https://github.com/telegramdesktop/tdesktop/blob/dev/changelog.txt) in its current form? Or should it include only the latest one?

TingPing:
@john-preston It can replace other formats if you want, `appstream-util` even has a command to convert it to a plain text file for released tarballs if you want. It should contain multiple releases as the app store can for example show you the past 2 releases if you haven't updated in a while but converting all 300 releases is surely overkill atm.

john-preston:
It should be done in some scripts, so that all this could be automated. In the best case there should be some script that will generate the valid block inside the appstream file from the changelog.txt (or vice versa), so that it even could be called any number of times and the call could be added to build/set_version script (both .bat and .sh). Currently this adds work for preparing each release, which is not cool =(

john-preston:
There is no sense in adding it one time with a pull request and currently there are no scripts in place to perform this automatically.

PR #4233, opened 2017-12-24
Fix handling of dashes in the style directory path

ForNeVeR:
Now Telegram Desktop could be built if a path to the source directory contains dashes.

For example, `C:\my-sources\Telegram`

Without that fix, codegen assumes, for example, that `C:\my` is a path to the image, and `sources\Telegram\someimage` is a modifier.


john-preston:
Thanks!

Issue #3508, opened 2017-06-09
-autostart and -startintray doesn't work on Ubuntu


vladislav805:
Want to add Telegram to autorun:
### Steps to reproduce
1. Open Terminal
2. `/opt/telegram/Telegram -autostart --` or `/opt/telegram/Telegram -startintray --`

### Expected behaviour
The Telegram was supposed to start in the system tray, without opening the window.

### Actual behaviour
In first case (autostart) nothing happened, process complete immediately. In second case (startintray) the icon appeared in the tray is clickable, but clicking on "Open Telegram" nothing happened. However, notifications about new messages "come" in correctly.

### Configuration
**Operating system:** Ubuntu 16.04 LTS

**Version of Telegram Desktop:** 1.1.7

**Used theme**: doesn't matter, same behavior

<details><summary><b>Logs</b>:</summary>
Insert logs here (if necessary)

</details>

john-preston:
@vladislav805 It was not supposed to work there (was not tested at all, this options were made for Windows version), but I'll try to debug it if I'll have some time.

A pull request with a fix for that bug and comments about the cause of it is also welcome.

vladislav805:
So far only see that the Linux version is missing features `psAutoStart` and `psSendToMenu`

Versions: [Linux (specific_linux.cpp:622)](https://github.com/telegramdesktop/tdesktop/blob/81e728ada01713f562b0d7694eec57ed954e1786/Telegram/SourceFiles/platform/linux/specific_linux.cpp#L622), [Windows (specific_win.cpp:735)](https://github.com/telegramdesktop/tdesktop/blob/81e728ada01713f562b0d7694eec57ed954e1786/Telegram/SourceFiles/platform/win/specific_win.cpp#L735)

In Windows version `_manageAppLnk` create a `.lnk` file in special directory, where programs run automatically after startup.

But how add program to autorun in Linux I don't know...

Aokromes:
#422 ?

Issue #3391, opened 2017-05-13
-startintray on Ubuntu 17.04 doesn't open window


tonyedgecombe:
### Steps to reproduce
1. telegram-desktop -startintray
2. Right click on tray icon
3. Select Open Telegram

### Expected behaviour
Telegram window should open.

### Actual behaviour
Nothing

### Configuration
Ubuntu 17.04

**Version of Telegram Desktop:**
1.0.29

**Used theme**:

<details><summary>**Logs**:</summary><pre>
Qt: Session management error: Could not open network socket
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for 4294967295, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for 4294967295, skipping unlock

(telegram-desktop:21175): libappindicator-CRITICAL **: app_indicator_set_icon_full: assertion 'IS_APP_INDICATOR (self)' failed
</pre>

TheAppService:
#422

SnuggIes:
I can confirm this bug occurs on Ubuntu 17.04.

Please see this video demonstrating this bug: https://www.youtube.com/watch?v=Lwx1g51Kv4E
Telegram **only** opens at 0:27.

**The only way to get the Telegram window to open is to:**
_1. Click on tray icon
2. Click "Minimize to tray"
3. Click on tray icon again
4. Click "Open Telegram"_

**System Configuration:**
![image](https://user-images.githubusercontent.com/26250962/29100887-82bd7c8c-7ca6-11e7-8d5b-493d4ce032a8.png)

Aokromes:
Duplicate of #422

Issue #2123, opened 2016-06-08
Telegram -startintray does not work in wayland. In wayland there is no tray icon at all


rbieb:
### Steps to reproduce
1. Run wayland
2. Run telegram -startintray
3. Check if you see a tray icon
### Expected behaviour

Tray icon should appear as expected
### Actual behaviour

There is no tray icon
### Configuration

Fedora 24 with gnome

**Version of Telegram Desktop:**
0.9.49

john-preston:
@rbieb Well, -startintray can possibly work only if the tray icon is working when you just run Telegram. Looks like the tray icon is not supported in Tdesktop for this desktop env. I use standard Qt tray icon QSystemTrayIcon class — perhaps it is a Qt issue that it is not working in wayland.

stek29:
@john-preston isn't it Gnome?

john-preston:
@stek29 For some Qt version I had to disable Qt tray for gnome (and pantheon), perhaps I need to test it once more with the current Qt version (5.6), maybe it works fine already.

diazbastian:
@rbieb Could you be more specific with your system configuration?
The tray icon don't work in gnome-shell using both X11 or Wayland unless you have installed the package "libappindicator"

Tested on:
• Fedora 23 Workstation (Gnome) using X11 and Wayland session
• Fedora 24 Workstation beta (Gnome) using X11 and Wayland session
• Telegram Desktop 0.9.49

rbieb:
I did have libappindicator installed.

diazbastian:
@rbieb It would be good to attach the log, because I can not replicate this error.

rbieb:
I would if I could, but I'm no longer running gnome or wayland, thanks for looking into it though.

GreenRiverRUS:
@diazbastian I can help to replicate error. I use Gnome on Wayland on Arch Linux and has no telegram icon in system tray. Libappindicator is installed:
```
~/$ sudo pacman -Qs libappindicator
local/libappindicator-gtk2 12.10.0-10
    Allow applications to export a menu into the Unity Menu bar
local/libappindicator-gtk3 12.10.0-10
    Allow applications to export a menu into the Unity Menu bar
local/libappindicator-sharp 12.10.0-10
    Allow applications to export a menu into the Unity Menu bar
```

Tomoms:
[...]

Issue #1820, opened 2016-03-25
Argument "-startintray" - Is it a known issue?


jacklaag:
### Steps to reproduce
1. Add Telegram to autostart with parameter "-startintray"

### Expected behaviour
Start in tray --> no window is opened

### Actual behaviour
A broken (transparent) Telegram window opens on startup

### Configuration
**Operating system: elementary OS**

**Version of Telegram Desktop: 0.9.33**

Telegram Support answer when I asked them about starting in tray:
"Hello! There was a command line argument "-startintray", but it worked somehow weird in Linux, so I'm not sure it will work for you."

Which non Linux user would possibly use this argument anyway? On Windows Telegram I can just klick a button in the settings for this.

john-preston:
@jacklaag Well, this command line parameter was left in Linux version, but I didn't manage to make it work properly. I'm not sure I know how to fully create good window in Linux without showing it, while this can be easily done in Windows (so this setting even made its way to Settings there).

jacklaag:
I have an idea for a dirty solution which should work everywhere:
Making -startintray open a smal dummy window (mabe with text "Telegram is connecting..." or similar) which then closes itself asap. This should then leave Telegram in tray and running in the backround just like if I did autostart it without parameters and then manually closed it.

btw: I just googled this problem and found a guy who had a similar idea to mine: http://askubuntu.com/a/663288

auchri:
> Is it a known issue?

Yes. Please search for existing issues before creating new ones.

#768 #512

DaFri-Nochiterov:
@auchri, you need bot to search existing issues and close duplicates 😀

auchri:
@DaFri-Nochiterov yes, that would be awesome :+1:

stek29:
@DaFri-Nochiterov Or GitHub needs improved duplicate search (like stack overflow sites)

DaFri-Nochiterov:
@stek29 But and this don't give warranties to people stop making duplicates

stek29:
@DaFri-Nochiterov Maybe it's easier to make people smarter?
(jk, lol, bots are better)
Also, it's turning into very off-topic conversation. Where is "Continue in chat" button?)

jacklaag:
Oh.. I didn't realize that I can comment closed issues.
If you look closely at my answer to @john-preston I am not only pointing at the problem, which has been discussed before, but also giving a possible solution to fix it. I knew pretty well that this issue is a known one (obviously...)

[...]

PR #4268, opened 2018-01-01
#948 Add option "Start in system tray" in "Settings" window

### LINKED ISSUE #948: Add option "Start in system tray" in "Settings" window

Thanks a lot! ;-)

Musikolo:
Implementation of feature request closes #948. Built and tested against latest upstream dev branch. 

If anything needs to be improved, please just let me know.


Telegram/Resources/langs/lang_es.strings
```diff
@@ -315,7 +315,8 @@ Copyright (c) 2014-2017 John Preston, https://desktop.telegram.org
 "lng_settings_workmode_tray" = "Mostrar el icono";
 "lng_settings_workmode_window" = "Mostrar icono de la barra de tareas";
 "lng_settings_auto_start" = "Abrir Telegram cuando el sistema inicie";
-"lng_settings_start_min" = "Abrir minimizada";
+"lng_settings_start_min" = "Iniciar minimizado";
+"lng_settings_start_in_tray" = "Iniciar en la bandeja de sistema";
```
stek29: You should only add English translation

Musikolo: Does this mean you want me to remove Spanish translations? 

I just want to confirm.

Thank you.

Musikolo:
For some reason CI test 5 failed while building 3rd party dependencies (Qt & range-v3) as you can see at https://travis-ci.org/telegramdesktop/tdesktop/builds/323815838

Previous built was completely successful though: https://travis-ci.org/telegramdesktop/tdesktop/builds/323729076

I don't think this error is related to my last code change, but I'm missing something, please, just let me know.

Thank you!

auchri:
Everything is ok with the builds:

screenshot_20180101-205535

john-preston:
@auchri I've restarted it.

Telegram/Resources/langs/lang_es.strings
auchri: yes.

Musikolo:
@auchri, the changes you are requested are done.

Please, let me know if anything else is needed.

Thank you!

john-preston:
1. This startInTray option currently is controlled by "-startintray" command line argument. After your changes the command line argument is ignored and the value is always read from settings.

This is not good, because "-startintray" argument is passed from "launch when system starts" shortcut on Windows and it is controlled by a separate "launch minimized" setting, which will be just ignored in case your changes are added.

2. This startInTray option and "-startintray" command line argument were added for this specific case - for "launch when system starts" + "launch minimized" on Windows.

It works there, but it was not tested and is known not to work in other systems and configurations, so it is a bad idea to give users a shiny checkbox in settings for something that perhaps doesn't work for them or works bad. See #512 #2123 #768 #3508 #3391 #1326 #422 #407 #511 #1820

With all this said I remind you that "we do not merge any pull requests that introduce any new user interface elements" was added to our contributing guidelines for a reason :(

Musikolo:
@john-preston, thank for your feedback.

I understand your concerns about the possible unwanted side effects my implementation could cause. Being honest, I have only tested the code in my Arch Linux box with KDE. Per the comments we have at #948, many people confirmed the `-startintray` command line option works for the Linux platform overall, plus I can confirm it works for me too. 

Furthermore, this feature requested in the issue #948 seems to be claimed or wanted by a relatively big number of users over a long period of time (~ 2.5 years). So, I would like to know your advise on how we can find a way of merging both worlds, i.e., do not alter the existing functionality while incorporating a settings option that improves the user experience. 

- Is there any possible approach I could follow that you feel comfortable with?
- Would it be an option to make the code I implemented Linux-platform specific (i.e., only available for the Linux platform)?

Looking forward to your feedback.

Thank you so much for your time! ;-)

Aokromes:
@Musikolo  if you change:

Implementation of feature request #948.

To

Implementation of feature request closes #948.
It will autocloses that ticket :)

Musikolo:
@Aokromes I don't want to close the issue #948 until there is an agreed solution to it. My PR hasn't been accepted yet, so it cannot be considered (yet) a solution for it. I'm waiting for @john-preston and/or other developers to provided me with guidance on how I could help it get such an issue solved.

Thanks for your suggestion!

Aokromes:
I mean, when this pr is merged with the suggested text it will close that issue :)

Musikolo:
@Aokromes, OK, I see now. It should be done now. ;)

I hope this feature is merged in the near future. 

Thank you!

Musikolo:
@john-preston / @auchri, I've been using Telegram Desktop with my patch seamlessly for over a month.  Following your last comments, the only scenario that changes the existing behavior requires the next steps:

1. Start Telegram Desktop.
1. Check the option "_Launch in system tray_" (new checkbox).
1. Exit Telegram Desktop.
1. Start Telegram from command line without the switch "`-startinsystemtray`".

In this scenario, the application shows in the system tray, but it does so because you manually selected an option to do exactly that, start in the system tray. If you want the previous behavior back, all you need to do is to uncheck such an option (_Launch in system tray_).

From my point of view, there is nothing wrong in what the patch does. I believe there are no other behavioral changes beyond the scenario described above. After 2.5 years waiting for a solution, I (and a few more #948 ) would truly like my patch incorporated into the upstream code base. If you don't feel comfortable with the code changes, I'm willing to update it the way consider necessary, provided I get some guidance on what's expected.

I appreciate your time and good job.

Looking forward to your reply!

stek29:
Maybe you'd implement changes @auchri has requested while waiting at least?

Musikolo:
@stek29, the only changes @auchri requested me [were implemented](https://github.com/telegramdesktop/tdesktop/pull/4268#issuecomment-354920857) a month ago in [this commit](https://github.com/telegramdesktop/tdesktop/pull/4268/commits/b0a0348d75f437cf3280e000e45819b1eed31ea4). However, if I've missed anything, I would be more than happy to get it done asap. 

Please, let me know, if this is the case.

Thank you!

stek29:
@Musikolo oh, sorry, I've missed that somehow.

abixadamj:
Strange behaviour on Mint 18.2 64bit
I have Installed 1.2.6 version yesterday. Everything works perfectly. I have checked 'Show tray icon' and Telegram runs OK. I have added Telegram to Autostart in Mint. Today I have started computer, and Telegram opened in normal windows, without system tray icon.
When I go to Preferences, uncheck 'Show tray icon" and then check it again, icon apperad in tray.

Musikolo:
As I mentioned 2 months ago, if there is anything I can do to meet your requirements and get this PR merged, please, let me know.

Will I ever get any reply or feedback? :-/

john-preston:
As I mentioned earlier a minimum requirement is that the current workflow on other platforms won't be broken. For that you must always start in tray in case "-startintray" argument was passed, no matter what is saved in the settings. This could be done if "-startintray" and your setting flag were held in different variables, not in the same one.

Also I think you need to merge the submodules as well (maybe run "git submodule update" after "git merge upstream/dev" or smth like that), because currently your patch show changes in submodules, which there should not be.

Musikolo:
Closing this PR since it's being replaced by new [PR #4782](https://github.com/telegramdesktop/tdesktop/pull/4782).

PR #4291, opened 2018-01-09
Change default when pinning to not notify all

JuanPotato:


JuanPotato:
https://github.com/Streetwalrus/dotfiles/blob/master/scripts/telegram-desktop

For those in the future who may happen upon this thread and it is closed, use this script.

john-preston:
@JuanPotato Sorry, but this is consistent across other official apps, I can't disable that in my app that way.

PR #4331, opened 2018-01-21
Update building-xcode.md

DanyaDaro:
Add depot_tools in $PATH, otherwise "crashpad" not compile.

### DIFF
--- a/docs/building-xcode.md
+++ b/docs/building-xcode.md
@@ -89,6 +89,7 @@ Go to ***BuildPath*** and run
 
     git clone https://chromium.googlesource.com/external/gyp
     git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
+    export PATH=`pwd`/depot_tools:"$PATH"    
     cd gyp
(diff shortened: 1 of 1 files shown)


docs/building-xcode.md
```diff
@@ -89,6 +89,7 @@ Go to ***BuildPath*** and run
 
     git clone https://chromium.googlesource.com/external/gyp
     git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
+    export PATH=`pwd`/depot_tools:"$PATH"
```
stek29: export PATH="$PWD/depot_tools:$PATH" <3

stek29: `export PATH="$PWD/depot_tools:$PATH"` <3

stek29:
It's not really needed because of https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/gyp/refresh.sh#L31 I guess, but it won't hurt.

stek29:
Thank you :)

PR #4345, opened 2018-01-25
Fix build instructions

alexandernst:
Current instructions won't work if followed blindly. Also, there is a fix for xcode-select using the wrong binaries.


stek29:
Thanks!

PR #4382, opened 2018-02-04
linux: use $HOME to determine actual user path

3v1n0:
It could be different in confined environments.

Reading this from `passwd` isn't always reliable.


john-preston:
Thanks!

PR #4398, opened 2018-02-08
linux-desktop-environment: detect Ubuntu properly enabling features

3v1n0:
In Ubuntu (running in GNOME) we support AppIndicator's and
Launcher counters still.


john-preston:
Thanks!

PR #4399, opened 2018-02-09
Linux ARM compile fixes

3v1n0:
Make telegram to compile properly in ARM hosts.

This also needs https://github.com/grishka/libtgvoip/pull/40


john-preston:
Thanks!

PR #4429, opened 2018-02-18
Let emoji suggestions be case insensitive

### LINKED ISSUE #3748: Improvement to emoji autocomplete
After #627 was introduced, I thought about two more things to improve the emoji autocomplete feature:

* [X] Use case insensitive match - for example, I'm using Hebrew as my main language and the fastest way to write a word in English would be with Capslock or Shift - I believe it would be correct to any non latin language.
* [ ] Add other keywords for the emojis, such as the keywords in [this list](https://github.com/muan/emojilib/blob/master/emojis.json) or [this one](http://unicode.org/emoji/charts/emoji-list.html)

### LINKED ISSUE #3985: Case Invariant Emoji Suggestions
As a user that chats in telegram in a language other than English, it is quite frustrating to switch to English every time I want to see the emoji suggestions.
It would be lovely if these suggestions will appear even if I type :HEART: instead of :_heart: (without the underscore), so I could just hold the SHIFT key down while typing an emoji search while the input language remains <insert non-english language here>.

Cheers!

pentix:
## What?
All this patch does is to allow emojis beeing suggested - even if their names were spelled using uppercase letters.

telegram_patch

## Why?
As mentioned in #3985 and in #3748 it can happen
that users of a non-US keyboard type the emoji's name
in uppercase letters because the fastest way of typing e.g. `:relaxed:` involves using the
[Shift] key, which can lead (due to fast/bad typing) to writing the first
letters of the emoji's name in capitals. Telegram does currently not come up
with a suggestion at this point and one is forced to correct the first few letters - making
it quite inconvenient.

## Conflicts
This patch should not be in conflict with anything. Also, the emoji names used are case-insensitive. (Meaning that there cannot be a "wrong" suggestion due to this patch)

## Consequences 
This closes #3985 and partially resolves #3748 

## Tests done
Compilation is successful and the suggestion works as expected. (See the GIF in the first paragraph)

### DIFF
--- a/Telegram/SourceFiles/chat_helpers/emoji_suggestions_widget.cpp
+++ b/Telegram/SourceFiles/chat_helpers/emoji_suggestions_widget.cpp
@@ -401,7 +401,7 @@ QString SuggestionsController::getEmojiQuery() {
 			}
 			position -= from;
 			_queryStartPosition = from;
-			return fragment.text();
+			return fragment.text().toLower();
 		}
 		return QString();
 	};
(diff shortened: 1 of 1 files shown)


john-preston:
Thanks, I'll think this should work fine.

Issue #4418, opened 2018-02-17
Build fails when linking CXX executable Telegram


R32fanboy:
Running on Ubuntu 16.04 LTS updated with the latest patches. 
Build was stored in the directory /home/user/Tbuild/ with the two branches being /tdesktop and /Libraries
### Steps to reproduce
1. Built dependencies with no Errors.
2. ran make on /Tbuild/tdesktop/out/release and /Tbuild/tdesktop/out/Debug and ended up with the same error.
![telefail2](https://user-images.githubusercontent.com/32889334/36346324-34a02b64-1402-11e8-87c4-833b92eebfa4.png)

/usr/local/tdesktop/Qt-5.6.2/lib/libQt5Network.a(qsslsocket_openssl_symbols.o): In function `q_EC_curve_nist2nid(char const*)':
/home/slighthoof/Tbuild/Libraries/qt5_6_2/qtbase/src/network/ssl/qsslsocket_openssl_symbols.cpp:426: undefined reference to `EC_curve_nist2nid'
collect2: error: ld returned 1 exit status
CMakeFiles/Telegram.dir/build.make:13117: recipe for target 'Telegram' failed
make[2]: *** [Telegram] Error 1
CMakeFiles/Makefile2:85: recipe for target 'CMakeFiles/Telegram.dir/all' failed
make[1]: *** [CMakeFiles/Telegram.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2

Looks like it's an error with the installer expecting an object from OpenSSL to be there and it is not, or it is not in the place it's looking for. OpenSSL was installed in correct order and is completely up to date in the repository.

pentix:
Hi, I faced the same error just yesterday. I found this

https://github.com/telegramdesktop/tdesktop/issues/2519#issuecomment-253171003

Make sure to checkout `OpenSSL_1_0_2-stable` before compiling OpenSSL. The documentation is wrong there for anything later then Ubuntu 12.04 🙄

R32fanboy:
That did the trick! We are up and running, thanks for the tip!

pentix:
Hi @auchri 

Because in the referred thread, the same was declined by @john-preston:

@EhsaanF 
> Nice job, John. Please change the docs, too. :)

@john-preston 
> Well, the docs are for Ubuntu 12.04, it works there I suppose.

🤷‍♂️

Issue #2519, opened 2016-10-11
Build failure on Linux


EhsaanF:
Build using GYP/CMake will be failed on Linux.
```
[ 76%] Linking CXX executable Telegram
/usr/local/tdesktop/Qt-5.6.0/lib/libQt5Network.a(qsslsocket_openssl_symbols.o): In function `q_EC_curve_nist2nid(char const*)':
/home/ehsaan/TBuild/Libraries/qt5_6_0/qtbase/src/network/ssl/qsslsocket_openssl_symbols.cpp:426: undefined reference to `EC_curve_nist2nid'
collect2: error: ld returned 1 exit status
CMakeFiles/Telegram.dir/build.make:12336: recipe for target 'Telegram' failed
make[2]: *** [Telegram] Error 1
CMakeFiles/Makefile2:238: recipe for target 'CMakeFiles/Telegram.dir/all' failed
make[1]: *** [CMakeFiles/Telegram.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2
```
Needed a fast fix.

john-preston:
It is not failing in travis (most of the times, there is some bug with dependencies, but that has nothing to do with openssl linking problems), see for example:

https://travis-ci.org/telegramdesktop/tdesktop/jobs/166813384

I don't know why it fails in your system specifically :( Are you using custom built openssl or your system openssl (libssl / libcrypto) for linking? Some guys from stackoverflow have the same problem:

http://stackoverflow.com/questions/35627998/linking-error-libqt5network-so-undefined-reference-to-ec-curve-nist2nid

EhsaanF:
@john-preston I followed your instructions step by step, didn't improvise or skipped something. I'm gonna build OpenSSL again and try again.
P.S.: Why `make` on `Telegram/out/debug|release` takes sooooooooooooooooooooo long? :(

john-preston:
@EhsaanF on Debug it should not take long, about 5-10 seconds. On release it is very long because of link time optimizations. They can be disabled in .gyp files if you don't want to wait so much - just remove all "-flto" flags from tdesktop/Telegram/gyp/telegram_linux.gypi.

I didn't try that instructions step by step myself :) should test them I guess.

EhsaanF:
@john-preston Test the instructions, please :) I think it has a few mistakes.
I rebuilt OpenSSL, removed -flto flag, and again, I get the same error :cry: 
```
[...]

PR #4430, opened 2018-02-18
building-cmake.md: Added troubleshooting section

pentix:
Added the solution to the issues #2519 and #4418 that was found in respective thread discussion.


john-preston:
I don't like adding such a thing. The instructions are for Ubuntu 14.04, I've completed full setup for plain empty Ubuntu 14.04 a couple of days ago. There were some problems with breakpad that I've fixed in the instructions but everything else was fine.

PR #4498, opened 2018-03-14
Fix wrong uname flag used

svenstaro:
This is required because uname -p actually returns "unknown" for some hardware. The uname help documents this by stating that -p is non-portable. The -m flag is the one to use.


stek29:
heh, it was fixed in libtgvoip but not here :\

john-preston:
Thanks!

Issue #4481, opened 2018-03-09
Build failure in i386


3v1n0:
32 bit build currently fails with this error (well, a warning, but considered an error).

```
[ 60%] Building CXX object CMakeFiles/Telegram.dir/build/1f414157aaf40e11e829de3c9a0a1573-xenial/parts/telegram/src/Telegram/SourceFiles/history/admin_log/history_admin_log_filter.o
In file included from /build/1f414157aaf40e11e829de3c9a0a1573-xenial/parts/telegram/build/out/Release/Telegram_pch/stdafx.h:64:0:
[...]

PR #4505, opened 2018-03-16
Snap packaging support

3v1n0:
Hello,

Since some time I've been experimenting [building snap packages for the telegram desktop client](https://github.com/3v1n0/telegram-snap/) in the same way it's done upstream, so that you guys could use to easily deliver this client to users running Ubuntu and many other distros supporting  [snapd](http://snapcraft.io/) (including Linux Mint, Zorin, Elementary, Manjaro, Debian, OpenSUSE, Solus, etc.).

And thanks to these efforts [I released a package](https://forum.snapcraft.io/t/call-for-testing-telegram-desktop/3954), and Telegram Desktop is now [available in the [Snap Store](https://snapcraft.io/telegram-desktop) with tens of thousands users already using it.

In any of the distros mentioned above (in some you might need to install `snapd` package), you can easily install `tdesktop` from their app stores or from terminal:

`snap install telegram-desktop`

While latest git revision can be installed using the `edge` channel, and latest alpha/beta using the `beta` one (appending `--<channel>` to the command above is enough to get that).

However, the main purpose of such kind of packaging system is to leave upstream developers the main control over what is distributed, and then it would be nice if you guys could now take the ownership of this (while I can still continue contributing in the maintenance of the packaging bits, together with all the other snapcrafters) and manage uploads and releases.

By merging this branch, you'll be basically able to generate a snap package by running `snapcraft` command, while the maintenance of the `snapcraft.yaml` file isn't really much different to what is already done with the the travis build script.

All this can be of course automated by [using travis](https://docs.snapcraft.io/build-snaps/ci-integration#using-travis) (but kind of problematic with this project, given that a free instance has limited time), or better (as we're doing now) using https://build.snapcraft.io/ providing an easy way to just add this repository and having a snap to be built at every `git push` (that will be released in the `edge` channel by default, while it's up to developer manual approvation relasing to other channels).

I'm currently the owner of both `telegramdesktop` and `ŧelegram-desktop` names, but we can move these to your control. But you need to register on snapcraft.io to manage all this.

For any other question, I'll be happy to help.


snap/snapcraft.yaml
```diff
@@ -0,0 +1,409 @@
...
+    plugin: cmake
+    build-packages:
+      - oss4-dev
+      - portaudio19-dev
+    configflags:
+      - -DCMAKE_BUILD_TYPE=Release
+      - -DALSOFT_EXAMPLES=OFF
+      - -DALSOFT_TESTS=OFF
+      - -DALSOFT_UTILS=OFF
+      - -DLIBTYPE=STATIC
```
stek29: These flags probably should be added to instructions/travis too :)

snap/snapcraft.yaml
```diff
@@ -0,0 +1,409 @@
...
+      - -qt-harfbuzz
+      - -qt-pcre
+      - -qt-xcb
+      - -qt-xkbcommon-x11
+      - -no-opengl
+      - -no-gtkstyle
+      - -static
+      - -dbus-runtime
+      - -openssl-linked
+      - -no-mirclient
```
stek29: Same, `-no-mirclient` and `-dbus-runtime` should probably be added too

snap/snapcraft.yaml
```diff
@@ -0,0 +1,409 @@
...
+      sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 60C317803A41BA51845E371A1E9377A2BA9EF27F
+      sudo apt-get update \
+        -o Dir::Etc::sourcelist="sources.list.d/ubuntu-toolchain-r.list" \
+        -o Dir::Etc::sourceparts="-" -o APT::Get::List-Cleanup="0"
+    install: |
+      set -x
+      sudo apt install gcc-7 g++-7 -o Debug::pkgProblemResolver=yes --no-install-recommends -y
+      sudo apt-mark auto gcc-7 g++-7
+      sudo rm -f /etc/apt/sources.list.d/ubuntu-toolchain-r.list
+    prime: [-./*]
```
stek29: Is it normal for snapcrafts to do this kinds of things?

stek29:
Also I'm not sure if it's because of you changing Qt patch or not but Travis went crazy on patching Qt step

evandandrea:
@3v1n0 can you put comments in the snapcraft.yaml explaining the rationale for anything not-obvious? For example, you don't explain why you need a newer version of GCC or what that GCC part is actually doing. There's a lot of assumed knowledge in here.

snap/patches/libtgvoip-only-enable-msse2-on-supported-archs.patch
```diff
@@ -0,0 +1,19 @@
...
+index 8639848..10cd0e7 100644
+--- a/libtgvoip.gyp
++++ b/libtgvoip.gyp
+@@ -382,8 +382,12 @@
+               'defines': [
+                 'WEBRTC_POSIX',
+               ],
+-              'cflags_cc': [
+-                '-msse2',
++              'conditions': [
```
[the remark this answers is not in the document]
3v1n0: Yeah, the correct thing to do would in fact bump the submodule version, but I didn't want to manage that at this level, however I could actually drop the patch from here and wait for that to be synced.

snap/scripts/telegram-launch
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+default_downloaddir="$SNAP_USER_DATA/Telegram Desktop"
+
+if [ -d "$default_downloaddir" ] && [ ! -L "$default_downloaddir" ]; then
+  dest_downloaddir="$SNAP_USER_COMMON/$(basename "$default_downloaddir")"
+  if [ -d "$dest_downloaddir" ]; then
+    mv -v "$default_downloaddir/*" "$dest_downloaddir/"
+    rmdir "$dest_downloaddir"
```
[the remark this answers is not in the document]
3v1n0: Yes indeed... This script is acting only on updates from old versions, but indeed it is needed.

Thanks.

snap/scripts/telegram-launch
```diff
@@ -0,0 +1,16 @@
...
+  else
+    mv -v "$default_downloaddir" "$SNAP_USER_COMMON"
+  fi
+  ln -sv $dest_downloaddir "$default_downloaddir"
```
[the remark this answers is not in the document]
3v1n0: ACK

snap/snapcraft.yaml
```diff
@@ -0,0 +1,409 @@
+name: telegram-desktop
+version: git
+summary: Telegram Desktop
+description: |
+  Telegram is a popular messaging protocol with encryption and security as
+  its key focus.
+
+  Fast and secure desktop app, perfectly synced with your mobile phone.
```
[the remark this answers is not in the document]
3v1n0: Mh... I can't remember, it was long time ago :)
But indeed that can be reworded, I would be happy to do it.

I was thinking to reuse part of the text used in the play store for the mobile app (also for adding translations to the store), but not sure if it's fine.

So please, let me know what you'd like to have there.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,409 @@
...
+  version_file=Telegram/build/version
+  version=$(sed -n "s/AppVersionStr[ ]\+\(.*\)\+/\1/p" $version_file)
+  alpha=$(sed -n "s/AlphaChannel[ ]\+\(.*\)\+/\1/p" $version_file)
+  beta=$(sed -n "s/BetaVersion[ ]\+\(.*\)\+/\1/p" $version_file)
+  head_sha=$(git rev-parse HEAD)
+
+  if [ "$alpha" != "0" ]; then
+    version="$version-alpha"
+  elif [ "$beta" != "0" ]; then
+    version="$version-beta"
```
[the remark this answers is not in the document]
3v1n0: Yeah, well... Anyway, I've just supported it as an option, then it's up to you guys decide what's reaching it or not.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,409 @@
...
+
+parts:
+  telegram:
+    plugin: gyp-cmake
+    source: .
+    source-type: git
+    prepare: |
+      set -x
+      patches_dir=$SNAPCRAFT_STAGE/../snap/patches
+      git -C ../src/Telegram/ThirdParty/libtgvoip apply -v3 $patches_dir/libtgvoip-only-enable-msse2-on-supported-archs.patch || true
```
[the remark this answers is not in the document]
3v1n0: Yeah, it was my idea too... Just wanted to wait your input.

armhf/i386 don't build anyway these days beause of some warnings, so could be the case to get off the `-werror` for those builds, but that's indeed a more generic issue too. (see #4481).

snap/snapcraft.yaml
3v1n0: Well, not really "normal", but there are not really many options to build this with gcc-7, so until there will be a snap core based on 18.04, we need such workarounds to compile.

This is although quite safe, since it won't really affect the system (even when not using a `cleanbuild`) since an `apt autoremove` will get you back to the previous state (same for other packages that snapcraft will install anyays).

3v1n0:
@stek29 
> Also I'm not sure if it's because of you changing Qt patch or not but Travis went crazy on patching Qt step

Well, that happens because the cached version is installed with `sudo make install` (while sudo when installing in local prefeix is not needed, and actually wrong), so we need to `sudo rm` too.

I guess all these should be changed (in a different branch), to never use sudo to install in local prefixes.

evandandrea:
Thanks for d2536c5, @3v1n0. That addresses my concerns perfectly.

.travis/build.sh
```diff
@@ -585,7 +585,7 @@ buildCustomQt() {
     rm -rf "$EXTERNAL/qt${QT_VERSION}"
   fi
   cd $QT_PATH
-  rm -rf *
+  sudo rm -rf *
```
stek29: what could go wrong

3v1n0: Nothing as per the statement above if the var is defined, but that was just a temporary change... What should be done is removing the `sudo` from here and from the install stanza.

But since I've not the powers to invalidate the cache, this should be probably something you guys can do in a different branch?

snap/snapcraft.yaml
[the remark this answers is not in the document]
stek29: Won't text from .desktop Comment fit? (Official desktop version of Telegram messaging app)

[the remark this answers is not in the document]
stek29: Won't text from .desktop Comment fit? (Official desktop version of Telegram messaging app)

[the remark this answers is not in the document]
3v1n0: Yes, sure... But we also need a more complete description for people who don't know what it is.

So, that is fine as `summary` but for description we can add a multi-line description that can be much more complete. So it would be nice if you could point me to something you'd like for that.

auchri:
I've cleaned the cache and restarted the travis build.

3v1n0:
Mh, I still see the failures...

Want me to do a PR to remove the sudo to the `make install`'s and then we can recreate a cache from there? Other option is to use chown on there, but again... Not the nicest way.

Telegram/Patches/qtbase_5_6_2.diff
```diff
@@ -239,6 +239,43 @@ index 41834b21ae..8cdf4ab145 100644
...
++    if (xdgCache.isEmpty()) {
++        xdgCache = QDir::cleanPath(QDir::homePath() + QStringLiteral("/.cache"));
++    }
++    return xdgCache;
++}
++
+ static const QString KDEItemFormat = QStringLiteral("org.kde.StatusNotifierItem-%1-%2");
+ static const QString KDEWatcherService = QStringLiteral("org.kde.StatusNotifierWatcher");
+-static const QString TempFileTemplate =  QDir::tempPath() + QStringLiteral("/qt-trayicon-XXXXXX.png");
++static const QString TempFileTemplate = cachePath() + QStringLiteral("/qt-trayicon-XXXXXX.png");
```
john-preston: Why is it better to place them in the cachePath instead of the tempPath? Will it definitely not break something for someone?

3v1n0: Snap `$TMPDIR` folder is defined by default as somewhere in `/tmp/snap.telegram-desktop-$random-numbers`. And that folder isn't accessible by system.

So the icon that the snap will place there won't be readable by the desktop shell. Also Qt accepted this solution upstream (see https://codereview.qt-project.org/#/c/182492/), so I've just replicated it here to make things work (and better to override the `TMPDIR` env at snap level)

omkarnathsingh:
This should be made official. I tried snap 1.2.17 version. Working great :)

stek29:
@3v1n0 do you mind adding "unofficial snap [for official desktop client ...]" to unofficial snap builds available in snap already? (If you're able to)

Telegram/Patches/qtbase_5_6_2.diff
omkarnathsingh: John, snap apps runs in a sandbox kind of environment, so its all good. I am waiting for this to become officially accepted. Though being snapcraft.yaml file open source, I am having trust on build systems and using it :D

3v1n0:
Hey, I finally had some time to merge this with current dev branch and fix some issues.

Everything should be fine on snap side, while in order to build this I had to cherry-pick the fix I proposed in #4935.

Let me know if you need anything else to merge this.

3v1n0:
Ah, one question about this... I've noticed that now your travis instance supports longer build times.

So, if you care I can also create a travis build for building the snap.
However, I guess this is something that would not that needed, as http://build.snapcraft.io/ would to the same for commits. But if you care about PRs as well, let me know and I'll get that in place.

Telegram/Patches/qtbase_5_6_2.diff
3v1n0: Ineed it won't break anything.... Also as said this is something [that was done in a similar way in the upstream Qt too](https://code.qt.io/cgit/qt/qtbase.git/tree/src/platformsupport/themes/genericunix/dbustray/qdbustrayicon.cpp#n68), so quite safe to merge I think

john-preston:
It offers "Squash and merge" here, with merge commits being disabled. Is it OK to have single snap packaging commit in this repository or should I keep the 117 commits history with a merge? @3v1n0 That's a lot of commits =)

PR #4513, opened 2018-03-18
add "Supported systems" icons

yiyu0x:
add some icons in "Supported systems" list


auchri:
Imho, this icons should be added to the repo insted if loading them from somewhere (under docs/assets).

Also, where did you get this icons from and what's the license of them?

yiyu0x:
i got those icon form [here](https://www.flaticon.com/) , this is [license](https://file000.flaticon.com/downloads/license/license.pdf) . i need to create a new pull/request? (im a beginner for github :) )

auchri:
Well, the license says 

> By placing a link with the text "designed by {Author's Name} from Flaticon" in a visible
spot, so the author's authorship is noticeable

which means that it is not enough to add the license to the list of licenses and therefore I think it's not necessary to add the icons and the extra license.

PR #4552, opened 2018-03-30
Fix #4544

### LINKED ISSUE #4544: Restore after maximize not preserve conversations list and message list width
### Steps to reproduce
1. Make program width as thin as possible, while conversations list still visible.
2. Maximize window.
3. Restore window.  

It happens and with non minimum width, but not in 100% of cases. With minimum width it's 100% reproducible.

### Expected behaviour
After restoring window must be the same as before maximizing.

### Actual behaviour
tg_r

### Configuration
**Operating system:** windows 7 x64

**Version of Telegram Desktop:** 1.2.15

**Used theme**: default

kirsan31:
Attempt to fix #4544. Tested on several windows machines. I am sorry, if I done something wrong. But any way, I think that my PR will help @john-preston to proper fix this...

### DIFF
--- a/Telegram/SourceFiles/platform/win/main_window_win.cpp
+++ b/Telegram/SourceFiles/platform/win/main_window_win.cpp
@@ -873,7 +873,9 @@ void MainWindow::updateSystemMenu(Qt::WindowState state) {
 }
 
 void MainWindow::psUpdateMargins() {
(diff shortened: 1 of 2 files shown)


kirsan31:
Dunno why one of Travis builds fails (timeout?), cos I modified only windows related files...

PR #4553, opened 2018-03-30
Add option for message formatting (-- to ‏‎−, >> to », << to «) in "Settings" window

### LINKED ISSUE #522: Option for message formatting (-- to ‏‎−, >> to », << to «)
Hi.
I want to report a bug in the telegram desktop version. (Linux)
The problem of the character: Dash. Or (-)
When two characters Dash, taken together, the result rather than the (--) becomes a long line. (‏‎−)
But when the two characters together Dash, the text first, this problem does not exist.
For example:
http://s4.picofile.com/file/8173777318/snapshot.png

PeterMX:
this closes #522  because add an option in settings to disable message formatting, it's enable by default so it's not a problem for people than use it.

### DIFF
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -296,6 +296,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 
 "lng_settings_section_chat_settings" = "Chat Settings";
 "lng_settings_replace_emojis" = "Replace emoji";
+"lng_settings_message_formatting" = "Message formatting";
 "lng_settings_suggest_by_emoji" = "Suggest popular stickers by emoji";
 "lng_settings_view_emojis" = "View list";
 "lng_settings_send_enter" = "Send by Enter";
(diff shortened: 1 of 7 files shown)


Telegram/Resources/langs/lang.strings
```diff
@@ -296,6 +296,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 
 "lng_settings_section_chat_settings" = "Chat Settings";
 "lng_settings_replace_emojis" = "Replace emoji";
+"lng_settings_message_formatting" = "Message formatting";
```
stek29: "message formatting" isn't clear enough imo

kirsan31: This _strange_ replacements are made by telegram, not clear enough too :)
I really can't understand for what\whom it's done, and therefore have strong feeling that this PR will be rejected...

PeterMX: someone knows how should it be called?

JuanPotato: replace double chars (<<, >>, --) ?

kirsan31: Or   
```
Double chars (<<, >>, --) replacement.
Double chars (<<, >>, --) to special chars.
Convert (<<, >>, --) to special chars.
```
But, imho, this _feature_ does not deserve an option in settings, and simple must be removed.

PeterMX: Should I wait for change request?

PeterMX: @stek29 It's clear enough now.

stek29:
See 4b763a7

PR #4633, opened 2018-04-27
fix libressl-2.7 compatibility

Johnnynator:


stek29:
Was going to ask whoever added initial libressl flag to check it and realized it was you :)

PR #4721, opened 2018-05-23
Update building-msvc.md

laghaee:
/property:Platform="x86" flag required in case a system is not x86 to prevent build error when compiling LzmaLib.sln


stek29:
?..

PR #4782, opened 2018-06-03
Feature/issue 948 add option start in system tray

### LINKED ISSUE #948: Add option "Start in system tray" in "Settings" window

Musikolo:
Implementation of feature request closes #948

This PR is a continuation of [PR4268 ](https://github.com/telegramdesktop/tdesktop/pull/4268).

It includes the changes as suggested by @john-preston . There should be no interference in current workflow.

Please, let me if anything is needed from my side.

Thank you!


Telegram/SourceFiles/storage/localstorage.cpp
```diff
@@ -518,7 +518,7 @@ enum {
 	dbiLastUpdateCheck = 0x0d,
 	dbiWindowPosition = 0x0e,
 	dbiConnectionTypeOld = 0x0f,
-	// 0x10 reserved
```
stek29: this doesn't seem right

Aokromes:
@ Musikolo  if this replaces https://github.com/telegramdesktop/tdesktop/pull/4268 close it :)

Musikolo:
@Aokromes done! ;-)

Telegram/SourceFiles/storage/localstorage.cpp
john-preston: It was used in some private fork of tdesktop for one of settings. It is not a problem right now but easily could be, if this fork was still maintained.

john-preston:
Unfortunately this still can't be merged.

Style problem: Two setting variables, StartInTray and StartInSystemTray with unrelated meaning is bad.

Logic problem: On Windows you get two settings checkboxes, "Launch in system tray" and "Launch minimized", if first one is checked, the second one doesn't have any effect, you'll launch minimized anyway. It should not be shown in case of checked "Launch in system tray" at all, because you always launch in system tray, no matter was you launched from autostart or not.

UX problem: In case of non-checked "Show tray icon" checkbox you just disable the checkbox "Launch in system tray" which is bad, because it looks the same and just doesn't give any reactions when you click it, which looks buggy. It should hide animated in case you uncheck "Show tray icon" and show animated in case you check "Show tray icon", like "Launch minimized" does when you toggle "Launch Telegram when system starts" (or like version string when you toggle "Update automatically" or "Notifications position and count" link when you toggle "Use Windows notifications").

Regression problem: In macOS if you launch application with this checkbox enabled (try to start app in tray with gStartInTray == true) it behaves awful: first launches the app, then minimizes with animation a fully grey window to the Dock, then unminimized with animation this fully grey window from the Dock and shows the window content. This is crazy.

(not related to @Musikolo or anyone specific, not first time and please no offence):

This is the reason I generally don't accept pull requests :( It is easy to code something up and suggest it to the main repository and leave the maintainer with the job to test that everything works in all possible cases and works the way it should - but I spend a lot (really) of time first searching for problems and then explaining them with no benefit to the project - it doesn't help to work on the project, but takes time and energy.

This is the reason why [Contributing Policy](https://github.com/telegramdesktop/tdesktop/blob/dev/.github/CONTRIBUTING.md) clearly states:

> We highly appreciate your contributions in the matter of fixing bugs and optimizing the Telegram Desktop source code and its documentation. In case of fixing the existing user experience please push to your fork and submit a pull request.
> 
> Unfortunately we do not merge any pull requests that have new feature implementations, translations to new languages and those which introduce any new user interface elements.

Musikolo:
@john-preston, honestly, I only use Linux, and I don't have any other OS. Besides, I completely understand the effort it takes for you testing in the other environments. So, would it an option implementing my feature only for Linux environments? Before I pushed my code, I could test it in Linux, but since I don't have the other OSes, I couldn't do it in any of the other supported OSes. I don't want to add any burden on you, and that's why I'm suggesting this approach. Thoughts?

mrmeszaros:
@Musikolo It seems to me, that this Linux-only setting / feature would increase the complexity of maintaining the code-base for @john-preston.
There are a few Windows only settings, though, like `Show Taskbar Icon`, so similar Linux only could also make sense.
Maybe an option to start minimized could be a middle ground.
With another option to minimize to tray (Linux only)?

(A Linux _main_ here that would appreciate this or a similar feature.)

PR #4827, opened 2018-06-09
Added an option to set UI scaling to 175%

0xADF:
Added an option to use 175% GUI scaling. Might be useful for those who think everything is too small at 150% and too large at 200%.

image

### DIFF
--- a/Telegram/SourceFiles/application.cpp
+++ b/Telegram/SourceFiles/application.cpp
@@ -418,8 +418,10 @@ void launch() {
 		cSetScreenScale(dbisOne);
 	} else if (dpi <= 132) { // 108-120-132
 		cSetScreenScale(dbisOneAndQuarter);
-	} else if (dpi <= 168) { // 132-144-168
+	} else if (dpi <= 150) { // 132-144
 		cSetScreenScale(dbisOneAndHalf);
+	} else if (dpi <= 168) { // 160-168
+		cSetScreenScale(dbisOneSevenFive);
 	} else { // 168-192-inf
 		cSetScreenScale(dbisTwo);
 	}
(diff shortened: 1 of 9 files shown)


john-preston:
I don't really understand how the graphics and emoji will work in that case. It looks (by the screenshot of the checkboxes) they'll be just smaller than they should be. Currently there is no graphics embedded for 1.75 scale factor and no emoji sprites for it either.

0xADF:
I haven't found this part in the code, but from what I see, they scale as if the factor was 2.0.

150%:


175%:


200%:

john-preston:
It uses 200% emoji (which isn't that bad, except they're larger than text and they'll look strange inside the text), but if you look on other places in your screenshot (that uses graphics from 150%) you'll see bad rendering (look at the choose-emoji button or at the "go to original message" arrow inside the chat) - they're smaller than they should be and they're not centered in places where they should be and there could be even worse examples throughout the app I'm afraid.

0xADF:
Do you think I would be able to address this in the code without having to produce new graphic elements?

john-preston:
@0xADF It is possible for graphics (125% / 150% are created from 200% while preprocessing the graphics before build), but it won't be easy to correctly support all this, it was not supposed to work that way. I have a task of supporting scales up to 400% (but this will require graphics being re-drawn for the largest case) and automatic scaling when being run (not when being built), but this is a large project in some far future :(

Emoji sprites are generated for all sizes right now, so it is not possible to easily support other scale factors in them. The large project in some far future includes one (largest) emoji sprite and down-scaling them when being run, but I don't know when I'll be able to do that.

dsk9:
add more scaling for 4k display . 200% so small in 4k device

PR #4869, opened 2018-06-19
Add caption to first item of album istead of sending separated message.

23rd:
Since appeared the albums in Telegram, for some reasons desktop client sends caption before album.
This pull request adds caption to first item of album to see caption under whole album.

What I do:
image

What I get now:
image

What I get after my patch:
image

### DIFF
--- a/Telegram/SourceFiles/apiwrap.cpp
+++ b/Telegram/SourceFiles/apiwrap.cpp
@@ -4309,7 +4309,7 @@ void ApiWrap::sendFiles(
 		return list.files.front().mime == qstr("image/webp");
 	};
 	if ((list.files.size() > 1 || isSticker())
-		&& !caption.text.isEmpty()) {
+		&& !caption.text.isEmpty() && !album) {
 		auto message = MessageToSend(options.history);
 		message.textWithTags = std::move(caption);
 		message.replyTo = options.replyTo;
@@ -4344,6 +4344,7 @@ void ApiWrap::sendFiles(
 			to,
 			caption,
 			album));
+		caption = TextWithTags();
 	}
 	if (album) {
 		_sendingAlbums.emplace(album->groupId, album);


stek29:
this is probably because initially clients wouldn't render caption of first element as caption of whole album, but that's no longer the case, and this behavior would be consistent with other apps.

Telegram/SourceFiles/apiwrap.cpp
```diff
@@ -3834,7 +3834,9 @@ void ApiWrap::sendFiles(
 		TextWithTags &&caption,
 		std::shared_ptr<SendingAlbum> album,
 		const SendOptions &options) {
-	if (list.files.size() > 1 && !caption.text.isEmpty()) {
+	bool firstFile = true;
```
stek29: I'd rather place it before the for loop (on line 3854), and use `auto` instead of `bool`.

stek29:
> c1xx: note: the system returned code 1455: The paging file is too small for this operation to complete

meh. And I'm unable to restart AppVeyor builds ¯\\\_(シ)\_/¯

Telegram/SourceFiles/apiwrap.cpp
23rd: Can't understand why we need to use `auto` for primitive data type.

stek29:
Please Rebase on top of latest dev branch so travis build passes

Telegram/SourceFiles/apiwrap.cpp
```diff
@@ -3897,6 +3901,7 @@ void ApiWrap::sendFiles(
 			to,
 			caption,
 			album));
+		firstFile = false;
```
john-preston: It would look nicer without `firstFile` at all and just assigning `caption = TextWithTags();` here, after a task was pushed. It is a bit less optimal, but the code is easier.

23rd: You are strongly right.

23rd:
Waiting for review.

23rd:
@john-preston Still hope. =(

23rd:
Replaced 3 tiny commits with single commit and rebased on top of latest dev branch.

23rd:
@john-preston To be honest, I don't understand why this PR has been ignored for the past six months.
Perhaps you have some questions or some doubts.
I have been using binary with this patch for six months and I have not had any errors or bugs.

john-preston:
@23rd The problem is I'm not sure that the proposed way is better :(

This applies to several of your requests — you propose not some fix of a bug or smth like that, but you take some intendedly coded behaviour and change to some other behaviour. Perhaps it is better, perhaps not. Most likely it is better for someone or in some cases and not better in others or for others.

In case of album captions — they indeed work that way on the Telegram as a whole (that the caption to the first album item is displayed as album caption). I don't really like that approach, in my opinion it is really confusing (for example when viewing album in the media viewer), so I did not promote this feature, leaving perfectly clear (single) way of adding a comment to a bunch of photos / files. But maybe it is better to do the sending your way and no one really cares how they look in the media viewer. I just still didn't decide.

delivrance:
@john-preston How about sending the caption as a reply to the last media *after* the whole album has been sent?

john-preston:
@delivrance This is harder to implement and it will be inconsistent with all other places where a comment is added to something (comment to several files being sent, comment to forwarded messages, comment to a shared post from channel or shared media from bot — they all send comment before the object, in all Telegram apps.

23rd:
@john-preston 

> I'm a bit concerned that if you send an album this way and while it is uploading you cancel the sending of the first photo you'll loose the caption as well. 

But this is quite normal behavior. In Android and iPhone applications, if the user sends an album and one image has a caption, then after canceling the sending of exactly this image, caption will be lost.

Perhaps the tdesktop should just display a message that canceling the first image will cause the caption to be lost. But for this you need to add a new language string.


Alternatively, sending of text as a caption can be provided through an additional keyboard shortcut (for example, `Ctrl + Enter`), only for advanced users, like as you added `Ctrl + Up/Down arrow` shortcuts for replies and `Alt + Shift + Click` shortcut for language pack. But I'm not sure this is good idea.

john-preston:
I've discussed this (at last, sorry :) and this way was considered a better one. Thanks!

I see a best way to handle caption on cancel to silently move the caption to the new "first" item of the album (this way you cancel first item, but still have the album caption as before). But this will require some really hacky things and should be done carefully.

23rd:
Oh wow.
Seems like this is a new year's gift to me. 😁

Lonami:
> macOS client started doing the opposite now

Think differently.

Lonami:
> macOS client started doing the opposite now

Think different.

PR #4872, opened 2018-06-20
Fix viewing U+0640 unicode character

procodr:
This fixes a bug in Telegram Desktop comming from Qt.
Typing U+0640 unicode character (shift + j with Persian keyboard layout) causes this bug:

that views the words incurrectly.
It comes from Qt 5.6 that is fixed and applied but is disabled with this comment:
> Disabled due to regressions. The font selection algorithm is not prepared for this change.

more info available [here](https://github.com/qt/qtbase/commit/0ec07b68ad34e135451dd5291732bf73d297ba0c) and [here](https://bugreports.qt.io/browse/QTBUG-28813)

But I have enabled this fix and it works.
In my test there was no problem with about 20 hours usage (maybe because of that tdesktop is not using font selection).
And here is the result:


john-preston:
I don't know how to test that. If they didn't commit that to their LTS version commenting that there is a regression I'm not sure this is a good idea to just go for it ignoring the comment. It could break / crash something in some circumstances that you (and I) didn't encounter.

john-preston:
I hope it is fixed in Qt 5.12.5 that is used now for the stable builds.

PR #4908, opened 2018-06-27
Remove unnecessary memory allocations in localstorage

MihaZupan:


john-preston:
@MihaZupan Are you sure it is unnecessary? When you write a QByteArray into a stream you don't write only the content itself (audio.size()), but also some meta information, like the size of the buffer and maybe something else.

MihaZupan:
My bad, only 1 quint32 is not needed. QDataStream prepends a quint32 for length with byte arrays and strings. In writeWebFile, one such prefix is already accounted for when calling stringsize.

john-preston:
@MihaZupan Then perhaps it should just use "Serialize::bytearraySize" there instead of this manual counting.

MihaZupan:
Updated it to use "Serialize::bytearraySize" and found a couple more instances of an extra quint32

PR #4935, opened 2018-07-02
lib_export: use includes paths as defined per platform

3v1n0:
Fixes the build failures locally (and hopefully) in travis

### DIFF
--- a/Telegram/gyp/lib_export.gyp
+++ b/Telegram/gyp/lib_export.gyp
@@ -14,10 +14,14 @@
     'includes': [
       'common.gypi',
       'qt.gypi',
+      'telegram_win.gypi',
+      'telegram_mac.gypi',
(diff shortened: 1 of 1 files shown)


stek29:
Thank you!

vhborges:
After this commit I cannot build Telegram anymore, when running gyp/refresh.sh I get this error:

> gyp: Undefined variable build_defines in lib_export.gyp

Is there something else to be updated?

3v1n0:
@vhborges try to pass a -Dbuild_defines='' to the gyp commandor that could be defined empty in the `lib_export.gyp` file too as other variables

john-preston:
This broke default Linux build and broke macOS build completely without a fast fix, because "mac_bundle" properties can't be set in "shared_library" targets :(

PR #4958, opened 2018-07-10
A new language file for Japanese people.

petitdream:


stek29:
thanks, but languages aren't managed here, see https://translations.telegram.org.

stek29:
For now you can distribute your custom lang file, and let other people choose it with [`loadlang`](https://github.com/telegramdesktop/tdesktop/wiki/Hidden-modes)

PR #5046, opened 2018-08-02
Show Emoji replacement tooltip on hover

magnus-gross:
@john-preston 
Hi, this pull request adds a tooltip to the emojis, that shows the replacement text to type in when using the keyboard. This feature was requested by multiple users.
Of course this is entirely disabled, if the user disables "Suggest Emoji Replacements".

I use the `GetAllReplacements()` list to find the replacement text, if you know a better or faster way to get it, I am happy to incorporate that. :)

I hope I followed all your guidelines, if not feel free to point out any changes.

alt text

### DIFF
--- a/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
+++ b/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
@@ -11,6 +11,9 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 #include "styles/style_chat_helpers.h"
 #include "ui/widgets/shadow.h"
 #include "lang/lang_keys.h"
+#include "emoji_suggestions_data.h"
+#include "emoji_suggestions_helper.h"
+#include "facades.h"
 
 namespace ChatHelpers {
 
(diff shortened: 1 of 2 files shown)


Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
```diff
@@ -667,6 +670,28 @@ Ui::Emoji::Section EmojiListWidget::currentSection(int yOffset) const {
 	return static_cast<Section>(sectionInfoByOffset(yOffset).section);
 }
 
+QString EmojiListWidget::tooltipText() const {
+	const auto &replacements = Ui::Emoji::internal::GetAllReplacements();
+	auto section = (_selected / MatrixRowShift);
```
john-preston: Please use `const auto` in all places that don't modify the value afterwards (if I didn't miss anything it will be all `auto` in this method).

Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
```diff
@@ -667,6 +670,28 @@ Ui::Emoji::Section EmojiListWidget::currentSection(int yOffset) const {
...
+	auto sel = _selected % MatrixRowShift;
+	if (_selected >= 0 && section < kEmojiSectionCount && sel < _emoji[section].size()) {
+		auto emoji = _emoji[section][sel];
+		auto text = emoji->text();
+		// find the replacement belonging to the emoji
+		auto it = std::find_if(replacements.begin(), replacements.end(), [&text](auto &one) {
```
john-preston: Please use `ranges::find_if(replacements, [&](const auto &one) {` for the search.

magnus-gross:
@john-preston I made your requested changes and rebased on top of the latest stable version. :)


(I would have rebased on latest dev, but the Arch build scripts that I use and that allow me to use latest Qt do not work on the latest commit, however this should still merge without a conflict)

john-preston:
@magnus-gross Thanks!

Issue #2603, opened 2016-11-06
Telegram Desktop Linux doesn't work with nimf


yangbeom:
I used UIM-byeoru, Nimf, fcitx, XIM, Ibus. 
but only can type korean to using fcitx.
I wanna using [Nimf](https://github.com/cogniti/nimf). 
so.. Do you have plan to support other input method??

### Configuration
**Operating system:**
Ubuntu 16.04.1 LTS
**Version of Telegram Desktop:**
0.10.19

stek29:
Does Nimf work in other Qt apps?

yangbeom:
may be...
because [Nimf issue #16](https://github.com/cogniti/nimf/issues/16) developers say you need patch like support fcitx.
so i tried it but fail :(
[Nimf developer's patch](https://github.com/cogniti/nimf/files/390379/qtbase_5_6_0-with-nimf.txt)

stek29:
So TDesktop would need one more patch to Qt which must be rewritten because TDesktop already patches Qt and patches are not compatible, and all that just to support a new input method no one uses?

yangbeom:
i tried test Nimf, But Build fail Nimf.
Nimf developers fix it in jan/2017
i will reply comment when i build Nimf and finish some test

mxa:
Is this fixed? I still can't use Nimf in Telegram Desktop.

yangbeom:
maybe not.... I forgot it.... didn't test yet.

yangbeom:
I tried to write korean using Nimf(2017.04.08) in Telegram Desktop, But Doesn't work.

stek29:
Then why won't you reopen the issue?

mxa:
@yangbeom should reopen this issue.

yangbeom:
@mxa Sure :)

mxa:
:) maybe change the title in "Telegram Desktop Linux doesn't work with nimf"

mxa:
Maybe @auchri can label this one bug, help-wanted, linux too?

auchri:
@mxa maybe 😉

mxa:
@stek29 "Does Nimf work in other Qt apps?" I tried Qtractor, and yes, I can name tracks there using Nimf.

devunt:
Nimf works with Kate, so it works in other Qt apps.

log from `G_MESSAGES_DEBUG=nimf kate`:
```
[...]

PR #5050, opened 2018-08-03
Add nimf inputcontext plugin for linux version

5HARK:
Add nimf support(#2603)

nimf Qt IM Plugin Repo(for this PR): https://github.com/5HARK/im-nimf-qt5
Nimf Project Repo: https://gitlab.com/nimf-i18n/nimf


john-preston:
I'm not sure I can merge that, because it adds a lot of new runtime dependencies. New shared library dependencies include:
```
libgio-2.0.so.0
libgobject-2.0.so.0
libgmodule-2.0.so.0
libz.so.1
libselinux.so.1
libresolv.so.2
libffi.so.6
```
That means that anyone without any of that libraries who will autoupdate Telegram Desktop will just get it not launching at all while working fine before the update.

john-preston:
After trying to get rid of those new dependencies I understood that the problem lies deeper.

I load either gtk3 or gtk2 depending on the appindicator library I was able to load. Then I resolve all the used gtk methods from the loaded shared library dynamically (like g_free, g_[signal_smth] etc). I'm afraid that this nimf plugin that is linked to gtk2 specifically will work unreliably if I load and use gtk3 myself and then use gtk2 in this plugin. Also it just wont work if gtk is absent.

To do that the right way I need a way to suggest my dynamically resolved gtk functions to the plugin and for plugin to work with them or just return false in check() if the functions were not loaded. This can be hacked inside the nimf plugin code, but I didn't yet figure out how to do that in a nice way. Some header sharing is required between nimf plugin and tdesktop code.

5HARK:
The same glib will be loaded if gtk2 and gtk3 are installed on your system. It does not affect nimf plugin.

If you have only gtk2 installed on your system, you can not install Nimf. isValid() returns false.

Therefore, I think there is no problem. Is there anything else to consider?

john-preston:
@5HARK Currently it brings new static dependencies. It should load the methods dynamically, otherwise I'm not sure that the app will continue working on the same systems it works right now.

5HARK:
@5HARK okay. i got it.
I will change im_nimf_qt5.cpp code for delete new static dependencies(glib, gio etc.)
and after that I will rewrite pull request again.

5HARK:
I updated nimf Qt IM Plugin Repo(for this PR): https://github.com/5HARK/im-nimf-qt5
Please review this PR again.

john-preston:
Ok, I'll try to build it with the new plugin code.

john-preston:
It builds fine and doesn't add new dependencies. I hope there won't be any weird runtime problems after such update.

stek29:
> It builds fine

https://travis-ci.org/telegramdesktop/tdesktop/jobs/464207607#L4244 :(

Issue #3366, opened 2017-05-06
Additional shortcut key options for sending messages 


Canis-Enigmas:
### Enhancement Suggestion
Shortcut keys for sending messages currently support either 'Enter' or 'Ctrl+Enter' for PCs and 'Command + Enter' for Macs. 
Due to the differences between keyboard layouts of PCs and Macs, would it be possible to add 'Alt+Enter' as an additional shortcut key for PCs.
This is a big convenience for those of us who switch between both devices often as the location of 'Alt+Enter' on PC keyboards is the same with 'Command+Enter' on Mac keyboards. 
Thank you.

Aokromes:
maybe it can be changed to a multiple selection like some programs allows.
IE
Enter.
Alt+Enter
Control+Enter
Control+Alt+Enter.

stek29:
Or maybe it could be changed to any with custom shortcut system

Canis-Enigmas:
Personally, I actually prefer 'Shift+Enter' from the old Trillian days, but I guess that might be asking a bit too much? Updated the title to better reflect suggestions here.

MasterGroosha:
There is `shortcuts-custom.json` file in `tdata/tdata` folder inside TDesktop's current folder, so in theory your issue might be solved in future if `send_message` action will be added to custom shortcuts list.

tiktuk:
Yes, alt+enter is standard on Mac I think, would be great not having to use a different shortcut in Telegram only.

Aokromes:
Yeah stek29, something like a lot of games where you can select any shortcut https://i.imgur.com/gqsl81f.png

Aokromes:
how about to rename this to "Custom shortcuts system" and put here
https://github.com/telegramdesktop/tdesktop/issues/4258
https://github.com/telegramdesktop/tdesktop/issues/2012

Aokromes:
maybe 

https://doc.qt.io/archives/qq/qq14-actioneditor.html
https://github.com/bjorn/tiled/commit/c1a48ad13b812d6531b372ec8295b19a769257d9

can help someone to create one pull request implementing custom user shortcuts allowing everyone to setup every shortcut they want.

Aokromes:
still wanted

ohhai:
Configurable shift+enter for sending a message would be nice
Also: #2671

23rd:
465a42a825bfc2be51bb7c77be5d031229e15b10
Why this issue is still open?

PR #5072, opened 2018-08-10
Added shortcut to open dialog for attachements.

23rd:
Small pull request adds shortcut `Ctrl + O` to open dialog for attachements.
I think it's a good idea to add ability for users to attach files via the keyboard. `Ctrl + O` is a well-known shortcut that is used by all programs to open files.


Aokromes:
maybe a good idea can be a future pull request to allow people to create own shortcuts and close https://github.com/telegramdesktop/tdesktop/issues/3366

23rd:
As I know, Telegram supports `shortcuts-default.json` for a long time.

Aokromes:
with a gui? :)

PR #5129, opened 2018-09-01
Switch to Ayatana Indicators

mymedia2:
Ayatana AppIndicator is a fork of Canonical's AppIndicator maintained by Debian community.
More info: https://lists.debian.org/debian-devel/2018/03/msg00506.html

### DIFF
--- a/Telegram/SourceFiles/platform/linux/linux_libs.cpp
+++ b/Telegram/SourceFiles/platform/linux/linux_libs.cpp
@@ -234,14 +234,14 @@ void start() {
 	bool gtkLoaded = false;
 	bool indicatorLoaded = false;
 	QLibrary lib_gtk, lib_indicator;
-	if (loadLibrary(lib_indicator, "appindicator3", 1)) {
+	if (loadLibrary(lib_indicator, "ayatana-appindicator3", 1) || loadLibrary(lib_indicator, "appindicator3", 1)) {
 		if (loadLibrary(lib_gtk, "gtk-3", 0)) {
 			gtkLoaded = setupGtkBase(lib_gtk);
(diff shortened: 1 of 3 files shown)


john-preston:
@mymedia2 What happens if I build the version with libappindicator headers and on the target system ayatana library will be loaded?

Are they fully binary compatible, so there is no difference at all which headers I use?

I'm not comfortable with the fact that currently my versions are built using one set of headers, but if I (for testing purposes or whatever) install ayatana packages on that system, the project silently starts using different set of headers. This can be fine only if those two different header sets are giving fully binary compatible interfaces.

mymedia2:
> What happens if I build the version with libappindicator headers and on the target system ayatana library will be loaded?

In this case Telegram Desktop will work. I've tested in Debian with Xfce desktop environment.

> Are they fully binary compatible, so there is no difference at all which headers I use?

Yes, they are compatible. One of the reasons for renaming Ayatana fork is to allow co-installing these two libraries.

> I'm not comfortable with the fact that currently my versions are built using one set of headers, but if I (for testing purposes or whatever) install ayatana packages on that system, the project silently starts using different set of headers. This can be fine only if those two different header sets are giving fully binary compatible interfaces.

They almost identical and provide a compatible interface. But I could add a special flag to the gyp script for switching the libraries.

john-preston:
@mymedia2 This will help only if you'll decide which runtime library to use based on the same flag that you use to decide which headers to include.

But if you say that they're compatible I think it should work as it is. Thanks!

I don't quite understand why Ayatana indicators should be preferred over the original Canonical library (in case both are installed), but let it be that way for now.

PR #5182, opened 2018-09-23
Removed unused variables in input_fields.cpp.

23rd:
These variables, according to commit history, were added about two years ago and have never been used, so they can be safely removed.


john-preston:
Thanks!

PR #5208, opened 2018-09-30
Add lots of missing includes

svenstaro:
Frankly, I'm not sure how this ever compiled for you with all these missing includes. Probably thanks to include order? At any rate, you should try to make sure that your travis build will actually fail if you don't add all the includes.

### DIFF
--- a/Telegram/SourceFiles/app.h
+++ b/Telegram/SourceFiles/app.h
@@ -7,6 +7,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 */
 #pragma once
 
+#include "base/assertion.h"
 #include "data/data_types.h"
 #include "data/data_peer.h"
 
--- a/Telegram/SourceFiles/base/algorithm.h
+++ b/Telegram/SourceFiles/base/algorithm.h
@@ -7,6 +7,10 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 */
 #pragma once
 
+#include <utility>
+#include <cstddef>
+#include <QLatin1String>
+
 namespace base {
 
 template <typename Type>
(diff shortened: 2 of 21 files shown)


Johnnynator:
https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/SourceFiles/base/base_pch.h#L12

That is the QMutex include that gets used.

svenstaro:
In that case, I think you should also support building without PCH as that is just an optimization.

svenstaro:
In fact, your patch [here](https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/Patches/gyp.diff#L11) seems to imply that building using PCH is an option. If it's an option, the non-PCH way should also be supported.

Johnnynator:
> In fact, your patch here seems to imply that building using PCH is an option. If it's an option, the non-PCH way should also be supported.

This patch only touches gyp itself and adds support for pch. It is the equivalent to this file you use in arch https://git.archlinux.org/svntogit/community.git/tree/trunk/CMakeLists.inj?h=packages/telegram-desktop

But I agree having a code base here that doesn't require pch that strictly would be nice.

john-preston:
@svenstaro Build requires including pch implicitly in all three lib_ libraries and in the main project itself. Building without it is not supported, would require a crazy amount of includes being added in the main project files and also would be unbelievably slow because of extremely bad modularization of the whole project.

svenstaro:
Alright, so you won't accept this PR? In that case, I suggest you make it clear in the docs that only PCH building is supported.

john-preston:
@svenstaro Did this pr allow you to build full Telegram or just libraries? Or you used pch for the Telegram itself?

svenstaro:
I used PCH for some parts. This allowed storage to build, at least.

john-preston:
@svenstaro Yeah, I don't think this partial fix of non-pch build is needed then.

john-preston:
@svenstaro The build instructions for all platforms use it, so following them should produce a successful build.

PR #5222, opened 2018-10-03
Remove duplicated packages in brew install

tiendq:
Removed redundant automake and libtool from brew install.


stek29:
thank you

Issue #4907, opened 2018-06-27
[Feature Request] Increase voice/video messages playback speed


aNNiMON:
This feature already exists in Android beta apps. It would be nice to have it in desktop version.

### Steps to reproduce
1. Play voice message or video message

### Expected behaviour
2. Playback bar appears with `2X` button.
![image](https://user-images.githubusercontent.com/1321838/41991973-e4944b64-7a4f-11e8-8ca3-03e530cba62c.png)

3. By toggling `2X` button voice/video playback speed should increase twice.

### Actual behaviour
2. Playback bar appears without `2X` button.
![image](https://user-images.githubusercontent.com/1321838/41991964-df1cebfa-7a4f-11e8-935a-3df939e7e5d6.png)

gwpl:
https://telegram.org/blog/unread-replace-2x
"2X Voice Messages"
on Android it works for both: Audio and Periscopes.

I would love to have speed up options on desktop as well ! 
I am huge fan of speed listening.

PR #5307, opened 2018-10-26
Implement double playback speed

magnus-gross:
Hi, this PR adds the functionality of double playback speed (see #4907), just like in the Android app, which means:

- Double playback speed is available for voice messages and round video messages
- One can toggle at any time with the 2x button between 1x and 2x playback speed
- The setting is global
- The setting is saved after a restart

Thus, everything is consistent with the Android app.

alt text

Of course, the pitch is not modified at 2x speed, here is how it works:

OpenAL provides an `AL_PITCH` attribute, that I set to 2.0, which changes both the pitch and playback speed. To reset the pitch back to normal, I use the `AL_EFFECT_PITCH_SHIFTER` effect to decrease the pitch by 12 semitones = 1 octave.

This effect is applied by connecting an effect slot with the stream. However, the unmodified stream would be still playing, so I use an `AL_FILTER_LOWPASS` to mute the dry (direct) path. Note, that the lowpass filter is set to filter all frequencies. This filter is then connected with the stream and the result is a stream playing at twice the speed without altering the pitch.

I got the icon from: https://github.com/DrKLO/Telegram/blob/master/TMessagesProj/src/main/res/drawable-xxhdpi/voice2x.png

Note, that my code is mostly written independent of the actual playback speed (so something else then 2x would be possible), except for the part with the pitch shifter effect. There the input for the pitch shifter effect is the number of semitones and with equal-tempered tuning the calculation would require unnecessary logarithm computations and since we are using 2x playback speed only anyway, I did not add that.


23rd:
Did you test your changes?
When I run voice from Dialog, I see speed button, but if I enable 2x speed, I hear nothing.
When I run voice from Shared Media, Telegram shows me empty widget and runs *.ogg voice file from external player.
image

magnus-gross:
Of course, I tested it.

Does the playback stop completely, when you say that you hear nothing (which would indicate an OpenAL error) or does it just resume normally?

Also please post your operating system and used OpenAL version.
Does it say anything in the log?

john-preston:
@magnus-gross Also, log.txt should contain all OpenAL errors.

23rd:
@magnus-gross 
```
Audio Info: recreating audio device and reattaching the tracks
Audio Playback Error: 40963, Invalid Value
Audio Info: Closing audio playback device.
```

> Does the playback stop completely, when you say that you hear nothing (which would indicate an OpenAL error) or does it just resume normally?

Playback is on, but no sound.

> Also please post your operating system and used OpenAL version.

Windows 10 x64.
Not sure where I should see OpenAL version.

I just downloaded debug version from Appveyor artifact.
This is not normal if application needs to install extra lib.

magnus-gross:
I can reproduce the problem in Windows.
And as it turns out, the version of OpenAL that we are using is too old: https://github.com/kcat/openal-soft/issues/242

@john-preston Would it be possible to update the OpenAL version? Support for the pitch shifter effect was only added in version `OpenAL Soft 1.19`, which is essential for this feature.

magnus-gross:
@john-preston In order to not bother you with me breaking the required version of dependencies, I created a PR at #5362 that deals with this. The only thing that is left to do now, is updating the build in https://github.com/telegramdesktop/dependencies_windows :)

john-preston:
If I understand correctly in this code there is a race condition on Global::VoiceMsgPlaybackSpeed, because right now it is read from random threads being locked by AudioMutex:

reattach() -> createStream() -> updatePlaybackSpeed() -> VoiceMsgPlaybackSpeed()

And reattach() is called from different threads, for example from audio loader.

But it is also written in the main thread without holding the AudioMutex. I think I can fix that by not reading it directly from non-main thread, because I don't like the idea of guarding by mutex all accesses to Global::VoiceMsgPlaybackSpeed.

john-preston:
Also, it's not stable right now (the playback changes sporadically from 1x to 2x and back while switching between voice messages if 2x is enabled).

john-preston:
I've managed to fix that and fix the race condition, I'll push the fixes after the PR.

Thanks for your work! This can be an exception for the [no features and new UI elements](/telegramdesktop/tdesktop/blob/dev/.github/CONTRIBUTING.md#what-contributions-are-accepted) rule, because the feature is small and consistent with the mobile apps.

magnus-gross:
Great!
BTW I think you forgot to update the OpenAL build in https://github.com/telegramdesktop/dependencies_windows
Right now, the Windows build should be broken because there is still the old version in that repo ;)

john-preston:
@magnus-gross The build from appveyor is not used in production, but yes, it is better to update the dependencies as well.

agilesteel:
I know this is an old PR, but any chance double playback speed can be made available for mp3s as well? It works on Android, but there is no button on desktop.

PR #5338, opened 2018-11-01
Fixed display of files with 'audio/' MIME in Shared Media.

23rd:
This PR fixes display of files with "audio/" MIME in Shared Media.
  
`DocumentData::isAudioFile()` checks the MIME, which leads to incorrect display of files involved in the audio.
Ordinary playlist files, such as m3u, are perceived by Telegram as playable files. After click on this file, Telegram tries to play it, but gets fail.
  
As far as I understand, Telegram does not need to check the "audio/" MIME at all, since file type is received from the server; even if the mp3 file is without an extension, it will still play perfectly and will be placed in "Audio files" category.

So, we have to remove `_mimeString.startsWith(qstr("audio/"), Qt::CaseInsensitive);`, but then `isAudioFile()` will return true only in one case, when `isSong()` is true. And we can safely combine `isSong()` and `isAudioFile()` in single method.

If you think these changes should be separated into two pull requests, I will do it.

Despite the fact that many methods are replaced here, this pull request does not break anything, all the displays in the history, dialogs and Shared Media work correctly.

Before patch:
image

After patch:
image

### DIFF
--- a/Telegram/SourceFiles/data/data_document.cpp
+++ b/Telegram/SourceFiles/data/data_document.cpp
@@ -1292,7 +1292,13 @@ bool DocumentData::isAudioFile() const {
 	} else if (isSong()) {
 		return true;
 	}
-	return _mimeString.startsWith(qstr("audio/"), Qt::CaseInsensitive);
+	const auto prefix = qstr("audio/");
+	if (!_mimeString.startsWith(prefix, Qt::CaseInsensitive)) {
+		return false;
+	}
+	const auto left = _mimeString.midRef(prefix.size()).toString();
+	const auto types = { qstr("x-wav"), qstr("wav"), qstr("mp4") };
+	return ranges::find(types, left) != end(types);
 }
 
 bool DocumentData::isSharedMediaMusic() const {
--- a/Telegram/SourceFiles/overview/overview_layout.cpp
+++ b/Telegram/SourceFiles/overview/overview_layout.cpp
@@ -888,7 +888,7 @@ Document::Document(
 
 void Document::initDimensions() {
 	_maxw = _st.maxWidth;
-	if (_data->isAudioFile()) {
+	if (_data->isSong()) {
 		_minh = _st.songPadding.top() + _st.songThumbSize + _st.songPadding.bottom();
 	} else {
 		_minh = _st.filePadding.top() + _st.fileThumbSize + _st.filePadding.bottom() + st::lineWidth;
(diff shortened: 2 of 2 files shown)


john-preston:
The idea was to allow to play in-app (with in-app player) even the audio files that were not added the appropriate attributes on the server (because the sending app could not parse and set them or because simply it didn't). Perhaps the "audio/" is not the correct way to check for the audio files, maybe a list of mimes / extensions should be hardcoded. But by just removing it I'm afraid some files that were playing in-app before will stop now.

23rd:
@john-preston 
As far as I could test different files from [wiki list of formats](https://en.wikipedia.org/wiki/Audio_file_format#List_of_formats), Telegram plays only two formats with `audio/x-wav` (`audio/wav`) and `audio/mp4` (m4a) MIME types. Everything else that has `audio/` MIME and is in the Files category is opened by external programs with writing in log file:
```
Audio Info: recreating audio device and reattaching the tracks
Audio Info: Closing audio playback device.
```

As I can see in `layout.cpp::documentIsValidMediaFile()` you have hardcoded a lot of extensions already, so maybe it will be okay if we add in `isSong()` couple of hardcoded MIME types.
There are too many false positives going on with the current `audio/` check, сonsidering the variety of all audio formats and playlists.

john-preston:
@23rd isSong() should be exactly (type == SongDocument), that should be found exactly by "documentAttributeAudio" from server. Because this is the property that is used by the server to group them to "Shared audio files" section, so on the client side we should group there the same file types.

I think that NOT isSong() files should create a FILE layout in the "Shared files" section, not a AUDIO one. That will fix the issue in the screenshot.

Perhaps isAudioFile() should check something more strictly, then "audio/".

23rd:
Oops, I meant `isAudioFile()` of course.

>I think that NOT isSong() files should create a FILE layout in the "Shared files" section, not a AUDIO one. That will fix the issue in the screenshot.

If I understand you correctly, playable files should look like FILE in `Files` category but be able to be played?

Telegram/SourceFiles/data/data_document.cpp
```diff
@@ -1292,7 +1292,10 @@ bool DocumentData::isAudioFile() const {
 	} else if (isSong()) {
 		return true;
 	}
-	return _mimeString.startsWith(qstr("audio/"), Qt::CaseInsensitive);
+	auto mime = _mimeString.split("/");
+	if (mime[0] != qstr("audio")) return false;
+	static QSet<QString> playableTypes {"x-wav", "wav", "mp4"};
+	return playableTypes.contains(mime[1]);
```
john-preston: This method is called quite often (for example it can be called on each painting of the item), so it should be pessimised. Before it was just checking with a QLatin1String, no allocations, everything local.

Now it is .split(), which will allocate two strings on the heap + search in a set, jumping around a binary tree.

Better way is something like that:
```
const auto prefix = qstr("audio/");
if (!_mimeString.startsWith(prefix, Qt::CaseInsensitive)) {
  return false;
}
const auto left = _mimeString.midRef(prefix.size());
const auto types = { qstr("x-wav"), qstr("wav"), qstr("mp4") };
return ranges::find(types, left) != end(types);
```

23rd: Okay, replaced last commit.
Sorry I didn't figure it out myself.  ツ

john-preston:
Thanks!

PR #5362, opened 2018-11-07
Update OpenAL-Soft

magnus-gross:
#5307 requires OpenAL version 1.19.

This PR bumps the version at all places that I found. I left out old documentation such as `docs/building-xcode-old.md` though.
There is still a missing version bump in https://github.com/telegramdesktop/dependencies_windows (upload 1.19 build of OpenAL), but this should be the only thing missing and I will leave this one to you @john-preston, because it looks like you [patched](https://github.com/telegramdesktop/dependencies_windows/commit/4ff6c54484fc28372d1738162ae517fb056ee6af#diff-84b6d6b47ee0d8fe0a4a2c37f9981d76) it slightly.

After that, #5307 should be working on all platforms.

Note, that the travis build currently follows the HEAD of the openal-soft master branch. I changed it to a stable tag, just like it is already the case in all documentation.

### DIFF
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -40,7 +40,7 @@ FFMPEG_PATH="$BUILD/ffmpeg"
 FFMPEG_CACHE_VERSION="3"
 
 OPENAL_PATH="$BUILD/openal-soft"
-OPENAL_CACHE_VERSION="3"
+OPENAL_CACHE_VERSION="4"
 
 GYP_DEFINES=""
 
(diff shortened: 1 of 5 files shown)


john-preston:
Thanks! I'll check this one with my conf-s.

magnus-gross:
@john-preston Thanks, I will investigate that.
I will try making a minimal reproducable example and report that upstream and then will come back once that is fixed.

john-preston:
@magnus-gross 

I started bisecting it and while doing so I stepped on an another (still bad) sounding of a recorded voice message %) So there will be two different commits that break the recording :( I guess I'll find first (and post it here) and after it'll be fixed in openal-soft, I'll have to find the next one. Or we'll be lucky and both will be fixed together.

For now I'll have to revert the PR.

john-preston:
I've made a PR in upstream repository.

PR #5370, opened 2018-11-09
Refactored key handler in mediaview.cpp.

23rd:
`MediaView::keyPressEvent` method has huge unnecessary condition at the last `} else if ` step. 
It completely repeats all terms of internal conditions and makes the expression more complicated and less readable.
All enumeration of conditions on 2490 line can be safely removed, just leave the `Qt::ControlModifier` check.

### DIFF
--- a/Telegram/SourceFiles/mediaview.cpp
+++ b/Telegram/SourceFiles/mediaview.cpp
@@ -2445,9 +2445,10 @@ void MediaView::paintThemePreview(Painter &p, QRect clip) {
 }
 
 void MediaView::keyPressEvent(QKeyEvent *e) {
+	const auto ctrl = e->modifiers().testFlag(Qt::ControlModifier);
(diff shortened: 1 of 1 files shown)


Telegram/SourceFiles/mediaview.cpp
```diff
@@ -2445,9 +2445,10 @@ void MediaView::paintThemePreview(Painter &p, QRect clip) {
 }
 
 void MediaView::keyPressEvent(QKeyEvent *e) {
+	bool ctrl = e->modifiers().testFlag(Qt::ControlModifier);
```
john-preston: Please use `const auto variable = expression;` in the new code, where it is applicable.

23rd: Done.

john-preston:
Thanks.

PR #5377, opened 2018-11-11
Fixed editing of the last message on KeyUp.

23rd:
This small PR fixes an issue that has been around for a long time.
Use case: if the user sends a message, then sends sticker, and then presses the KeyUp, the message editing does not work. The same goes for video-messages and call-messages.

Here's an illustration. Client on the left is before patch, client on the right is after patch.


john-preston:
Thanks!

PR #5389, opened 2018-11-15
Refactored FallbackFontConfig() to use native Qt methods

xvitaly:
Refactored FallbackFontConfig() to use native Qt methods.


ZaMaZaN4iK:
@stek29 Do you have any questions/suggestions?

stek29:
@ZaMaZaN4iK sorry, I had connection issues and comment wasn't sent.
I think that it'd be better to abstract out a different function, something like RunShellCommandWithOutput, even if it's used just in one place (all other invocations of RunShellCommand don't need output).

john-preston:
@xvitaly Thanks!

PR #5425, opened 2018-11-29
Update building-msvc.md

hzexe:
Clone qt should with --recursive


john-preston:
I think it shouldn't. If you use "perl init-repository --module-subset=qtbase,qtimageformats" as said after, you'll fetch only those submodules. But if you clone recursive, you'll fetch all the submodules, which is a huge difference.

Issue #5169, opened 2018-09-17
IDs for chats / message authors


tp971:
When exporting in JSON format, neither chats nor message authors have a unique identifier:
* Chat objects in `chats.list[*]` only have `name` containing the display name, which is not unique and can be changed.
* Message objects `chats.list[*].messages[*]` only have a `from` field containing the display name of a user, which is not unique and can be changed.

Suggestions:
* Add a field like `id` to chat objects in `chats.list[*]`.
* Change the `from` field from string to an object containing a user id and a username (if set) and the displayname.

ShadowJonathan:
I would like to add to this and say that a singular chat also does not have an option to export to JSON *at all*, and the HTML that comes out of it has a structure that's not very easily scrapable.

I'm going to make an issue of it's own out of it, i just think it's mention-worthy here.

ShadowJonathan:
@tp971, fixed with #5426? Or are there other fields that would need to be added?

Aokromes:
no answer after near a year, closing.

Issue #5104, opened 2018-08-27
[Feature Request] Export chat data as JSON


OctoNezd:
Right now it is only possible to export chat data as HTML... 
![image](https://user-images.githubusercontent.com/15308404/44656363-3ecb4c00-aa01-11e8-8575-89bfa831f7f4.png)

...and whole telegram data as JSON or HTML
![image](https://user-images.githubusercontent.com/15308404/44656378-47bc1d80-aa01-11e8-8201-22c59a722a92.png)

It would be nice if I was able to export specific chat data as JSON

ShadowJonathan:
As i said in #5169, this is severely lacking in the current desktop client.

Another feature missing mentioned in that issue was the absence of IDs in JSON exports, making it impossible to be ingested into a relational database of sorts.

ShadowJonathan:
I'm going to start looking at Tdesktop's code, and prepare for a pull request that will fix both #5169 and this issue.

john-preston:
@ShadowJonathan Right now additional options for single chat history export (like html vs json) won't fit well in the design of the export configuration box :( But feel free to add IDs to JSON export format.

alegaballo:
@john-preston maybe setting a default options in the application menu could solve the issue, this way you wouldn't have to worry about the export configuration box

ShadowJonathan:
> @ShadowJonathan Right now additional options for single chat history export (like html vs json) won't fit well in the design of the export configuration box :( But feel free to add IDs to JSON export format.

Front-end has no precedence over Backend, in terms of "can-do", right now all i'm planning to do is implement a small re-write of the global chat export (from the settings menu), and then i'll dive into individual chat JSON exports.

In the latter case, it can be as easy as adding a checkbox which says "export in machine-readable format (JSON)"

Multimodcrafter:
I'd love to be able to export individual chats in json format! Please add it 😃

p0358:
Bump, this should be a no-brainer. If there's scroll in global chat export, it isn't any issue to include it here as well. It doesn't break anything, doesn't make anything worse, takes little effort to implement and the lack of it is a bummer.
Exporting chats by average users is not something used every day and this JSON export option is really needed. Non-advanced users can still take the advantage of analysis tools, and such exporting option would make it much easier for them. Please implement this.

p0358:
[...]

PR #5426, opened 2018-11-29
Add IDs to JSON exports

ShadowJonathan:
Currently, tdesktop doesn't have options to export only 1 chat to JSON,
Furthermore, it doesn't have unique IDs embedded in that exported JSON, which does not grant functionality to uniquely store or ingress data with unique IDs, mapped to the same message/chat/contact instant regardless of which client accesses it. (currently, it only provides string names as identifiers)

This commit must address above problems, adding radio buttons to solo chat exports, and adding ID fields to multiple unique objects in the JSON.

#### Initial PR Commit message:
```
Added IDs to various JSON objects.
Including an `_id` addition inside chat message, which get appended to `actor` or `from` entries.

Added HTML and JSON radio buttons to individual chat export window.
By effect, removed a "normalizing" entry which normalized solo chat only to be exported to HTML.
```

### DIFF
--- a/Telegram/SourceFiles/export/data/export_data_types.cpp
+++ b/Telegram/SourceFiles/export/data/export_data_types.cpp
@@ -617,6 +617,7 @@ User ParseUser(const MTPUser &data) {
 	auto result = User();
 	result.info = ParseContactInfo(data);
 	data.match([&](const MTPDuser &data) {
+		result.id = data.vid.v;
 		if (data.has_username()) {
 			result.username = ParseString(data.vusername);
 		}
--- a/Telegram/SourceFiles/export/data/export_data_types.h
+++ b/Telegram/SourceFiles/export/data/export_data_types.h
@@ -189,6 +189,7 @@ UserpicsSlice ParseUserpicsSlice(
 struct User {
 	ContactInfo info;
 	Utf8String username;
+	int32 id;
 	bool isBot = false;
 	bool isSelf = false;
 
--- a/Telegram/SourceFiles/export/export_controller.cpp
+++ b/Telegram/SourceFiles/export/export_controller.cpp
@@ -24,7 +24,6 @@ Settings NormalizeSettings(const Settings &settings) {
 		return base::duplicate(settings);
 	}
 	auto result = base::duplicate(settings);
-	result.format = Output::Format::Html;
 	result.types = result.fullChats = Settings::Type::AnyChatsMask;
 	return result;
 }
--- a/Telegram/SourceFiles/export/export_controller.cpp
+++ b/Telegram/SourceFiles/export/export_controller.cpp
@@ -24,6 +24,7 @@ Settings NormalizeSettings(const Settings &settings) {
 		return base::duplicate(settings);
 	}
 	auto result = base::duplicate(settings);
+	result.format = Output::Format::Html;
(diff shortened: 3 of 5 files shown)


ShadowJonathan:
Side-note: I saw preliminary code referencing YAML and TEXT export options, plus some code for said TEXT exporter, is there currently any effort to create such exporters?

(Pinging @john-preston, who made the commits including these format options)

ShadowJonathan:
(My apologies, this pull request should solve #5104 and #5169)

john-preston:
@ShadowJonathan The "text" was used initially for testing, nothing was done for yaml and there are no plans to.

john-preston:
@ShadowJonathan 

There is a problem with "add option to export JSON for solo chats". I advise you to split the PR, because it is easy to accept IDs in JSON format, but it's complicated for the new option.

(see "Unfortunately we do not merge any pull requests that have new feature implementations, translations to new languages and those which introduce any new user interface elements." in the Contributing Guidelines)

The problem: right now all options for single chat export fit inside the fixed-height export box without scrolling, they're nice and compact. And your new option, that will be used by a very small percent of "export single chat" feature users (most users are not geeks with automatic analysis tools, but they may be interested in a chat export), adds a lot of noise and clutter in this settings box and adds unnecessary scrolling.

This is not the same with the whole data export. It has this option because:
- it is a more advanced feature, hidden deeply in settings, not in the three-dot menu of every chat.
- it has already a lot of cluttering options and scrolling in the options box.
- GDPR requires such an option.

The second problem: for the HTML export of single chat a different format of output is used (different folder layout / files / paths) then for HTML export of all data. I'm sure to be done nicely the JSON export of single chat should have some changes in the export JSON tree / folder layout / files / paths, but no changes about that in this PR.

The second problem can be handled fairly easily, but the first one is not that small and requires a complicated solution (something like "Advanced options" button, that opens some box inside the window with export date limits and export formats) with some thought-through design.

ShadowJonathan:
Unfortunately I do not have the knowledge of Qt, C++, and this repository's intended structure to know how to implement such a new sub-menu, so I alone would not be able to solve #5104.

The second problem is also quite irritating and unfit for me to deal with because of the way that JSON is generated, without me knowing a structure to work towards, I would not know where to start.

I will roll back changes to the single chat export, plus UI, and leave it to a more experienced programmer.

ShadowJonathan:
To be clear; I am just reverting any UI changes and not messing with it because I fear I might not understand the structure of this repo, and because of aforementioned contribution excerpt.

This UI change is not that critical, just efficient and/or convenient for whomever wants to ingest just one chat, and not wait for other chats to finish exporting.

The UI change can even be very simple: Add a button which toggles the current layout of the export menu between "simple" and "advanced", with an internal boolean keeping track, and adding default values to the settings object once the user clicks "export", and it's in the "simple" mode.

Currently I'll let this PR just focus on adding IDs to JSON exports, which is a little more needed than a UI convenience.

john-preston:
_Telegram/SourceFiles/export/view/export_view_settings.cpp_ still shows some whitespace changes, please reset it to the original file.

ShadowJonathan:
I commited that little request, is there anything that needs to be changed before it gets merged?

john-preston:
@ShadowJonathan No, I think I can merge it in to the next version. Thanks.

Issue #5418, opened 2018-11-24
[Feature Request] Stripping @ in url


SmartManoj:
Stripping @ in url while resolving tg protocol
https://t.me/@GmailBot will just open telegram only
It works in telegram x

stek29:
That sounds more like Telegram X inconsistency, and something that should be fixed in TGX tbh,.

Lonami:
> and something that should be fixed in TGX tbh

What's the reason to keep this open?

PR #5432, opened 2018-12-02
 Stripping @ in url

SmartManoj:
#5418 If we copy username it includes '@'. If we add t.me before it to generate link,it would be easy since username doesn't contain '@' there is no problem.


john-preston:
And also they don't compile, because `usernameValue` is declared as `const`.

SmartManoj:
I tried to strip '@'

Issue #2958, opened 2017-02-01
Set native windows frame


kamuz:
Like into VK Messager

![2017-02-01_22-33-40](https://cloud.githubusercontent.com/assets/587267/22525380/9c181bee-e8cf-11e6-888d-a753b73fde69.png)

That is will be nice for Windows 7, 8, 10

It is better

![2017-02-01_22-35-37](https://cloud.githubusercontent.com/assets/587267/22525313/5e7367ee-e8cf-11e6-90cd-49e1da5f64ad.png)

then this

![2017-02-01_22-36-14](https://cloud.githubusercontent.com/assets/587267/22525324/67151938-e8cf-11e6-87e9-1538579ea293.png)

Twilightssuperb:
Only as option. To my mind, Telegram window looks OK with present frame. However the user should have a choice. 🤔

liquidplace:
It's necessary because the title bar is now shorter than before and quite difficult to _catch_ minimize button in the first try. In addition, there's no shortcut key for minimizing/closing the window to the tray icon like ESC or double ESC (esc key works only once), except ALT+F4 which is also confusing.

john-preston:
@liquidplace You can press Ctrl+W, it closes the window with minimizing to tray. Ctrl+M will minimize the window (leaving the icon in taskbar) and Ctrl+Q will quit the app.

Aokromes:
you can allways right click over bar and http://i.imgur.com/VkvKVEu.png

ralesk:
A much bigger issue than the bar being small is that it doesn't show difference between focused and unfocused window.

MasterGroosha:
@ralesk this is more theming issue since title bar colors are now separated (focused/unfocused states)

ralesk:
@MasterGroosha Ah, back in the day when window managers did that for you... :)

Aokromes:
this is needed for small sized windows tables, click on the buttons is hard on a sub 10" tablet.

maxoku:
I agree, there should be that option. I really hate when apps force their themes into UI elements like title bar.

pedrogarcia:
The Close, minimize and maximize buttons are too small on my computer, I can barely click them. If I increase the interface scale, then text and all other things become too big!

stek29:
@pedrogarcia That's #2818 :)

BoffinbraiN:
[...]

Issue #2818, opened 2017-01-12
Title bar too small with the new themes interface


virtualdj:
Themes are a great addition, even though I still prefer the classic one, but there's a quirck IMHO.
The **title bar became too small** with 100% UI size, comparing to other UI elements.

![title_bar](https://cloud.githubusercontent.com/assets/3251704/21902482/56fda4ee-d8fc-11e6-863f-49021465c3a4.png)

So minimizing or closing the window with the mouse on Windows is a lot more difficul than before.
Can this element height be adjusted, at least 200% than actual size?

bertmsk:
+++

cidwel:
Hope this mainbar is also customizable so we can keep bigger buttons

FireNero:
I'd like to see default OS title bar here instead of this small custom implementation. Or, at least, return title bar from previous version.

auchri:
@FireNero I also vote for a native title bar, since there are almost no differences

JensRex:
My biggest gripe with this new titlebar, is that it doesn't change state when the window gains or loses focus. There doesn't appear to be support for this in the theme files either.

I'd really prefer Telegram use native OS decorations, but AT LEAST fix this focus bug.

john-preston:
@JensRex Well, it never changed state, so it is not this new titlebar — it always worked that way. But yes, perhaps it could be themed.

john-preston:
@JensRex Native title bar will work far worse with color themes :( Currently I can't find any way to colorize it in Win32 apps, like it is possible in the latest macOS.

cobalto:
I just want it a little bigger, the old size was ok.
Its really hard to hit it now and completely discrepant with the system.

![image](https://cloud.githubusercontent.com/assets/1196314/21955228/a8233dfa-da4d-11e6-8923-56ebe4d0ec8e.png)

FireNero:
@john-preston Yes, completely forgot about color themes, native title bar won't look good with most of them. Then you can just increase title bar size as @cobalto mentioned and it will be fine.

luckydonald:
With themes: How about letting the themes set a size?

KreAch3R:
And without themes: try to emulate the default dimensions and artwork of the title buttons of the default Windows 10 theme.

AlexMedia:
[...]

PR #5434, opened 2018-12-02
Make Telegram Windows Controls more Native Looking

onenachi:
Windows Users unfortunately don't get a native titlebar. Instead they get a small titlebar with non windows like window buttons.

This commit basically makes the titlebar controls and titlebar size more like a native Windows 8+ program for windows users by changing the titlebar dimensions and adding titlebar window button assets that look nearly identical to the Windows 10 ones.

image


onenachi:
Referencing this relevant issue:

https://github.com/telegramdesktop/tdesktop/issues/2818

john-preston:
That's interesting. I think if we want to replicate the default Windows 10 top bar we need to check if the rendering is pixel-perfect in both regular window and maximized window (the title height and buttons sizes are different in maximized window for the native Windows 10 windows).

onenachi:
>           That's interesting. I think if we want to replicate the default Windows 10 top bar we need to check if the rendering is pixel-perfect in both regular window and maximized window (the title height and buttons sizes are different in maximized window for the native Windows 10 windows).

Unfortunately I don't know how to do maximized window sizes in QT. If you can give me an icon size, I can however make the icons for it.

onenachi:
>           That's interesting. I think if we want to replicate the default Windows 10 top bar we need to check if the rendering is pixel-perfect in both regular window and maximized window (the title height and buttons sizes are different in maximized window for the native Windows 10 windows).

Actually @john-preston the maximized window titlebar size in Windows 10 UWP applications is the same as when not maximized (tested for OneNote, Edge, Windows 10 Mail, the Windows 10 Store...).

john-preston:
@onenachi It doesn't change for Edge and other UWP apps I guess. But it does change for native apps, like File Explorer or mspaint (I guess you name that Old Windows Styling):

image

image

It makes sense for Edge, because it has tabs inside the title bar, but I don't see any sense for such thick bar in maximized tdesktop.

john-preston:
I see buttons 92x56 (I think it's 2x, so they're 46x28), maximized 92x44 (so 46x22). But this is Parallels on a Mac, so maybe your dimensions (46x30, 46x23) are more accurate.

john-preston:
So, in my system it doesn't exactly fit (top one native, bottom one tdesktop from this PR):

image

Native bar has less height and the icons are a bit less as well (this is 2x: 2x20, 20x20, 20x20).

onenachi:
> @onenachi It doesn't change for Edge and other UWP apps I guess. But it does change for native apps, like File Explorer or mspaint (I guess you name that Old Windows Styling):
> 
> image
> 
> image
> 
> It makes sense for Edge, because it has tabs inside the title bar, but I don't see any sense for such thick bar in maximized tdesktop.

Edge and UWP apps are native Apps. So are Visual Studio and so on. The apps you are talking about such as the Windows explorer are legacy Win32 applications that MS hasn't got around to rebuilding in the modern framework (Settings will eventually fully replace control panel and so on).

Obviously the easiest test for this is to develop a basic winforms/wpf application in dotnet core 3 since they were newly launched and see how they will scale. I will try that and post screenshots.

onenachi:
> So, in my system it doesn't exactly fit (top one native, bottom one tdesktop from this PR):
> 
> image
> 
> Native bar has less height and the icons are a bit less as well (this is 2x: 2x20, 20x20, 20x20).

Looking at some native and nonnative apps:

- Slack and other WinRT apps (windows 8) have 30px tall menubars maximized and 31px tall menubars restored plus a 1px border.
- UWP/Modern apps have 32px tall menubars maximized and 32px tall menubars restored plus a 1px border
- Legacy apps have a 23px tall menubar maximized and a 30px tall menubar restored plus a 1 px border
- Modern WPF and Windows Forms applications also render like legacy applications, which is even more confusing.

onenachi:
> So, in my system it doesn't exactly fit (top one native, bottom one tdesktop from this PR):
> 
> image
> 
> Native bar has less height and the icons are a bit less as well (this is 2x: 2x20, 20x20, 20x20).

The dimensions I have used is
                     (padding 10)
(padding 18) icon 10x10 (padding 18)
                     (padding 10)

So at 100% scaling they seem fine for me. I'll look into the scaling.

onenachi:
@john-preston check out my latest pull. It should be pixel perfect with the legacy style of things.

What is remaining:
- Well I cant seem to figure out how to do the 1px border in QT. The border must also disappear when maximized.

image

screenshot 16

john-preston:
@onenachi No border for now, I was not going to add it.

23rd:
I think this is not good idea, to make title bar "more Native Looking".
If you want it to look native, you need to consider drawing the title on all supported Windows systems, not only Win 10.
And even so, you spent a lot of time to perfectly fit the height pixels. But this PR does not take into account whether user has changed size of text in the title bar, which leads to a change in height of this title bar.

For example, on the value of the 8 title bar will be smaller, on the value of 12 - bigger.

I think the current title bar looks like it should.

image

onenachi:
@23rd IMO the problem with the current titlebar is it's simply too small. Especially if you happen to have a touch device which the rest of Windows 10 was designed for. And it looks out of place on Windows vs the Mac and Linux applications which look native. The Mac app is also a custom titlebar themed to look like Mac.

In addition, while I do agree that titlebar font doesn't respect Windows font scaling, titlebar size does respect Windows scaling as it did prior, and that's the major Windows scaling. It's a step up from the current titlebar.

I do agree that maybe going fully native with the titlebar and themeing that may be a better option, but as John said, coloring that perfectly to match telegrams theme is hard if not impossible.

Gordon01:
+1 for native window titlebar

grishka:
> coloring that perfectly to match telegrams theme is hard if not impossible

It is kinda possible by going deep into how Windows draws windows. For example, handling the `WM_NCPAINT` message allows apps to draw whatever they want over the native window border/titlebar. But then you can probably still use the custom title bar but scale it to match the system-defined height which [you can obtain programmatically](https://stackoverflow.com/questions/21566307/how-to-get-the-title-bar-height-and-width-for-aero-and-basic-design).

Yash-Garg:
Great idea of native titlebar 👍

yassine-ah:
The current one is too small, let get this one even if it is not 100% perfect
After having it, we can start talking about details and create new issues to improve it 😄

ArmeF97:
I don't think it's tiny. I'd make it even smaller, if it were to me. If you are going to add this, please at least let the users keep the small one, by a setting option

rezaei:
[Feature Request] grouping channels and groups and chats separately [make diffrent attributes such as color or category] #6420
 —— we need to have the ability to manage and categorize groups and channels and chats, because there are many channels and groups and chats that make me confuse and also make telegram unhelpful, I think its better for me if I can add these elements in the desired categorize with desired names like 'education' 'industry' 'family' 'friends' 'colleagues' 'sport' and so on ...

kcohar:
I think this is a great idea because the current title bar is too small and the only apps that I use that seem to have this problem are Telegram and Discord. This would make Telegram's title bar more like WhatsApp's and bring it up to snuff.

Does this patch work with the night/dark version too?

Aokromes:
Conflicting files
Telegram/SourceFiles/platform/win/window_title_win.cpp
Telegram/SourceFiles/window/window.style

Aokromes:
@onenachi

ilya-fedin:
Maybe it's better to get the buttons via winapi? Like Chrome & Firefox does
Then rendering will be pixel-perfect :thinking:

noobj2:
I think if they used native title bars now, it would look weird. i don't know if i want native title bars for telegram desktop after all these years of using non-native ones.
but yea, current title bar is too small, i would love to have a bigger title bar. same as it is, just a bit bigger.

tooomm:
I assume this is superseded with the fix of #2958 now?

john-preston:
yup

PR #5439, opened 2018-12-04
Allow UnityCounters for KDE5 resolves #4554

### LINKED ISSUE #4554: [Feature Request] Bring back Unread Badge Count in dash (Linux)
In Ubuntu 17.10 (Gnome 3.26) the badge indicating unread messages in the dash does not appear anymore. Only the icon in the system tray does show this badge. As far as I found out, this is because the badge display used some feature from Unity - I managed to bring back the badge count by starting Telegram with 
`env XDG_CURRENT_DESKTOP=Unity /snap/bin/telegram-desktop -startintray` 
in autostart. 

Since Unity won't return, I would be happy to get the badges back in Gnome sessions without using this workaround ;-)

kanocz:
UnityCounters is supported by KDE5 IconOnlyManager and Latte-dock, but using `env XDG_CURRENT_DESKTOP=Unity` causes some problems. This minimalistic change allows comfortable use Telegram in KDE5

### DIFF
--- a/Telegram/SourceFiles/platform/linux/linux_desktop_environment.cpp
+++ b/Telegram/SourceFiles/platform/linux/linux_desktop_environment.cpp
@@ -121,7 +121,7 @@ bool PreferAppIndicatorTrayIcon() {
 }
 
 bool TryUnityCounter() {
-	return IsUnity() || IsPantheon() || IsUbuntu();
+	return IsUnity() || IsPantheon() || IsUbuntu() || IsKDE5();
 }
 
 } // namespace DesktopEnvironment


magnus-gross:
What are those problems, that the env variable causes?
I am using this env variable, but I have not seen any problems.

xvitaly:
> why don't to use counters for KDE5 if it's just supported way and change is so small and will work for everybody?

1. Not all KDE users use Ubuntu.
2. Unity was deprecated by Canonical some months ago.

john-preston:
@xvitaly This PR will allow KDE5 **try** using unity counters. It'll try loading libunity and resolve the required methods. If they're not available, they just won't be used. Or there can be a problem with even the attempt to load and use them?

kanocz:
@xvitaly this is nit ubuntu-specific feature, this will for for any KDE5-based distribution, and yes, Unity is deprecated, but libunity for counters is still used even in latest versions of ubuntu which is not unity-based (but gnome-based)

john-preston:
@kanocz qq

kanocz:
@john-preston looks like it will only *try* to load libunity, but nothing happen it no valid libunity library found in system

magnus-gross:
I can confirm, that at least when using the env variable, that if `libunity` is not installed, Telegram will work just fine, it will just not show a badge on the icon.

john-preston:
Ok, thanks.

PR #5450, opened 2018-12-07
Generalize Unity Integration to not depend on libunity

Johnnynator:
This removes the need of libunity in favor of a more general approach of setting the Icon Counter to have it working on Linux distributions that don't ship Unity. While DBus API is marked as unstable I highly doubt that it will change any time soon since unity is deprecated. Also removes the `TDESKTOP_DISABLE_UNITY_INTEGRATION` since it is not really needed anymore. It will probably need some additional changes for Snaps but should work fine otherwise as it seems to work for my systems.

Edit: desktopFile needs to be set to `snap-name-telegramdesktop.desktop` to work. Also I'm aware of the whitespace issues, will fix them later.


Johnnynator:
The code itself should work. The Snap AppArmor Policy needs some changing since it currently blocks the dbus message.

john-preston:
@Johnnynator Can you provide me with some more details? If I merge this, where and what exactly should I change so that it'll work not worse than it worked before?

Johnnynator:
> @Johnnynator Can you provide me with some more details?

Not much currently, I'm out of reach of a machine with support for snaps, so nothing I can do before next year. Generally the `com.canonical.Unity` dbus Interface needs probably be whitelisted for Telegram in snap. I haven't read the documentation about it yet. Maybe @3v1n0 can help.

> If I merge this, where and what exactly should I change so that it'll work not worse than it worked before?

Outside of snaps it should work better as it does now

Telegram/SourceFiles/platform/linux/main_window_linux.cpp
```diff
@@ -543,23 +535,28 @@ void MainWindow::psCreateTrayIcon() {
...
+			} else if(!QStandardPaths::locate(QStandardPaths::ApplicationsLocation, "Telegram.desktop").isEmpty()) {
+				_desktopFile = "telegramdesktop.desktop";
+				LOG(("Found Unity Launcher entry telegramdesktop.desktop!"));
+				useUnityCount=true;
 			} else {
 				LOG(("Could not get Unity Launcher entry!"));
 			}
+		} else {
+			LOG(("SNAP Enviroment detected, setting Launcher entry to %1-telegramdesktop.desktop!").arg(snapName));
+			_desktopFile = snapName + "_telegramdesktop.desktop";
```
3v1n0: You can probably keep `telegramdesktop.desktop` as a generic shared global name

Johnnynator: @3v1n0 Good to know, will test it later, I just mimicked the behavior of libunity here.

Johnnynator: I finally found time to look at it again, will push changes after verified that they are working locally.
I need to set the `app_uri`/`desktop-file` this way, since the default AppArmor profile demands that the called path is `/com/canonical/unity/launcherentry/[0-9]*` where `[0-9]*` is the djb hash of the app_uri.

Johnnynator:
Should be ready for review now, an unsigned snap package of the latest commit can be found here http://144.76.1.220/void/ `sha256sum 4a5244ba4eac8ac0845d7f2b51968559f031d1d08b0ada73bc9953197e3e9f68`

john-preston:
Sorry, I've broken something here. But when the conflicts are resolved, I'll add this to the next beta release. Thanks for your work!

Johnnynator:
Ok rebased on your changes.

john-preston:
Thanks!

PR #5491, opened 2018-12-18
Add these limits to admin permission [REQUEST]

har4s:
Add these limits to admin permission
- Access to subscribers list
- Access to admins list
- Access to recent actions

### DIFF
--- a/Telegram/SourceFiles/boxes/about_box.cpp
+++ b/Telegram/SourceFiles/boxes/about_box.cpp
@@ -87,12 +87,18 @@ void AboutBox::keyPressEvent(QKeyEvent *e) {
 }
 
 QString telegramFaqLink() {
-	auto result = qsl("https://telegram.org/faq");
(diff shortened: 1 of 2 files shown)


fabyRojas:
Suggestions:
- Possibility to make video messages and take pictures from a desktop telegram.
- Voice messages separated from video messages in multimedia files
- Streaming and secret chat.
- Tabs (or even if they appear as a hidden secret as the workmode)
- Folders (as a way to store chat, groups or channels by category, according to each user), with tabs and chat anchors that option would be perfect

fabyRojas:
Suggestions:
- Possibility to make video messages and take pictures from a desktop telegram.
- Voice messages separated from video messages in multimedia files
- Streaming and secret chat.
- Tabs (or even if they appear as a hidden secret as the workmode)
- Folders (as a way to store chat, groups or channels by category, according to each user), with tabs and chat anchors that option would be perfect

PR #5500, opened 2018-12-20
Fixed caption when sending single WEBP as file.

23rd:
This PR fixes losing of caption if the user uploads single webp as file.
The tdesktop sends the webp as sticker. Stickers have no captions, so a typed text just loss.
If the user sends more than 1 webp file, a caption is sent as a separated message just like as it should.
This patch implements same behavior but for a single webp file.
This works well even with renamed jpg/png extensions to webp.


Telegram/SourceFiles/apiwrap.cpp
```diff
@@ -4302,7 +4302,14 @@ void ApiWrap::sendFiles(
 		TextWithTags &&caption,
 		std::shared_ptr<SendingAlbum> album,
 		const SendOptions &options) {
-	if (list.files.size() > 1 && !caption.text.isEmpty()) {
+	const auto isSticker = [&]() {
```
john-preston: `()` are redundant here, everywhere in code just `[&] {` is used.

john-preston:
Yes, this should be good to fix.

Telegram/SourceFiles/apiwrap.cpp
23rd: Done.
Missed this moment.

john-preston:
I'll add that to the next version, thanks.

PR #5509, opened 2018-12-23
 Small refactor of an editing of the last message on KeyUp.

23rd:
This PR just slightly optimizes the way KeyUp editing code works.
At the moment the code works in such a way that the `lastSentMessage()` method is called first, which has within a double `for` loop, and only after loops complete the work, there are other simple checks, such as `_field->empty()`.
So it makes sense to swap these conditions.
Fistly simple checks.
```
if (_field->empty() && !_editMsgId && !_replyToId) {
```
And if it's ok, start looking for the last message through the loops.
```
const auto item = _history ? _history->lastSentMessage() : nullptr;
```


john-preston:
Thanks. But I don't consider this as an improvement, sorry :(

It could be considered as an optimization if this code was running at least in mouse-move handler (or better painting :) or smth else very frequent.

But in this case it is on a key up press, which is a rare event. And it complicates code a bit, adding one level of nesting.

PR #5511, opened 2018-12-23
Update building-xcode.md

caffeinum:
We don't need this line: we're already in **_BuildPath_**

### DIFF
--- a/docs/building-xcode.md
+++ b/docs/building-xcode.md
@@ -21,7 +21,7 @@ Go to ***BuildPath*** and run
     MACOSX_DEPLOYMENT_TARGET=10.8
 
     ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
-    brew install automake fdk-aac git lame libass libtool libvorbis libvpx opus sdl shtool texi2html theora wget x264 xvid yasm
+    brew install automake fdk-aac git lame libass libtool libvorbis libvpx ninja opus sdl shtool texi2html theora wget x264 xvid yasm
 
     sudo xcode-select -s /Applications/Xcode.app/Contents/Developer
 


docs/building-xcode.md
```diff
@@ -26,7 +26,8 @@ Go to ***BuildPath*** and run
     sudo xcode-select -s /Applications/Xcode.app/Contents/Developer
 
     git clone --recursive https://github.com/telegramdesktop/tdesktop.git
-
+    
+    mkdir Libraries
```
john-preston: You've unpacked "xz-5.0.5" to BuildPath/Libraries/xz-5.0.5 above, so no need to create it here.

docs/building-xcode.md
```diff
@@ -135,6 +136,9 @@ Go to ***BuildPath*** and run
     ./configure -prefix "/usr/local/tdesktop/Qt-5.6.2" -debug-and-release -force-debug-info -opensource -confirm-license -static -opengl desktop -no-openssl -securetransport -nomake examples -nomake tests -platform macx-clang
     make -j4
     sudo make install
+    cd ../..
```
john-preston: You were in BuildPath/Libraries/qt5_6_2 and then you write `cd ../..` going to BuildPath and then one more `cd ..` below, going no one knows where. In the old version you just stayed in Libraries, like after any other library build.

john-preston:
I didn't understand necessity of either of those two changes.

docs/building-xcode.md
caffeinum: Okay, that makes sense. Did not notice it.

docs/building-xcode.md
caffeinum: That makes sense according to the code, but during the build, I did not arrive at the right directory. Okay, let's say I just had some typo.

caffeinum:
@john-preston ok, these were my errors. The instruction does not work, still. See #5512.

For example, it lacks `ninja` install.

```bash
$ build/gyp_crashpad.py -Dmac_deployment_target=10.8
$ ninja -C out/Debug
-bash: /Users/caffeinum/tbuild/depot_tools/ninja: No such file or directory
$ ninja -C out/Release
-bash: /Users/caffeinum/tbuild/depot_tools/ninja: No such file or directory
```

docs/building-xcode.md
```diff
@@ -21,12 +21,13 @@ Go to ***BuildPath*** and run
...
     ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
-    brew install automake fdk-aac git lame libass libtool libvorbis libvpx opus sdl shtool texi2html theora wget x264 xvid yasm
+    brew install automake fdk-aac git lame libass libtool libvorbis libvpx ninja opus sdl shtool texi2html theora wget x264 xvid yasm
 
     sudo xcode-select -s /Applications/Xcode.app/Contents/Developer
 
     git clone --recursive https://github.com/telegramdesktop/tdesktop.git
-
+    
+    mkdir Libraries
```
caffeinum:
```suggestion
```

caffeinum:
Well, now it worked for me without `ninja` installed.

```bash
git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
export PATH="$PWD/depot_tools:$PATH"
```

Probably this line exports it.

Somehow it didn't install `depot_tools` on the first run. Well, then I have no other objections to the current version. :)

PR #5520, opened 2018-12-25
Fixed an editing of own unsupported messages.

23rd:
This PR fixes an issue, when a user can edit own unsupported messages (sent from an other client).
Of course after a resending of the edited message user receives `You cannot edit this message` alert. 
I think this is not right that the user has the ability to go into edit mode.

image

In the code.
I think it makes no sense to create a new subclass of unsupported message, just add one bool variable.
As I could see in other classes, it's normal to create a setter, that will be called from only one place. 
For example:

https://github.com/telegramdesktop/tdesktop/blob/e5536880fb9de5a46cc261e1f38c36230ec9f1c4/Telegram/SourceFiles/boxes/peer_list_box.h#L140-L145

I'm not sure that `!_isUnsupportedMessage` condition should be placed in `isTooOldForEdit()` method, so, unfortunately, I had to complicate a little the condition that you have recently simplified (by a moving of some things to `canStopPoll()`) in the `allowsEdit()` method.

### DIFF
--- a/Telegram/SourceFiles/history/history_item.cpp
+++ b/Telegram/SourceFiles/history/history_item.cpp
@@ -63,6 +63,7 @@ not_null<HistoryItem*> CreateUnsupportedMessage(
 	text.entities.push_front(
 		EntityInText(EntityInTextItalic, 0, text.text.size()));
 	flags &= ~MTPDmessage::Flag::f_post_author;
+	flags |= MTPDmessage_ClientFlag::f_is_unsupported;
 	return new HistoryMessage(
 		history,
 		msgId,
--- a/Telegram/SourceFiles/history/history_message.cpp
+++ b/Telegram/SourceFiles/history/history_message.cpp
@@ -603,7 +603,8 @@ bool HistoryMessage::isTooOldForEdit(TimeId now) const {
 bool HistoryMessage::allowsEdit(TimeId now) const {
 	return canStopPoll()
 		&& !isTooOldForEdit(now)
-		&& (!_media || _media->allowsEdit());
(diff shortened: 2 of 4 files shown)


john-preston:
I don't like an idea of adding a separate bool variable to all messages just for this very rare case.

For now I can offer adding a new `MTPDmessage_ClientFlag` (instead of unused `f_hidden_by_group`, for example) and use flags bit for that. In the future I think there'll be a special media type that will hold "can't edit" info.

23rd:
@john-preston Updated the code.

I'm not sure whether to leave the current method with `bool value` argument or replace it with a method that just sets the `f_is_unsupported` flag bit, given that the method will never be called with `false`.

Telegram/SourceFiles/history/history_item.cpp
```diff
@@ -73,6 +73,8 @@ not_null<HistoryItem*> CreateUnsupportedMessage(
 		from,
 		QString(),
 		text);
+	historyMessage->setIsUnsupportedMessage();
```
john-preston: I would suggest not adding `setIsUnsupportedMessage` at all. There is no need to ever set it after the message is created and here is a perfectly valid place to add this flag - in `CreateUnsupportedMessage` function. Also, the future message flags are modified here anyway, so you can just add this flag here.

(also use `const auto` instead of `auto` where it is possible, but here there is no need for a variable at all).

23rd: Done.

john-preston:
Thanks!

PR #5523, opened 2018-12-26
Fix some insecure links

ArmeF97:


Telegram/SourceFiles/ui/widgets/input_fields.cpp
```diff
@@ -2361,7 +2361,7 @@ TextWithTags InputField::getTextWithAppliedMarkdown() const {
 	const auto &originalTags = _lastTextWithTags.tags;
 
 	// Ignore tags that partially intersect some http-links.
-	// This will allow sending http://test.com/__test__/test correctly.
```
john-preston: Why would you change that?..

ArmeF97: http is going to be shut down (sooner or later), so even a test link, it's better to upgrade it

Telegram/Resources/qrc/qmime/freedesktop.org.xml
```diff
@@ -58,7 +58,7 @@
   <!ELEMENT sub-class-of EMPTY>
   <!ATTLIST sub-class-of type CDATA #REQUIRED>
 ]>
-<mime-info xmlns="http://www.freedesktop.org/standards/shared-mime-info">
```
stek29: [xmlns should not be changed](https://stackoverflow.com/questions/30707609/xml-namespace-uri-with-https)

Telegram/Resources/qrc/qmime/freedesktop.org.xml
```diff
@@ -372,7 +372,7 @@
     <alias type="text/mathml"/>
     <sub-class-of type="application/xml"/>
     <glob pattern="*.mml"/>
-    <root-XML namespaceURI="http://www.w3.org/1998/Math/MathML" localName="math"/>
+    <root-XML namespaceURI="https://www.w3.org/1998/Math/MathML" localName="math"/>
```
stek29: Same thing applies here

stek29:
You should not change links in xml, and the only change that's left is test.com one, which doesn't make much sense either, sorry.

Issue #5525, opened 2018-12-27
Incorrect count notifications


WiRight:
### Steps to reproduce
1. Install TDesktop on new machine
2. In tray i see 1 or more unread messages (but no one exist)

### Expected behaviour
No one notify (0)

### Actual behaviour
1 or more show

### Configuration
**Operating system:**
Linux 4.9.0-8-amd64 #1 SMP Debian 4.9.130-2 (2018-10-27) x86_64 GNU/Linux

**Version of Telegram Desktop:**
1.5.4

**Used theme**:
Standard

RememberTheAir:
I've had a similar problem on 1.5.4 too. The unread counter was showing one notification less than the actual count (eg. with no unread messages, the tray icon counter was showing "-1"). This both on Linux Mint and Windows 10. I've tdesktop set up to count the chats (and not the number of messages), and only non-muted chats. I don't know how to reproduce

biancofla:
Same issue here. Currently, I'm on Ubuntu 16.04 with Telegram 1.5.4 and the tray icon shows constantly one unread message, even if no one is really present.

dimon2242:
Same here on Ubuntu 16.04

WiRight:
Reproduce it on ArchLinux.
After install TDesktop, i haven't read some messages (7 items) and reboot my system.
After start in tray TDesktop always showing const value 7 unreaded messages

emilengler:
Same issue here on Ubuntu 18.04

WiRight:
@martinaglv Thx! That fix me it for now!
But in new machine bug will be reproduced

emilengler:
@martinaglv I'm not sure but I think you mean the `_trayIconImageFile()` function.

biancofla:
After a research in the commits history, I've found this: https://github.com/telegramdesktop/tdesktop/commit/0f67f75bed0cc12f354450ffd032e8bdeaa4f15a. This specific commit seems to be the initial cause of the broken tray icon.

martinaglv:
@biancofla thank you for tracking it down! Maybe @john-preston has an idea how to fix this.

emilengler:
I created #5556.
This basically undo the change and should hopefully work.

RafaelKr:
The workaround @martinaglv does it for me. The icon has the correct name, but the image is wrong.

This means, that the image is generated with a wrong message count.

[...]

PR #5556, opened 2019-01-06
Hopefully a fix for #5525

emilengler:
As described here #5525 there is an issue with the message counter mostly on Linux systems.
The bug was added with the commit https://github.com/telegramdesktop/tdesktop/commit/0f67f75bed0cc12f354450ffd032e8bdeaa4f15a.

I provided this fix which basically undo the change.
I haven't tested it because I have a barely knowledge on Qt and don't have a Qt dev environment installed at the moment.

### DIFF
--- a/Telegram/SourceFiles/history/history.cpp
+++ b/Telegram/SourceFiles/history/history.cpp
@@ -974,8 +974,8 @@ bool History::addToUnreadMentions(
 		: false;
 	if (allLoaded) {
 		if (type == UnreadMentionType::New) {
+			++*_unreadMentionsCount;
 			_unreadMentions.insert(msgId);
-			setUnreadMentionsCount(*_unreadMentionsCount + 1);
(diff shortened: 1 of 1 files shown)


john-preston:
Why do you think this can cause such an issue? Unread mentions count is not related to the badge counter.

emilengler:
@john-preston People in the issue came to the conclusion that this commit created the bug. I took a look at it and it seemed that it was the only modification which changed the increment/decrement of a variable.

john-preston:
@emilengler I have no idea why they decided this. At least someone proposing a fix should check that it indeed fixes something, sorry.

john-preston:
@emilengler Right now your change would at least break the thing the mentioned commit was fixing.

emilengler:
@john-preston Ok then excuse me. This change in the commit was the only logical thing which could have created the bug.

PR #5608, opened 2019-01-23
Added an icon for bots in the dialog list.

23rd:
This PR adds an icon for bots in the dialog list.
It's pretty weird that groups and channels have own icons, but the icon for bots is missed.
In the Telegram for macOS and in the Telegram for iOS channels, groups and bots have no any icons.
In the Telegram for Android icons are presented for all types of dialogs.
In the Telegram Desktop icons are presented, but not completely. =)

Source of the icon:
https://github.com/DrKLO/Telegram/blob/master/TMessagesProj/src/main/res/drawable-xxhdpi/list_bot.png

image
image
image
image


john-preston:
Thanks! I'll add this to the next beta release.

Issue #5607, opened 2019-01-23
MediaView does not fill the entire screen. Linux.


thirtysat:
### Steps to reproduce
1. Start telegram-desktop
2. Open any media (photo, video)
3. Close the view and open it again.

### Expected behaviour
Media viewing should fill the entire screen.

### Actual behaviour
MediaView initially shows everything is normal. All subsequent opening of viewing media files shift view under the taskbar (tint2, xfce4-panel), closing some of the control elements located below. 

If you close Telegram and restart, the first launch of viewing the media will be normal, the subsequent ones will be broken.

### Configuration
**Operating system:** Manjaro XFCE

**Version of Telegram Desktop:** 1.5.8 official

**Used theme**: Any, original or custom

**Screens**:
![_2019-01-23_17-29-45](https://user-images.githubusercontent.com/29095597/51618616-49016000-1f37-11e9-9ffd-e183a444ef1b.png)
![_2019-01-23_17-30-20](https://user-images.githubusercontent.com/29095597/51618617-49016000-1f37-11e9-9baa-6368e4903d28.png)

Aokromes:
https://github.com/telegramdesktop/tdesktop/issues/5075

thirtysat:
Many thanks, now it works as it should!

Issue #5075, opened 2018-08-12
Picture is not fullscreen


Djaler:
### Steps to reproduce
1. Open any picture in message 

### Expected behaviour (and it works like this earlier)
Picture is opened on fullscreen

### Actual behaviour
The image together with the interface is shifted down relative to the top panel

### Configuration
**Operating system**: elementary OS Juno (based on Ubuntu 18.04)

**Version of Telegram Desktop**: 1.3.12 alpha

![2018-08-12 12 58 05](https://user-images.githubusercontent.com/7964583/44000864-30ef7958-9e30-11e8-9d12-9c030c96866b.png)

evan3334:
Can confirm this is also an issue on Ubuntu 18.04. 
![image](https://user-images.githubusercontent.com/11367199/44759716-795fe200-aaf0-11e8-9b00-177526090ab0.png)
The fullscreen image view can also be dragged around like a normal window if you grab near the top:
![image](https://user-images.githubusercontent.com/11367199/44759761-af9d6180-aaf0-11e8-8e0e-c4af708acf52.png)

rleppink:
Can confirm. The image only fullscreens correctly once, the first image you open after starting Telegram. After that, all images are opened in windowed mode.

stek29:
#5048 related

esl51:
The same problem on Manjaro XFCE with Telegram Desktop 1.5.4.

sprite-1:
Is there a way to force the picture overlay to be on top of all other windows? As a temporary workaround until the issue gets resolved

Djaler:
I can confirm this. Thanks!

Issue #5048, opened 2018-08-03
Image viewer regression in tiling window managers


PandorasFox:
### Steps to reproduce
1. Use a tiling window manager (ie i3)
2. Open a picture to view
3. picture viewer pops in its own window. If I float the window, it does not properly align to the top of my screen.

### Expected behaviour
The behaviour before was much better. It popped over the full telegram window as expected (even if it was a bit wonky when I'd change workspaces). It should just be attached to the telegram window and not its own window.

### Actual behaviour
I end up with two windows (telegram + telegram image view)

Example when viewing a profile pic:

![image](https://user-images.githubusercontent.com/3235827/43624527-dd9923e8-969c-11e8-986d-78ef3954a4fc.png)

### Configuration
**Operating system:** Arch Linux

**Version of Telegram Desktop:** 1.3.11 Alpha

**Used theme**: built-in night mode

john-preston:
@PandorasFox I don’t know what to do with that one :( the old behavior was glitchy in a number of ways, because of inability to have its own focus or whatever. When app lost focus on some DEs (by alt tab) it would just stop the process unresponsive until manually killed. Perhaps a handcoded list of exceptions is the only way :/

Ristovski:
Can confirm on Gentoo Linux with XMonad WM.

@john-preston Perhaps an option in the settings menu that allows reverting back to the old behavior would suffice?

john-preston:
@Ristovski I'm not sure an option would be easy to understand. Is there an easy way of DEs that should use the old behaviour? The ones, that don't support separate fullscreen windows.

Ristovski:
@john-preston Maybe an environment variable? 

This bug mainly affects tech-savvy Linux users who use tiling window managers.
(_something that imo requires you to at least have basic enough knowledge to be able to prepend a environment variable before a command_). 

Having a runtime env var (`TG_USE_OLDWINDOWING` or something similar) would be quite nice as it would both allow users who experience this to launch Telegram with the old behavior while not adding any new GUI options that could confuse other users.

PandorasFox:
@john-preston When I inspect the image-viewer window, I see basically the same window attributes:

```
┌arcana@lapfox-(~)   24h:58m:24s
└> ~/dotfiles/bin/i3-get-window-criteria.sh 
[class="TelegramDesktop" id=6291463 instance="Telegram" title="TelegramDesktop"]
┌arcana@lapfox-(~)   
[...]

PR #5615, opened 2019-01-25
fixing mediaview fullscreen on X11

farseerfc:
Trying again to fix fullscreen of mediaview on X11, related to #5607 , #5075 and #5048.

Tested it fixed mediaview in Arch Linux under:
* KDE Plasma (kwin 5.14.90)
* i3wm (4.16)

But strangely AwesomeWM 4.2 still misplaced the fullscreen window. But previous version of awesome (3.4 tested) was able to show the fullscreen window correctly so I suspect there are some bugs with  awesomewm 4.2 's end.

### DIFF
--- a/Telegram/SourceFiles/mediaview.cpp
+++ b/Telegram/SourceFiles/mediaview.cpp
@@ -138,7 +138,7 @@ MediaView::MediaView()
 	});
 	handleAuthSessionChange();
 
-	setWindowFlags(Qt::FramelessWindowHint);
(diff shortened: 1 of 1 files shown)


farseerfc:
Also tested fixed the fullscreen problem on Gnome 3.30.2

john-preston:
Thanks! I hope it didn't break at least on Ubuntu :) Then I'll add this to the next beta release.

Issue #5528, opened 2018-12-27
my last seen is online after closing telegram desktop


siamak2:
### Steps to reproduce
1. open telegram desktop.
2. close telegram desktop using x button.

### Expected behaviour
my last seen should be like "last seen today at ##:##".
### Actual behaviour
my last seen is online.
### Configuration
**Operating system:**
windows 7
**Version of Telegram Desktop:**
1.5.4
**Used theme**:
classic

john-preston:
@siamak2 Maybe it is for a short time?

siamak2:
my last seen is online for minutes after closing tdesktop. I didn't check it to see whether I'm online for hours or not.
I want my last seen to be offline (last seen today at ##:##) immediately after closing tdesktop.
currently I minimize tdesktop to go offline and then right click on it and close it (in this way I go offline immediately).

Edicksonjga:
There is no problem in Beta 1.5.5 with Windows 7, if your telegram is in the foreground (shown on the screen) after 30 seconds of inactivity, your status "Online" changes to "Last time ...", if minimizes your telegram (shown in the task bar) its status "Online" immediately changes to "Last time ..." (if it does not change, exit and enter the chat again and you will see that it has changed), if you close your telegram in the "X" (not shown in the taskbar), its status "Online" will immediately change to "Last time ...".

By the way, in Linux, the disconnection for inactivity after 30 seconds works, but when they write to you, their status "Online" only activates and marks the messages as read, this does not happen in Windows. #4664

Operating system:
windows 7
Version of Telegram Desktop Beta:
1.5.5
Used theme:
classic

siamak2:
I tried beta 1.5.5 but the problem still exist.
in both version (1.5.4 and 1.5.5 beta) my last seen status is online for 20 minutes after closing tdesktop.
I want my last seen status to go offline immediately (or a few seconds later) not after 20 minutes.
minimizing is OK (my last seen status go offline immediately).

PR #5618, opened 2019-01-26
Fixed a keeping online status after application quit.

23rd:
[This issue](https://github.com/telegramdesktop/tdesktop/issues/5528) occurs because the application does not send the offline status when it quits.

Use case 1:
- User wants to quit the application.
- User opens a tray context menu by right click. — In this moment the main window loses focus and the application makes the status offline.
- User clicks "Quit Telegram". 
- Everything is okay.

Use case 2:
- User wants to quit the application.
- User presses `Ctrl + Q`.
- The application closes without changes of the online status.
- User is still online for awhile.

After this patch, user after pressing `Ctrl + Q` will have the offline status immediately.

### DIFF
--- a/Telegram/SourceFiles/core/application.cpp
+++ b/Telegram/SourceFiles/core/application.cpp
@@ -1039,6 +1039,11 @@ void Application::preventWindowActivation() {
 void Application::QuitAttempt() {
 	auto prevents = false;
 	if (AuthSession::Exists() && !Sandbox::Instance().isSavingSession()) {
+		if (auto mainwidget = App::main()) {
+			if (mainwidget->isQuitPrevent()) {
+				prevents = true;
+			}
+		}
 		if (Auth().api().isQuitPrevent()) {
 			prevents = true;
 		}
--- a/Telegram/SourceFiles/mainwidget.cpp
+++ b/Telegram/SourceFiles/mainwidget.cpp
@@ -3777,7 +3777,14 @@ void MainWidget::updateOnline(bool gotOtherOffline) {
 
 		_lastWasOnline = isOnline;
 		_lastSetOnline = ms;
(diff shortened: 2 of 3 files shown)


john-preston:
I guess this depends now on whether the app will be able to quit before the network request will be sent or after. Kind of a race.

A more complex (but better) approach would be to send this request and wait for small determined time for it to complete. This is done already for the cloud drafts (when you quit with unsaved drafts the save request is sent and app waits for some time: if the requests are done or failed then we quit right away, if they hang - for example we don't have the network right now - we quit after a timeout).

23rd:
@john-preston Okay, I updated the code.

Not sure but I think it would be better to call `MTP::send(MTPaccount_UpdateStatus)` with `done` and `fail` handlers only when `App::quitting()` is `true`.

By the way, I guess if you'll apply this code, then next lines will be unnecessary.
https://github.com/telegramdesktop/tdesktop/blob/69b1f6c4e15f3d96e2b6c37c4fb2c76a50a05692/Telegram/SourceFiles/app.cpp#L817-L819

A method `saveDraftToCloud()` is called already inside a method `updateOnline()`.

john-preston:
Sorry, I've broken this one with my refactorings.

23rd:
Not big deal. 
Code is updated.

john-preston:
Thanks!

john-preston:
@23rd (by fix I refer to the `isOnline` definition, not adding of `const` :)

Regarding `const` - all tdesktop code is awful legacy mess :/ I slowly change some parts of coding style, now I try to define all variables as `const auto name = value;` if you can, and only `auto name = value;` if you have to.

Speaking of legacy, I tried to move all the online updating code from `MainWidget` to `ApiWrap`, but decided to postpone that, because existing changes should be released as soon as possible.

PR #5656, opened 2019-02-05
Set urgent to X11 for linux

### LINKED ISSUE #223: Feature Request: Set X11 urgent flag on incoming message
right now I only get a notification popup. If I don't see the popup (e.g. because I'm afk), I'll never notice that someone wrote a message. Other IM clients use the urgent flag as a permanent modifier.

Mephistophiles:
Fixes #223

### DIFF
--- a/Telegram/SourceFiles/platform/linux/main_window_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/main_window_linux.cpp
@@ -343,6 +343,10 @@ void MainWindow::unreadCounterChangedHook() {
 void MainWindow::updateIconCounters() {
 	updateWindowIcon();
 
+	foreach(QWidget *widget, qApp->topLevelWidgets()) {
+		QApplication::alert(widget);
+	}
+
 	const auto counter = Core::App().unreadBadge();
 
 	if (useUnityCount) {


Mephistophiles:
Thank you for review. This function sets urgent bit. In some graphical shells, these workspaces are highlighted (with urgent bits). will suit  you, if I wrap the code under the condition that unread count is greater than 0?

Mephistophiles:
upd: 
```diff
diff --git a/Telegram/SourceFiles/platform/linux/main_window_linux.cpp b/Telegram/SourceFiles/platform/linux/main_window_linux.cpp
index 536969165..8361b0539 100644
--- a/Telegram/SourceFiles/platform/linux/main_window_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/main_window_linux.cpp
@@ -344,6 +344,13 @@ void MainWindow::updateIconCounters() {
 	updateWindowIcon();
 
 	const auto counter = Core::App().unreadBadge();
+	const auto muted = Core::App().unreadBadgeMuted();
+
+	if (counter > 0 && !muted) {
+		foreach(QWidget *widget, qApp->topLevelWidgets()) {
+			QApplication::alert(widget);
+		}
+	}
 
 	if (useUnityCount) {
 		QVariantMap dbusUnityProperties;
@@ -386,9 +393,6 @@ void MainWindow::updateIconCounters() {
 			QByteArray path = QFile::encodeName(iconFile.absoluteFilePath());
 			icon = QIcon(path.constData());
 		} else {
-			const auto counter = Core::App().unreadBadge();
-			const auto muted = Core::App().unreadBadgeMuted();
-
 			auto &bg = (muted ? st::trayCounterBgMute : st::trayCounterBg);
 			auto &fg = st::trayCounterFg;
 			icon.addPixmap(App::pixmapFromImageInPlace(iconWithCounter(16, counter, bg, fg, true)));

```

xvitaly:
Flashing of taskbar icon is not good. This must be an option.

magnus-gross:
I agree with @xvitaly, flashing every time you get a message would be very annoying. Most programs only use the flashing, if something really requires your attention.

Isn't the unread messages badge already a good enough indicator, that you got a new message?
img

Mephistophiles:
Ok, i will close this pull request. Thanks for code review!

john-preston:
@Mephistophiles Perhaps I'll add that together with a setting later.

john-preston:
@Mephistophiles There is a place in code for that, where it is done on macOS and Windows:

https://github.com/telegramdesktop/tdesktop/search?utf8=%E2%9C%93&q=FlashBounce&type=Code

It is empty on Linux, but studying the Qt docs it looks like QApplication::alert() can be used instead of all platform-specific code. But I guess a setting should be added as well.

Rasie1:
@magnus-gross it's not enough. I think, urgency flag might be the only clean way to check if there are unread messages. Some of other apps use it for indicating new messages (for example, Skype).
Some desktop environments (like, i3wm), make use of it to do fancy things like "jump to urgent window" shortcut, or highlighting urgent workspaces.
I would love to have this option

magnus-gross:
Yeah I agree that it might be useful, but it should be an optional feature, as I don't want Telegram to flash every minute or so.

taketwo:
> Perhaps I'll add that together with a setting later.

@john-preston was there any progress on this one?

PR #5686, opened 2019-02-12
Improved "Create link" box title display.

23rd:
This PR slightly improves the display of the "Create link" box title.
When url in box already exists title changes to "Edit link".
Seems like the `lng_formatting_link_edit_title` key was created for this, but it was forgotten to insert in the code.

image  image


john-preston:
The wrong phrase even made it to the translations platform :) Thanks!

PR #5694, opened 2019-02-14
Fixed Ctrl + F shortcut for full screen toggle in media view.

23rd:
This PR fixes a non-working Ctrl + F shortcut in the media view, which is covered by a `Command::Search` shortcut.

### DIFF
--- a/Telegram/SourceFiles/media/view/media_view_overlay_widget.cpp
+++ b/Telegram/SourceFiles/media/view/media_view_overlay_widget.cpp
@@ -2487,9 +2487,9 @@ void OverlayWidget::paintThemePreview(Painter &p, QRect clip) {
 void OverlayWidget::keyPressEvent(QKeyEvent *e) {
 	const auto ctrl = e->modifiers().testFlag(Qt::ControlModifier);
 	if (_clipController) {
-		auto toggle1 = (e->key() == Qt::Key_F && ctrl);
-		auto toggle2 = (e->key() == Qt::Key_Enter || e->key() == Qt::Key_Return) && (e->modifiers().testFlag(Qt::AltModifier) || ctrl);
-		if (toggle1 || toggle2) {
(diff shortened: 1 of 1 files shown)


Telegram/SourceFiles/media/view/media_view_overlay_widget.cpp
```diff
@@ -2487,9 +2487,9 @@ void OverlayWidget::paintThemePreview(Painter &p, QRect clip) {
 void OverlayWidget::keyPressEvent(QKeyEvent *e) {
 	const auto ctrl = e->modifiers().testFlag(Qt::ControlModifier);
 	if (_clipController) {
-		auto toggle1 = (e->key() == Qt::Key_F && ctrl);
-		auto toggle2 = (e->key() == Qt::Key_Enter || e->key() == Qt::Key_Return) && (e->modifiers().testFlag(Qt::AltModifier) || ctrl);
-		if (toggle1 || toggle2) {
+		// Ctrl + F for full screen toggle is in eventFilter().
```
john-preston: Is it correct that if you change Search shortcut from Ctrl+F to something else and have no shortcuts bound to Ctrl+F then it'll stop working because no QEvent::ShortcutOverride will be sent? Shorcuts can be overriden in the shorcuts-custom.json file.

23rd: No, this is not correct.
If you replace `Qt::Key_F`, for example, with `Qt::Key_N`, which has no binds in the code and in the json, it will be work.

By the way, as I could test, you cant override `Command::Search` via `shorcuts-custom.json`. 
I can add in this json next lines.
```
    {
        "command": "search",
        "keys": "ctrl+b"
    }
```

And then `Command::Search` will be work and with `Ctrl + F`, and with `Ctrl + B`.

Issue #5550, opened 2019-01-02
Can't send polls as reply to other messages


delivrance:
### Steps to reproduce
1. Double click onto a message to enable the reply ui
2. Create and send a poll via the three dots menu
3. Hi Preston

### Expected behaviour
Poll messages should mention the replied message, like in this picture:
![image](https://user-images.githubusercontent.com/14043624/50612546-99f2db00-0eda-11e9-9c18-50cbb229a42b.png)

### Actual behaviour
They don't

### Configuration
**Operating system:** macOS Mojave 10.14.2

**Version of Telegram Desktop:** 1.5.4

**Used theme**: Stock Matrix

<details><summary><b>Logs</b>:</summary>
Hi Viktor
</details>

PR #5701, opened 2019-02-17
Fixed sending polls as reply to other messages.

23rd:
This PR fixes [this one](https://github.com/telegramdesktop/tdesktop/issues/5550).
But I'm not completely sure about the way of my solution.
Particularly not sure about using of the `MainWidget` singleton method to pass the `replyToId` value, but it is easier than passing the value through 2 child classes.
Reading of the local draft is needed when a user is in `Channel/Group info` and wants to create the poll from there. 
Generally it works and repeats patterns of the Android client's behavior.

### DIFF
--- a/Telegram/SourceFiles/mainwidget.cpp
+++ b/Telegram/SourceFiles/mainwidget.cpp
@@ -1118,6 +1118,15 @@ bool MainWidget::historyInSelectionMode() const {
 	return _history->inSelectionMode();
 }
 
+MsgId MainWidget::currentReplyToIdFor(not_null<History*> history) const {
+	if (_history->history() == history) {
+		return _history->replyToId();
+	} else if (const auto localDraft = history->localDraft()) {
+		return localDraft->msgId;
+	}
+	return 0;
+}
+
 void MainWidget::sendBotCommand(PeerData *peer, UserData *bot, const QString &cmd, MsgId replyTo) {
 	_history->sendBotCommand(peer, bot, cmd, replyTo);
 }
--- a/Telegram/SourceFiles/mainwidget.h
+++ b/Telegram/SourceFiles/mainwidget.h
@@ -217,6 +217,8 @@ public:
 	TimeMs highlightStartTime(not_null<const HistoryItem*> item) const;
 	bool historyInSelectionMode() const;
 
+	MsgId currentReplyToIdFor(not_null<History*> history) const;
+
 	void sendBotCommand(PeerData *peer, UserData *bot, const QString &cmd, MsgId replyTo);
 	void hideSingleUseKeyboard(PeerData *peer, MsgId replyTo);
 	bool insertBotCommand(const QString &cmd);
--- a/Telegram/SourceFiles/window/window_peer_menu.cpp
+++ b/Telegram/SourceFiles/window/window_peer_menu.cpp
@@ -35,6 +35,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 #include "data/data_poll.h"
 #include "data/data_channel.h"
 #include "data/data_chat.h"
+#include "data/data_drafts.h"
 #include "data/data_user.h"
 #include "dialogs/dialogs_key.h"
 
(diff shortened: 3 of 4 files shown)


Telegram/SourceFiles/mainwidget.cpp
```diff
@@ -1118,6 +1118,17 @@ bool MainWidget::historyInSelectionMode() const {
 	return _history->inSelectionMode();
 }
 
+void MainWidget::historyReplyToId(ApiWrap::SendOptions &options) {
```
john-preston: At first glance it looks like `MsgId historyReplyToId() const;` is better. And just get `_history->replyToId()`, moving all the logic handling `clearDraft` to the caller's side.

Telegram/SourceFiles/mainwidget.h
```diff
@@ -217,6 +218,8 @@ class MainWidget
 	TimeMs highlightStartTime(not_null<const HistoryItem*> item) const;
 	bool historyInSelectionMode() const;
 
+	void historyReplyToId(ApiWrap::SendOptions &options);
```
john-preston: If the signature becomes `MsgId historyReplyToId() const` then no need to include `apiwrap.h`.

Telegram/SourceFiles/window/window_peer_menu.cpp
```diff
@@ -636,7 +636,9 @@ void PeerMenuCreatePoll(not_null<PeerData*> peer) {
 		if (std::exchange(*lock, true)) {
 			return;
 		}
-		const auto options = ApiWrap::SendOptions(peer->owner().history(peer));
+		auto options = ApiWrap::SendOptions(peer->owner().history(peer));
+		App::main()->historyReplyToId(options);
```
john-preston: And here it will first get `options.replyTo = App::main()->historyReplyToId();` and then `if (const auto localDraft = options.history->localDraft()) { // fill clearDraft and replyTo if it's still zero`.

john-preston:
> Reading of the local draft is needed when a user is in `Channel/Group info` and wants to create the poll from there.

This is a bit more complicated then I suggested in my review. In that case you should ignore `_history->replyToId()` if not that history is opened. It may be `MainWidget::currentReplyToIdFor(not_null<History*> history)` which will first check if this history is opened and only then return `replyToId()` from `_history`, otherwise return `0`.

23rd:
> `MainWidget::currentReplyToIdFor(not_null<History*> history)` which will first check if this history is opened and only then return `replyToId()` from `_history`, otherwise return `0`.

But if the history is not opened then `replyToId()` will return `0` and we can just use it.

john-preston:
@23rd You can have one group with one replyTo opened and view info of some other group with completely different local draft and replyTo in it.

23rd:
@john-preston Code is updated.

Here is another problem: if user replies a message in `chat1` and sends a poll to `chat2`, then the reply will be dropped. But I think this issue is not related to this PR, issue is reproduced in current master too.

Telegram/SourceFiles/window/window_peer_menu.cpp
```diff
@@ -636,7 +637,17 @@ void PeerMenuCreatePoll(not_null<PeerData*> peer) {
 		if (std::exchange(*lock, true)) {
 			return;
 		}
-		const auto options = ApiWrap::SendOptions(peer->owner().history(peer));
+		auto options = ApiWrap::SendOptions(peer->owner().history(peer));
+		if (App::main()->currentReplyToIdFor(options.history)) {
```
john-preston: That way here we'll get just `if (const auto id = App::main()->currentReplyToIdFor(options.history)) { options.replyTo = id; }` and then `if (const auto localDraft = options.history->localDraft()) { options.clearDraft = localDraft->textWithTags.text.isEmpty(); }`, so `replyTo` and `clearDraft` will be filled independently.

john-preston:
@23rd About the replyTo clearing in another chat, it should be fixed in `HistoryWidget`: it should ignore all `sendActions` that don't belong to its current history. I'll fix that in a separate commit. Thanks for the note.

23rd:
Okay, now we have a single method in `MainWidget` but two `if (const auto localDraft = options.history->localDraft()) {` checks and an extra include. =)

john-preston:
👍

23rd:
Thanks a lot for supporting with code.

john-preston:
Thanks for your efforts!

Issue #5695, opened 2019-02-14
auto-lock windows,miss recording voice message


muhammadzadeh:
### Steps to reproduce
1. auto lock is enabled and scheduled 5min
2. recording voice message with a duration greater than 5min
3. screen will be locked
4. voice message lost
5.and I should record again😢

### Configuration
**Operating system: winndows 10**

**Version of Telegram Desktop: 1.5.15**

**Used theme: default theme, night mode**

john-preston:
@muhammadzadeh Thanks for your report!

PR #5715, opened 2019-02-20
Fixed auto-lock while recording voice message.

23rd:
This PR fixes [this one](https://github.com/telegramdesktop/tdesktop/issues/5695).
If you don't mind, I also added some `const` to vars within `checkAutoLock()`.
If this is not good for single commit, I'll remove it.

### DIFF
--- a/Telegram/SourceFiles/auth_session.cpp
+++ b/Telegram/SourceFiles/auth_session.cpp
@@ -483,11 +483,9 @@ void AuthSession::checkAutoLock() {
 	}
 
 	Core::App().checkLocalTime();
-	auto now = crl::now();
-	auto shouldLockInMs = Global::AutoLock() * 1000LL;
-	auto idleForMs = psIdleTime();
-	auto notPlayingVideoForMs = now - settings().lastTimeVideoPlayedAt();
-	auto checkTimeMs = qMin(idleForMs, notPlayingVideoForMs);
+	const auto now = crl::now();
+	const auto shouldLockInMs = Global::AutoLock() * 1000LL;
+	const auto checkTimeMs = now - Core::App().lastNonIdleTime();
 	if (checkTimeMs >= shouldLockInMs || (_shouldLockAt > 0 && now > _shouldLockAt + kAutoLockTimeoutLateMs)) {
 		Core::App().lockByPasscode();
 	} else {
--- a/Telegram/SourceFiles/mainwidget.cpp
+++ b/Telegram/SourceFiles/mainwidget.cpp
@@ -3767,7 +3767,7 @@ void MainWidget::updateOnline(bool gotOtherOffline) {
 	bool isOnline = !App::quitting() && App::wnd()->isActive();
 	int updateIn = Global::OnlineUpdatePeriod();
 	if (isOnline) {
-		auto idle = psIdleTime();
+		const auto idle = crl::now() - Platform::LastUserInputTime();
 		if (idle >= Global::OfflineIdleTimeout()) {
 			isOnline = false;
 			if (!_isIdle) {
@@ -3883,7 +3883,7 @@ void MainWidget::writeDrafts(History *history) {
 }
 
 void MainWidget::checkIdleFinish() {
-	if (psIdleTime() < Global::OfflineIdleTimeout()) {
+	if (crl::now() - Platform::LastUserInputTime() < Global::OfflineIdleTimeout()) {
 		_idleFinishTimer.cancel();
 		_isIdle = false;
 		updateOnline();
--- a/Telegram/SourceFiles/platform/linux/specific_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/specific_linux.cpp
@@ -207,10 +207,6 @@ bool psIdleSupported() {
 	return false;
 }
 
-crl::time psIdleTime() {
-	return crl::now() - _lastUserAction;
-}
-
 void psActivateProcess(uint64 pid) {
 //	objc_activateProgram();
 }
@@ -447,6 +443,10 @@ bool OpenSystemSettings(SystemSettingsType type) {
 		}) != end(options);
 	}
 	return true;
+}
+
+crl::time LastUserInputTime() {
+	return _lastUserAction;
 }
 
 namespace ThirdParty {
--- a/Telegram/SourceFiles/platform/linux/specific_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/specific_linux.cpp
@@ -193,16 +193,6 @@ void psDeleteDir(const QString &dir) {
 	_removeDirectory(dir);
 }
 
-namespace {
-
-auto _lastUserAction = 0LL;
-
-} // namespace
-
-void psUserActionDone() {
-	_lastUserAction = crl::now();
-}
-
 bool psIdleSupported() {
 	return false;
 }
@@ -446,7 +436,7 @@ bool OpenSystemSettings(SystemSettingsType type) {
 }
 
 crl::time LastUserInputTime() {
-	return _lastUserAction;
+	return 0LL;
 }
 
 namespace ThirdParty {
--- a/Telegram/SourceFiles/platform/mac/specific_mac.mm
+++ b/Telegram/SourceFiles/platform/mac/specific_mac.mm
@@ -90,16 +90,6 @@ void psDeleteDir(const QString &dir) {
 	objc_deleteDir(dir);
 }
 
-namespace {
-
-auto _lastUserAction = 0LL;
-
-} // namespace
-
-void psUserActionDone() {
-	_lastUserAction = crl::now();
-}
-
 bool psIdleSupported() {
 	return objc_idleSupported();
 }
@@ -267,7 +257,7 @@ bool OpenSystemSettings(SystemSettingsType type) {
 
 crl::time LastUserInputTime() {
 	auto idleTime = 0LL;
-	return objc_idleTime(idleTime) ? (crl::now() - crl::time(idleTime)) : _lastUserAction;
+	return objc_idleTime(idleTime) ? (crl::now() - crl::time(idleTime)) : 0LL;
 }
 
 } // namespace Platform
--- a/Telegram/SourceFiles/platform/mac/specific_mac.mm
+++ b/Telegram/SourceFiles/platform/mac/specific_mac.mm
@@ -104,11 +104,6 @@ bool psIdleSupported() {
 	return objc_idleSupported();
 }
 
-crl::time psIdleTime() {
-	auto idleTime = 0LL;
-	return objc_idleTime(idleTime) ? idleTime : (crl::now() - _lastUserAction);
-}
-
 QStringList psInitLogs() {
 	return _initLogs;
 }
@@ -268,6 +263,11 @@ bool OpenSystemSettings(SystemSettingsType type) {
 		break;
 	}
 	return true;
+}
+
+crl::time LastUserInputTime() {
+	auto idleTime = 0LL;
+	return objc_idleTime(idleTime) ? (crl::now() - crl::time(idleTime)) : _lastUserAction;
 }
 
 } // namespace Platform
--- a/Telegram/SourceFiles/platform/win/specific_win.cpp
+++ b/Telegram/SourceFiles/platform/win/specific_win.cpp
@@ -128,17 +128,6 @@ namespace {
 	}
 }
 
-namespace {
-
-crl::time _lastUserAction = 0;
-
-} // namespace
-
-void psUserActionDone() {
-	_lastUserAction = crl::now();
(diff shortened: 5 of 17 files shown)


23rd:
I have realized that it's better to use `std::min()` instead of `qMin()`.

john-preston:
Thanks for the fix. I think this can be improved in terms of code meaning. All the autolock delays are due to the computer not being idle. So all those are `idleFor` really.

I suggest changing `psIdleTime` to `Platform::LastUserInputTime` and making one `crl::time` value `_lastNonIdleTime` in `Application` (instead of `lastTime*At` variables in AuthSessionSettings) + add `[[nodiscard]] crl::time lastNonIdleTime() const` and `void updateNonIdle();` in `Application`, the first will return `std::max(Platform::LastUserInputTime(), _lastNonIdleTime)`, the second will set `_lastNonIdleTime` to `crl::now()`.

Because we really are not interested when was the last update in video playback or voice recording. And also they really are not current user settings in any sense, but the last non idle time is something about the whole application (and the platform-specific last user input time is taken even from the whole system, not single application, where there is platform API to do that).

That way both `Auth().settings().setLastTime*At(crl::now())` will become just `Core::App().updateNonIdle()`.

john-preston:
Maybe the same logic can be applied to other places where `psIdleTime` is used right now. For example where the notifications are delayed on the screen until the user input happens - it will be good if they won't be delayed in case video is watched or voice message is recorded, so `Core::App().lastNonIdleTime()` can be used there too.

23rd:
@john-preston 

Thanks for replies.
Unfortunately, the `std::max(Platform::LastUserInputTime(), _lastNonIdleTime)` does not work in the way that you suggested. I replaced it with `return _lastNonIdleTime;` and it came to work as it should.

>For example where the notifications are delayed on the screen until the user input happens - it will be good if they won't be delayed in case video is watched or voice message is recorded, so `Core::App().lastNonIdleTime()` can be used there too.

Yeah, I was surprised to find that checking of the last user input properly works only for Windows. And it repeats the code of the `Platform::LastUserInputTime()` from `specific_win.h`, so I replaced it with a hope that it will work for all platforms. I splited it into 2 commits, because notification delay is a little bit another story.

Count of changed files is huge now. =\

john-preston:
Why it doesn't work? I guess, 'LastUserInputTime' in platform-specific code is not on the same scale / origin, as crl::now()? But it could be fixed by LastUserInputTime returning `crl::now() - psIdleTime()` for the old implementation of `psIdleTime`.

john-preston:
Last time I checked it worked properly on macOS as well. Maybe they've changed something in the API since then, I'm not sure.

john-preston:
@23rd UPD I see now. As I understand you've renamed `psIdleTime` to `LastUserInputTime`, but that was not what I've suggested. `LastUserInputTime` was supposed to return the last time there was an input event, so in terms of old code `LastUserInputTime() { return crl::now() - psIdleTime(); }`, but removing `psIdleTime` and leaving the related code just in `LastUserInputTime` and using it in `Application::lastNonIdleTime` - max of the last time user input was detected by platform-specific APIs, last time a video was played or last time a voice message was recorded.

Telegram/SourceFiles/platform/win/specific_win.cpp
```diff
@@ -350,6 +344,12 @@ QString CurrentExecutablePath(int argc, char *argv[]) {
 	return QString();
 }
 
+crl::time LastUserInputTime() {
+	LASTINPUTINFO lii;
+	lii.cbSize = sizeof(LASTINPUTINFO);
+	return GetLastInputInfo(&lii) ? (GetTickCount() - lii.dwTime) : (crl::now() - _lastUserAction);
```
john-preston: So, this one will be `GetLastInputInfo(&lii) ? (crl::now() + lii.dwTime - GetTickCount()) : _lastUserAction`.

Telegram/SourceFiles/mainwidget.cpp
```diff
@@ -3767,7 +3767,7 @@ void MainWidget::updateOnline(bool gotOtherOffline) {
 	bool isOnline = !App::quitting() && App::wnd()->isActive();
 	int updateIn = Global::OnlineUpdatePeriod();
 	if (isOnline) {
-		auto idle = psIdleTime();
+		const auto idle = Platform::LastUserInputTime();
```
john-preston: Here it'll be `idle = crl::now() - Platform::LastUserInputTime()`.

Telegram/SourceFiles/mainwidget.cpp
```diff
@@ -3883,7 +3883,7 @@ void MainWidget::writeDrafts(History *history) {
 }
 
 void MainWidget::checkIdleFinish() {
-	if (psIdleTime() < Global::OfflineIdleTimeout()) {
+	if (Platform::LastUserInputTime() < Global::OfflineIdleTimeout()) {
```
john-preston: Here it'll be `crl::now() - Platform::LastUserInputTime() < Global::OfflineIdleTimeout()`.

Telegram/SourceFiles/platform/linux/specific_linux.cpp
```diff
@@ -449,6 +445,10 @@ bool OpenSystemSettings(SystemSettingsType type) {
 	return true;
 }
 
+crl::time LastUserInputTime() {
+	return crl::now() - _lastUserAction;
```
john-preston: Here it'll be just `_lastUserAction`.

Telegram/SourceFiles/platform/mac/specific_mac.mm
```diff
@@ -104,11 +104,6 @@ bool psIdleSupported() {
 	return objc_idleSupported();
 }
 
-crl::time psIdleTime() {
-	auto idleTime = 0LL;
-	return objc_idleTime(idleTime) ? idleTime : (crl::now() - _lastUserAction);
```
john-preston: Here it'll be `objc_idleTime(idleTime) ? (crl::now() - crl::time(idleTime)) : _lastUserAction`.

Telegram/SourceFiles/platform/mac/specific_mac.mm
```diff
@@ -270,6 +265,11 @@ bool OpenSystemSettings(SystemSettingsType type) {
 	return true;
 }
 
+crl::time LastUserInputTime() {
+	auto idleTime = 0LL;
+	return objc_idleTime(idleTime) ? idleTime : (crl::now() - _lastUserAction);
```
john-preston: Here it'll be `objc_idleTime(idleTime) ? (crl::now() - crl::time(idleTime)) : _lastUserAction`.

Telegram/SourceFiles/window/notifications_manager_default.cpp
```diff
@@ -588,14 +586,8 @@ void Notification::prepareActionsCache() {
...
 
-	auto wasUserInput = true; // TODO
-#ifdef Q_OS_WIN
-	LASTINPUTINFO lii;
-	lii.cbSize = sizeof(LASTINPUTINFO);
-	BOOL res = GetLastInputInfo(&lii);
-	wasUserInput = (!res || lii.dwTime >= _started);
-#endif // Q_OS_WIN
-	if (wasUserInput) {
+	if ((crl::now() - Platform::LastUserInputTime() > _started)
```
john-preston: Here it'll be just `Core::App().lastNonIdleTime() > _started`.

Telegram/SourceFiles/auth_session.cpp
```diff
@@ -483,11 +483,9 @@ void AuthSession::checkAutoLock() {
...
 
 	Core::App().checkLocalTime();
-	auto now = crl::now();
-	auto shouldLockInMs = Global::AutoLock() * 1000LL;
-	auto idleForMs = psIdleTime();
-	auto notPlayingVideoForMs = now - settings().lastTimeVideoPlayedAt();
-	auto checkTimeMs = qMin(idleForMs, notPlayingVideoForMs);
+	const auto now = crl::now();
+	const auto shouldLockInMs = Global::AutoLock() * 1000LL;
+	const auto checkTimeMs = std::min(Platform::LastUserInputTime(), now - Core::App().lastNonIdleTime());
```
john-preston: Here it'll be just `now - Core::App().lastNonIdleTime()`.

john-preston:
@23rd UPD I see now about the `only on Windows` - that's because on Linux I didn't find such API and on macOS (at least starting with OS X 10.8) only native notifications are used, so the code of custom notifications that are kept on the screen while there is no user input is never used anyway. But it is a good idea to use the same methods as in other places, that's good.

23rd:
Oh...
To be honest, I don't see much advantage, we do subtraction in one place, but add in another.

23rd:
@john-preston 
The code updated.

Sorry for misunderstanding about `psIdleTime` and `LastUserInputTime`.

john-preston:
@23rd 

> we do subtraction of crl::now() in one place, but add it in another.

Yes, this is not really a simplification in the amount of operations / code, but a more consistent naming - one platform-specific time value - from which point of time we think there was no user attention to the app (last user input time) + one in-app value - with the same idea (last time we've updated voice message recording or video frame was shown). And one application-wide method, that returns max of those two values.

Telegram/SourceFiles/window/notifications_manager_default.cpp
```diff
@@ -588,14 +586,7 @@ void Notification::prepareActionsCache() {
...
 
-	auto wasUserInput = true; // TODO
-#ifdef Q_OS_WIN
-	LASTINPUTINFO lii;
-	lii.cbSize = sizeof(LASTINPUTINFO);
-	BOOL res = GetLastInputInfo(&lii);
-	wasUserInput = (!res || lii.dwTime >= _started);
-#endif // Q_OS_WIN
-	if (wasUserInput) {
+	if ((Core::App().lastNonIdleTime() > _started)) {
```
john-preston: No need for double parentheses here? Just `if (Core::App().lastNonIdleTime() > _started) {`?

john-preston:
No, not the last one. Now I see one more way to simplify this, sorry.

Right now all three platform-specific code paths track `_lastUserAction` time in `psUserActionDone` call. Really this call can be made no-op everywhere except Windows version (where it does something else too) and we should just call `Core::App().updateNonIdle()` where we now call `psUserActionDone()`.

That way `_lastUserAction` is not needed at all, it will be tracked inside `_lastNonIdleTime` together with video watching and voice message recording. And on Linux `Platform::LastUserInputTime()` will simply return zero, because there is no API for that (other systems will return value if they could get it and zero otherwise).

john-preston:
I can be `Core::App().userActionDone()` that'll call both `psUserActionDone()` and `updateNonIdle()`.

It is possible that `psUserActionDone` is not needed at all, even on Windows, but right now I'm not sure, so I'd leave it as it is.

23rd:
In that case, probably should not hurry to merge this PR, I would test it a little more.

https://github.com/telegramdesktop/tdesktop/blob/e174025a92c67f5a5c6545fbcda6e44dbd7f3348/Telegram/SourceFiles/platform/win/specific_win.cpp#L137-L140

As I can see, this call is just to be sure that the sound of notifications would not be skipped after the log off or the lock.

john-preston:
Yes, it should work even without that, but to be sure that we are in the right state in case of any user input we force us to be not in session-logged-out mode.

23rd:
@john-preston 

So, I added few `LOG()` in `case WM_WTSSESSION_CHANGE: {` and tested lock in Windows 7, Windows 8.1 and Windows 10. Everywhere handler works well, it is called even twice.

If we still want to be sure that we a not logged off, I suggest to replace
```
if (wParam == WTS_SESSION_LOGOFF || wParam == WTS_SESSION_LOCK) {
	setSessionLoggedOff(true);
} else if (wParam == WTS_SESSION_LOGON || wParam == WTS_SESSION_UNLOCK) {
	setSessionLoggedOff(false);
}
```
with
```
setSessionLoggedOff(wParam == WTS_SESSION_LOGOFF || wParam == WTS_SESSION_LOCK);
```

But I see no reason not to trust windows event handler.

john-preston:
@23rd Yes, perhaps we can check if that works fine.

john-preston:
@23rd I mean, try without handling user input for that and see if there'll be complaints.

23rd:
I don't get you.
In the latest commit I removed `psUserActionDone()` with `EventFilter::getInstance()->setSessionLoggedOff(false);` and a little tested it.
In Windows with locked user the sound of notifications was skipped, but when I loged in the sound returned.

john-preston:
@23rd Yes, I've meant exactly that - try that version (that you already pushed, I didn't see that, was replying to your comment) in production and see if there are any complaints. I think there won't be any and everything is fine. Thanks!

PR #5725, opened 2019-02-23
Fixed extra space in a name of the post author after forwarding message.

23rd:
This PR fixes a pretty rare improper display of the post author.
If user has no the last name, his forwarded messages locally will be display with extra space in the author section.
So I unified the creation of the `messagePostAuthor` string in the `ApiWrap` class.

image


john-preston:
Thanks!

Issue #5510, opened 2018-12-23
[Feature Request] Show channel's post's author when possible on mouse over


ArmeF97:
### Steps to reproduce
1.  Mouse over the name

### Expected behaviour
Display full name

### Actual behaviour
Don't display full name

### Configuration
**Operating system:** Windows 10

**Version of Telegram Desktop:** 1.5.3

**Used theme**: Dark

![immagine](https://user-images.githubusercontent.com/6240983/50384844-1e6b7f80-06cb-11e9-9e73-bea008915e19.png)

PR #5727, opened 2019-02-24
Added author of channel post in tooltip.

23rd:
When the author name of the channel post is too long, it is possible to see the author from the message header only when you forward this post to another chat.
So PR adds the author of the channel post in the tooltip when it is cut off.
Also this PR closes [this one](https://github.com/telegramdesktop/tdesktop/issues/5510).

image

image

### DIFF
--- a/Telegram/SourceFiles/history/history_item_components.cpp
+++ b/Telegram/SourceFiles/history/history_item_components.cpp
@@ -50,11 +50,12 @@ void HistoryMessageVia::resize(int32 availw) const {
 }
 
 void HistoryMessageSigned::refresh(const QString &date) {
-	auto time = qsl(", ") + date;
 	auto name = author;
-	auto timew = st::msgDateFont->width(time);
-	auto namew = st::msgDateFont->width(name);
-	if (timew + namew > st::maxSignatureSize) {
+	const auto time = qsl(", ") + date;
+	const auto timew = st::msgDateFont->width(time);
+	const auto namew = st::msgDateFont->width(name);
+	isElided = (timew + namew > st::maxSignatureSize);
+	if (isElided) {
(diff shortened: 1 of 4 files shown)


Telegram/SourceFiles/history/history_item_components.h
```diff
@@ -37,6 +37,7 @@ struct HistoryMessageSigned : public RuntimeComponent<HistoryMessageSigned, Hist
 	void refresh(const QString &date);
 	int maxWidth() const;
```
john-preston: Please provide default value in-class for all plain types (= false;)

Telegram/SourceFiles/history/history_item_components.cpp
```diff
@@ -50,11 +50,11 @@ void HistoryMessageVia::resize(int32 availw) const {
...
 void HistoryMessageSigned::refresh(const QString &date) {
-	auto time = qsl(", ") + date;
 	auto name = author;
-	auto timew = st::msgDateFont->width(time);
-	auto namew = st::msgDateFont->width(name);
-	if (timew + namew > st::maxSignatureSize) {
+	const auto time = qsl(", ") + date;
+	const auto timew = st::msgDateFont->width(time);
+	const auto namew = st::msgDateFont->width(name);
+	if (isElided = timew + namew > st::maxSignatureSize) {
```
john-preston: I suppose for some compilers not to issue a warning here it is better to bring the assignment outside of the if clause. Other solution would be to add additional pair of parentheses, but in this case it looks like 

isElided = (timew + ... );
If (isElided) {..

Would be less clumsy.

23rd:
The code is updated.

john-preston:
Thanks!

Issue #3895, opened 2017-09-14
Search for messages from someone who has been left the group


PyBagheri:
Telegram desktop cannot search for messages from someone who has been left the group. It can just search for messages from someone that already is in the group. I think this a problem... no?

tdesktop-closer:
Hey there!

We're automatically closing this issue since there was no activity in this issue since 388 days ago. We therefore assume that the user has lost interest or resolved the problem on their own. Closed issues that remain inactive for a long period may get automatically locked.

Don't worry though; if this is in error, let us know with a comment and we'll be happy to reopen the issue.

Thanks!

<sub>(Please note that this is an [automated](https://github.com/auchri/github_auto_closer) comment.)</sub>

Remu-rin:
Please reopen. Is it possible to implement searching messages from users that left the group via typing `from:username` manually? Or is it api issue?

Aokromes:
f979df3dfe2ac46e47fe60d2fa9acb0d77e4dfb6

PR #5728, opened 2019-02-24
Allow searching for messages by a user who has left a chat, fixes #5667

### LINKED ISSUE #5667: Can't search for messages of left users
### Steps to reproduce
1. User1 joins a group.
2. User1 posts some messages.
3. User1 leaves the group.
4. User2 clicks the search icon in the group.
5. User2 clicks the silhouette icon in the search field.

### Expected behaviour
All users that have posted non-deleted messages in the group are listed.

### Actual behaviour
Only current members of the group are listed.

### Configuration
**Operating system:** Debian GNU/Linux 10

**Version of Telegram Desktop:** 1.5.4

**Used theme**: Classic

TheWug:
fixes #5667 

Change `Dialog::SearchFromController` to inherit not from`ParticipantsBoxController`, which only shows the users in the group, but instead from`AddSpecialBoxController` which shows the group's users but falls back on global search as well.

I know you don't generally accept pull requests for features and calling this a bug is very borderline but I figured I'd post it anyway.

### DIFF
--- a/Telegram/SourceFiles/boxes/peers/add_participants_box.cpp
+++ b/Telegram/SourceFiles/boxes/peers/add_participants_box.cpp
@@ -281,6 +281,8 @@ void AddSpecialBoxController::prepare() {
 	delegate()->peerListSetSearchMode(PeerListSearchMode::Enabled);
 	const auto title = [&] {
 		switch (_role) {
+		case Role::Members:
+			return langFactory(lng_profile_participants_section);
 		case Role::Admins:
 			return langFactory(lng_channel_add_admin);
 		case Role::Restricted:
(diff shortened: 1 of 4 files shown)


Aokromes:
@TheWug if you change #5667 thing to closes #5667 merging this pr will close that ticket :)

Telegram/SourceFiles/boxes/peers/add_participants_box.cpp
```diff
@@ -412,6 +414,7 @@ void AddSpecialBoxController::loadMoreRows() {
 void AddSpecialBoxController::rowClicked(not_null<PeerListRow*> row) {
 	auto user = row->peer()->asUser();
 	switch (_role) {
+	case Role::Members: return;
```
john-preston: I suggest placing ‘Unexpected(“Role in AddSpecialBoxConyroller::rowClicked”);’, because you should never get here if there is no bugs.

Aokromes:
i suggest you to add closes https://github.com/telegramdesktop/tdesktop/issues/3895

Hobbix:
Still not fixed?

rezaei:
[Feature Request] grouping channels and groups and chats separately [make diffrent attributes such as color or category] #6420 
—— we need to have the ability to manage and categorize groups and channels and chats, because there are many channels and groups and chats that make me confuse and also make telegram unhelpful, I think its better for me if I can add these elements in the desired categorize with desired names like 'education' 'industry' 'family' 'friends' 'colleagues' 'sport' and so on ...

Issue #3145, opened 2017-03-23
infinite time sending zero-width-space (\u200B)


iansowinski:
### Steps to reproduce
1. in mac terminal type ```printf "\xE2\x80\x8B" | pbcopy```
2. paste to chat on telegram
3. send

Often [zero-width-space](http://www.fileformat.info/info/unicode/char/200B/index.htm) breaks everything, let's make telegram resistant to it!

### Expected behaviour

It sholud bahave the same as in case of sending normal space ass message.

### Actual behaviour

It tries to send the message but ends up beeing freezed message

### Configuration
**Operating system:**

OS X 10.10

**Version of Telegram Desktop:**

1.0.14

**Used theme**:

default

<details><summary>**Logs**:</summary>
Insert logs here (if necessary)

</details>

stek29:
Btw, API thinks that that message is empty (if you try to send it via webogram you get MESSAGE_EMPTY error)
So MainWidget::sendMessageFail should handle MESSAGE_EMPTY, just like HistoryWidget's saveEditMsgFail

PR #5729, opened 2019-02-24
Added handler for "MESSAGE_EMPTY" error to delete a local message.

23rd:
This PR is related to [this issue](https://github.com/telegramdesktop/tdesktop/issues/3145).
In another sending methods the zero width space is handled fine.
Perhaps the message box is unnecessary.

image


john-preston:
Really there are some checks for sending an empty message (when you enter several spaces or smth). Perhaps just add the characters to the list of checked in that case?

23rd:
I thought this solution would be more generic than checking a single character locally.
But most likely you are right, it will be better without unnecessary UI actions.

john-preston:
@23rd I see. But if we have the character it's better to add it, so the UK won't flash that message. Removing of the message can be kept, without the message box.

john-preston:
Thanks!

PR #5731, opened 2019-02-24
Added scroll animation in dialogs list when it scrolls to top.

23rd:
It's a bit sad that there is scroll animation in the dialogs, but not in the dialog list. =(
So I added an animation to the scroll when it goes to top (e.g. the user pins the dialog).
The animation of the scroll to top repeats patterns of the scroll animation's behavior in the dialogs.

image


john-preston:
Thanks!

PR #5743, opened 2019-03-02
Added forwarded info in tooltip of sticker.

23rd:
Unfortunately, it is almost impossible to get the information from whom the sticker was forwarded.
The only way to see that is to start forwarding this forwarded sticker to someone, then it will be seen who the original author.
image

So this PR added a little more information in the tooltip of forwarded stickers.
image


john-preston:
I try to avoid dynamic downcasts in the new code. I’m this case it looks legit to add a new virtual method to HistoryMedia, like ‘[[nodiscard]] virtual bool hidesForwardedInfo() { return false; }’ and override it in HistorySticker.

23rd:
Done.

john-preston:
Thanks!

PR #5748, opened 2019-03-03
 Added grouping of album files into a single notification.

23rd:
It is pretty annoying when someone sends you an album and you get several notifications of the same message.
So this PR combines the album messages in the single notification.
### The first commit.
image

image

In the code.
I have noticed that the album grouping algorithm is very similar to the forwarded messages grouping algorithm. So I crossed them. A process is almost the same. In addition to checking the author of forwarded messages, I added checking the `groupId()`.

Also.
The `forwardedItem` was renamed to `forwardedOrGroupedItem`.

Removed unused `lng_in_dlg_album` keys in `MediaPhoto`, seems like it will never be used.
https://github.com/telegramdesktop/tdesktop/blob/e70465c633e03bbda2847f6cc42f892e1674c17a/Telegram/SourceFiles/data/data_media_types.cpp#L300-L312

Removed unused `ms` in `System::showNext()`.
https://github.com/telegramdesktop/tdesktop/blob/e70465c633e03bbda2847f6cc42f892e1674c17a/Telegram/SourceFiles/window/notifications_manager.cpp#L313

Refactored a multiple `nextNotify = nullptr;` with the `continue;`. It allows not to use several `} else {` with the same single line.
https://github.com/telegramdesktop/tdesktop/blob/e70465c633e03bbda2847f6cc42f892e1674c17a/Telegram/SourceFiles/window/notifications_manager.cpp#L344-L352

### The second commit.
The second commit is pretty related to grouping of notifications.
If the user receives a lot of forwarded messages, notifications will be divided into random parts.
The second commit adds a small delay so that the Telegram has time to load all the forwarded messages and show the single notification. The album grouping is also subject to this issue, so I put fix to this PR too. I tried to cover the code with comments, because it is easy to get confused by timers.

image
image

### DIFF
--- a/Telegram/SourceFiles/window/notifications_manager.cpp
+++ b/Telegram/SourceFiles/window/notifications_manager.cpp
@@ -307,12 +307,14 @@ void System::showNext() {
 				_waitTimer.callOnce(next - ms);
 				break;
 			} else {
-				auto forwardedItem = notifyItem->Has<HistoryMessageForwarded>() ? notifyItem : nullptr; // forwarded notify grouping
-				auto forwardedCount = 1;
+				const auto isForwarded = notifyItem->Has<HistoryMessageForwarded>();
+				const auto isAlbum = notifyItem->groupId();
 
-				auto ms = crl::now();
-				auto history = notifyItem->history();
-				auto j = _whenMaps.find(history);
+				auto groupedItem = (isForwarded || isAlbum) ? notifyItem : nullptr; // forwarded and album notify grouping
+				auto forwardedCount = isForwarded ? 1 : 0;
+
+				const auto history = notifyItem->history();
+				const auto j = _whenMaps.find(history);
 				if (j == _whenMaps.cend()) {
 					history->clearNotifications();
 				} else {
@@ -323,9 +325,9 @@ void System::showNext() {
 							break;
 						}
 
-						j.value().remove((forwardedItem ? forwardedItem : notifyItem)->id);
+						j.value().remove((groupedItem ? groupedItem : notifyItem)->id);
 						do {
-							auto k = j.value().constFind(history->currentNotification()->id);
+							const auto k = j.value().constFind(history->currentNotification()->id);
 							if (k != j.value().cend()) {
 								nextNotify = history->currentNotification();
 								_waiters.insert(notifyHistory, Waiter(k.key(), k.value(), 0));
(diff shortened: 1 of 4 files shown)


Telegram/SourceFiles/window/notifications_manager.h
```diff
@@ -108,6 +110,9 @@ class System final : private base::Subscriber {
 
 	std::unique_ptr<Media::Audio::Track> _soundTrack;
 
+	int _lastForwardedCount = 0;
+	HistoryItem *_lastHistoryItem = nullptr;
```
john-preston: You should not save `HistoryItem*` unless you're really sure you need it (for example for performance reasons). That's because they may be deleted in any cycle of the event loop, so once you saved a pointer and left current method you have no idea when it'll be invalidated.

So in case you're not required to hold a pointer it is better to hold `FullMsgId _lastHistoryItemId;` and set it like `_lastHistoryItemId = forwardedOrGroupedItem->fullId();` and then read the item like `if (const auto item = App::histItemById(_lastHistoryItemId)) { ...`.

In case you're required to hold a pointer you should subscribe to `Auth().data().itemRemoved()` producer and invalidate your pointers when such event happens with your pointer value.

Telegram/SourceFiles/window/notifications_manager.cpp
```diff
@@ -307,12 +329,14 @@ void System::showNext() {
...
 			} else {
-				auto forwardedItem = notifyItem->Has<HistoryMessageForwarded>() ? notifyItem : nullptr; // forwarded notify grouping
-				auto forwardedCount = 1;
+				const auto isForwarded = notifyItem->Has<HistoryMessageForwarded>();
+				const auto isAlbum = notifyItem->groupId();
 
-				auto ms = crl::now();
-				auto history = notifyItem->history();
-				auto j = _whenMaps.find(history);
+				auto forwardedOrGroupedItem = isForwarded || isAlbum ? notifyItem : nullptr; // forwarded and album notify grouping
```
john-preston: Please use parentheses in such expressions `(isForwarded || isAlbum) ? notifyItem : nullptr`.

Telegram/SourceFiles/window/notifications_manager.cpp
```diff
@@ -334,24 +358,57 @@ void System::showNext() {
...
-							if (forwardedItem) {
-								auto nextForwarded = nextNotify->Has<HistoryMessageForwarded>() ? nextNotify : nullptr;
-								if (nextForwarded
-									&& forwardedItem->author() == nextForwarded->author()
-									&& qAbs(int64(nextForwarded->date()) - int64(forwardedItem->date())) < 2) {
-									forwardedItem = nextForwarded;
-									++forwardedCount;
-								} else {
-									nextNotify = nullptr;
+							if (forwardedOrGroupedItem) {
```
john-preston: I'd say that this is `groupedItem` or `forwardedOrAlbumItem`. The whole process here is grouping items for the notification, so we shouldn't use "group" / "grouping" here both in terms of album grouping and in terms of notifications grouping.

Here we can use the term "forwarded" and "album" to describe, what type of items we are grouping.

Telegram/SourceFiles/window/notifications_manager.cpp
```diff
@@ -334,24 +358,57 @@ void System::showNext() {
...
+								const auto isNextGood = (isForwarded && nextNotify->Has<HistoryMessageForwarded>())
+									|| (isAlbum && nextNotify->groupId());
+								const auto nextItem = isNextGood ? nextNotify : nullptr;
```
john-preston: I would suggest using `groupedItem` instead of `forwardedItem`. So here it will be not `isNextGood`, but `canNextBeGrouped`. And if next can be grouped and next item is in the same group with current `groupedItem` you update `groupedItem = nextItem;` and in case of forwarded increment `forwardedCount`.

Telegram/SourceFiles/window/notifications_manager.cpp
23rd: Yeah, just a little confused by `groupId()`.

Telegram/SourceFiles/window/notifications_manager.h
23rd: Thanks for your advice.

john-preston:
Great, thanks.

PR #5752, opened 2019-03-03
Fixed crash in window resizing when layer is opened.

23rd:
#### Steps to reproduce:
1. Resize the window to minimum width.
2. Open `View profile`/`View channel info`/`Settings`.
3. Maximize the window.
4. Return size of the window.
5. Crash.

The crash is due to the connection icon that was optimized here 26f1ade5ba22b2b351b12e3ce8cc179f2e013758.

image


john-preston:
Thanks!

PR #5765, opened 2019-03-06
Added Search Shortcut in Shared Media.

### LINKED ISSUE #5756: CTRL+F doesn't work in shared documents/audios/links
### Steps to reproduce
1. open the shared documents/audios/links screen of a chat
2. use CTRL+F to search for something

### Expected behaviour
The focus is brought to the search input box and you can write in it

### Actual behaviour
Nothing happens when you use CTRL+F

### Configuration
**Operating system:** Linux Mint, Windows 10

**Version of Telegram Desktop:** 1.5.15

</details>

23rd:
This PR closes #5756.

In the code.
I moved a lambda of the search button into a separated method to avoid repeating parts of the code and to make it possible to call this method from the parent class.
Also I added shortcut as `Command::Search` instead of `Ctrl + F` to reach more flexibility.

### DIFF
--- a/Telegram/SourceFiles/info/info_top_bar.cpp
+++ b/Telegram/SourceFiles/info/info_top_bar.cpp
@@ -213,10 +213,7 @@ void TopBar::createSearchView(
 
 	auto button = base::make_unique_q<Ui::IconButton>(this, _st.search);
 	auto search = button.get();
-	search->addClickHandler([=] {
-		_searchModeEnabled = true;
-		updateControlsVisibility(anim::type::normal);
-	});
+	search->addClickHandler([=] { showSearch(); });
 	auto searchWrap = pushButton(std::move(button));
 	registerToggleControlCallback(searchWrap, [=] {
 		return !selectionMode()
@@ -284,6 +281,11 @@ void TopBar::createSearchView(
 		_searchModeAvailable = visible || alreadyInSearch;
 		updateControlsVisibility(anim::type::instant);
 	}, wrap->lifetime());
+}
+
+void TopBar::showSearch() {
+	_searchModeEnabled = true;
+	updateControlsVisibility(anim::type::normal);
 }
 
 void TopBar::removeButton(not_null<Ui::RpWidget*> button) {
--- a/Telegram/SourceFiles/info/info_wrap_widget.cpp
+++ b/Telegram/SourceFiles/info/info_wrap_widget.cpp
@@ -21,6 +21,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 #include "ui/wrap/fade_wrap.h"
 #include "ui/search_field_controller.h"
 #include "calls/calls_instance.h"
+#include "core/shortcuts.h"
 #include "window/window_controller.h"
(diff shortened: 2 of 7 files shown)


john-preston:
What happens if the shared media section is opened in the third column together with the chat? Will Ctrl+F open search in the chat, in the shared media, in both, in one by random? I think that search in the chat should have a priority (shortcut handlers can provide a priority when they offer a handler).

23rd:
Yeah, sorry for this miss.
Switched priority from `1` to `0`.

Telegram/SourceFiles/info/info_top_bar.cpp
```diff
@@ -213,10 +213,7 @@ void TopBar::createSearchView(
 
 	auto button = base::make_unique_q<Ui::IconButton>(this, _st.search);
 	auto search = button.get();
-	search->addClickHandler([=] {
-		_searchModeEnabled = true;
-		updateControlsVisibility(anim::type::normal);
-	});
+	search->addClickHandler([this] { showSearch(); });
```
john-preston: Right now I try to use `[=]` for `this` capture in safe contexts (almost always there is just `[=]` for lambdas that are passed somewhere outside and will be called after current code and `[&]` for all places where lambda is used locally only while it was created).

I'm aware that in the future standard capturing `this` by `[=]` will be deprecated and I'll be forced to write explicitly `[this]` or `[=, this]` or `[=]` in all the same places - and it is very sad. But we're not there yet :)

Telegram/SourceFiles/info/info_wrap_widget.cpp
```diff
@@ -78,6 +79,19 @@ WrapWidget::WrapWidget(
 	restoreHistoryStack(memento->takeStack());
 }
 
+void WrapWidget::setupShortcuts() {
+	Shortcuts::Requests(
+	) | rpl::filter([=] {
+		return requireTopBarSearch();
+	}) | rpl::start_with_next([=](not_null<Shortcuts::Request*> request) {
+		using Command = Shortcuts::Command;
+		request->check(Command::Search, 0) && request->handle([=] {
```
john-preston: grep by `->check(Command::Search` gives `->check(Command::Search)` in HistoryWidget (because it is never destroyed and always is called) and `->check(Command::Search, 1)` in all other places.

If we want HistoryWidget to have higher priority than WrapWidget, then WrapWidget should use priority 0 `->check(Command::Search)`, HistoryWidget should use `1` and other places should use `2`.

Telegram/SourceFiles/info/info_top_bar.cpp
23rd: Yeah, I just saw `_forward->entity()->addClickHandler([this] { performForward(); });` and I thought to stick to a style within a single file, especially since the commit associated with this line was not very long ago.

Telegram/SourceFiles/info/info_wrap_widget.cpp
23rd: Oops, I greped it by `->check(Command::Search, 1)` and missed one result.

john-preston:
Great, thanks.

PR #5772, opened 2019-03-08
Added ability to see attached stickers on photos.

23rd:
This PR adds an ability to see the attached stickers on the photos.

In the code.
The first of all I added a missed `bool hasSticker` to `PhotoData`, that receives a value from the MTP data type.
The second thing I added is a new `ApiWrap::requestAttachedStickerSets` method that uses the RPC `messages.getAttachedStickers` method. After a successful response we check how many sticker packs on the photo.
If the photo has the only one sticker pack, we show the simple `StickerSetBox`.
If the photo has 2+ sticker packs, we should show the `StickersBox` that has a list of sticker packs.

Unfortunately, the `StickersBox` has been slightly modified to show the attached stickers. 
What was added:
 - Added a new `Attached` value to the `Section` enum.
 - Added a new constructor for `StickerBox`, that takes a `MTPVector<MTPStickerSetCovered>` (the response data of the `messages.getAttachedStickers` method) argument.
 - Added a new `_attached` tab with index 0.
 - Added a new `showAttachedStickers` method, that parses the `MTPVector<MTPStickerSetCovered>` and appends the sets to the `_attached` tab.

And finally a call of the `ApiWrap::requestAttachedStickerSets` method was added to context menu of the photo and to context menu in the `Media Viewer`.

image


Telegram/SourceFiles/apiwrap.cpp
```diff
@@ -2624,6 +2626,42 @@ void ApiWrap::resolveWebPages() {
 	}
 }
 
+void ApiWrap::requestAttachedStickerSets(not_null<PhotoData*> photo) {
+	const auto requestId = request(MTPmessages_GetAttachedStickers(
```
john-preston: It's better to have `_attachedStickerSetsRequestId` field in the object and here you first do `request(_attachedStickerSetsRequestId).cancel();` and then `_attachedStickerSetsRequestId = request(..).send();`, so that while you don't have internet connection (or you're a really fast guy) you will still have only one pending request and one box being shown after the request is done.

Telegram/SourceFiles/apiwrap.cpp
```diff
@@ -2624,6 +2626,42 @@ void ApiWrap::resolveWebPages() {
...
+		MTP_inputStickeredMediaPhoto(photo->mtpInput())
+	)).done([=](const MTPVector<MTPStickerSetCovered> &result) {
+		if (result.v.size() == 1) {
```
john-preston: I'd say:
```
if (result.v.isEmpty()) {
.. show not found box ..
  return;
} else if (result.v.size() > 1) {
.. show StickersBox with the list ..
  return;
}
.. main code path for a single set ..
```

Telegram/SourceFiles/apiwrap.cpp
```diff
@@ -2624,6 +2626,42 @@ void ApiWrap::resolveWebPages() {
...
+			const auto &setCovered = result.v.front();
+			const MTPDstickerSet *setData = nullptr;
+			switch (setCovered.type()) {
```
john-preston: This is a legacy way of inspecting a TL object. Now there is a better way to handle a TL object with different constructors (or even only a single one):
```
const auto setData = result.v.front().match([&](const MTPDstickerSetCovered &data) {
  return data.vset.match([&](const MTPDstickerSet &data) {
    return &data;
  });
}, [&](const MTPDstickerSetMultiCovered &data) {
  return data.vset.match([&](const MTPDstickerSet &data) {
    return &data;
  });
});
```
This has better compile time checks (that no constructors were forgotten) and doesn't require a mutable variable. Also in this case it can be even shorter (but it is unlikely to be autocompleted in the IDE):
```
const auto setData = result.v.front().match([&](const auto &data) {
  return data.vset.match([&](const MTPDstickerSet &data) {
    return &data;
  });
});
```

Telegram/SourceFiles/apiwrap.cpp
```diff
@@ -2624,6 +2626,42 @@ void ApiWrap::resolveWebPages() {
...
+			} break;
+			case mtpc_stickerSetMultiCovered: {
+				const auto &d = setCovered.c_stickerSetMultiCovered();
+				if (d.vset.type() == mtpc_stickerSet) {
+					setData = &d.vset.c_stickerSet();
+				}
+			} break;
+			}
+
+			MTPInputStickerSet setId = (setData->vid.v && setData->vaccess_hash.v)
```
john-preston: Will it work with `const auto setId = (..) ? .. : ..`?
Also please indent the lines with `? MTP_inputStickerSetID(..)` and `: MTP_inputStickerSetShortName(..)`.

Telegram/SourceFiles/boxes/stickers_box.cpp
```diff
@@ -145,6 +145,50 @@ StickersBox::StickersBox(QWidget*, not_null<ChannelData*> megagroup)
...
 
+StickersBox::StickersBox(QWidget*, const MTPVector<MTPStickerSetCovered> &attachedSets)
+: _section(Section::Attached)
+, _attached(0, this, Section::Attached)
+, _attachedSets(attachedSets) {
+}
+
+void StickersBox::showAttachedStickers() {
+	auto addedSet = false;
+	for_const (const auto &stickerSet, _attachedSets.v) {
```
john-preston: I'm going away from my `for_const` idea now, don't like macros :) Here it is not required because `_attachedSets` is already `const`. In other places where Qt containers are const (like in all MTProto request results) I use just a range-based for loop too. In other places I move to STL containers and it's not required too. In the small amount of cases where I still have mutable Qt containers `for (blabla : std::as_const(container)) {` can be used.

Telegram/SourceFiles/boxes/stickers_box.cpp
```diff
@@ -145,6 +145,50 @@ StickersBox::StickersBox(QWidget*, not_null<ChannelData*> megagroup)
...
+		const MTPDstickerSet *setData = nullptr;
```
john-preston: The same idea with `stickerSet.match` can be used here. And that way we see that there can't be `nullptr` in `setData` in any way.

Telegram/SourceFiles/apiwrap.cpp
23rd: Actually I'm not sure the server can ever give to us an empty vector. In this case, it is likely to give an error. 
I tested it even with blocked sticker pack.
But you are right, 
```if (size == 1) {
} else {
}```
this is not enough for handler.

File: Telegram/SourceFiles/apiwrap.cpp (hunk 1 above)
Comment (23rd) [reply Telegram/SourceFiles/apiwrap.cpp]: Can you explain me in which cases I should use
`request(base::take(_someRequestId)).cancel();`
and when I should check it
`if (_someRequestId) {`
?

File: Telegram/SourceFiles/apiwrap.cpp (hunk 3 above)
Comment (23rd) [reply Telegram/SourceFiles/apiwrap.cpp]: Oh, yes, it's much more convenient, I really did not like these almost identical checks.

File: Telegram/SourceFiles/apiwrap.cpp (hunk 1 above)
Comment (john-preston) [reply Telegram/SourceFiles/apiwrap.cpp]: @23rd It depends on what you want to do with repeated attempts to send the request.

If you want each repeated attempt to send a copy of request, you do just `request(..).send();`.

If you want to send one at a time and ignore all other attempts while the first one didn't finish, you do:
```
if (_requestId) return;
_requestId = request(...).done([=](..) {
  _requestId = 0;
  ..
}).fail([=](..) {
  _requestId = 0;
  ..
}).send();
```

If you want only the last request to complete and invoke the handlers you do:
```
request(_requestId).cancel();
_requestId = request(...).send();
```

If in addition to that you want to know if a request is performed right now (have a valid request_id or zero, if none is working right now), you combine the last two:
```
request(base::take(_requestId)).cancel(); // this puts 0 to _requestId.
_requestId = request(..).done([=](..) {
  _requestId = 0;
  ..
}).fail([=](..) {
  _requestId = 0;
  ..
}).send();
```

File: Telegram/SourceFiles/apiwrap.cpp (hunk 2 above)
Comment (john-preston) [reply Telegram/SourceFiles/apiwrap.cpp]: Another point was about handling a short case first and then the long one, to make less indentation :)

File: Telegram/SourceFiles/data/data_session.h [hunk 7]
```diff
@@ -342,6 +342,7 @@ class Session final {
 		const uint64 &access,
 		const QByteArray &fileReference,
 		TimeId date,
+		const bool &hasSticker,
```
john-preston: I sometimes do this with `uint64` (although I shouldn't, it's small and it's fine to pass by value), but `const bool &` is definitely not needed :)

Telegram/SourceFiles/apiwrap.cpp
23rd: Oh, that way.

Telegram/SourceFiles/apiwrap.cpp
23rd: Thanks for explaining.

Telegram/SourceFiles/data/data_session.h
23rd: Another one miss. =(

john-preston:
That's a lot of work, thanks. Sadly, you almost never can use attached stickers (that are usually masks) in tdesktop, so you see them, but only to see a message to use mobile apps for them.

PR #5775, opened 2019-03-11
Added Shortcut to jump to the Saved Messages.

23rd:
It is a little bit sad that we have `Ctrl+1-5` shortcuts for pinned dialogs, but still have no shortcut for the Saved Messages.
We just have to have it. =)
And of course it should be editable.


john-preston:
Thanks!

PR #5820, opened 2019-03-21
Removed "Export history" from context menu for empty chats.

23rd:
There's nothing to export, so we don't need this option when chat is empty.

image

### DIFF
--- a/Telegram/SourceFiles/data/data_peer.cpp
+++ b/Telegram/SourceFiles/data/data_peer.cpp
@@ -7,8 +7,6 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 */
 #include "data/data_peer.h"
 
-#include <rpl/filter.h>
-#include <rpl/map.h>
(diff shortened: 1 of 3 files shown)


Telegram/SourceFiles/data/data_peer.cpp
```diff
@@ -376,6 +377,20 @@ void PeerData::setPinnedMessageId(MsgId messageId) {
 	}
 }
 
+bool PeerData::canExportChatHistory() const {
+	if (_owner->history(id)->blocks.empty()) {
```
john-preston: This check looks redundant.

23rd: When an empty conversation with the user is opened, this check is triggered and we avoid using `for` loops.

23rd: But double accessing to `_owner->history(id)` is worse. =)

john-preston:
Thanks!

PR #5822, opened 2019-03-21
 Fixed drawing of custom thumbnails in edit_caption_box.

23rd:
Unfortunately, tdesktop does not respect sizes of a custom video thumbnails. =(
This PR fixes it only for the EditCaptionBox.

In the code.
This is pretty easy when the media is a video or a not loaded GIF — we just take the dimensions from the thumbnail. 
But it comes harder when we have the loaded GIF. In that case we should separately save the dimensions of thumbnail and GIF. Then we can draw the thumbnail at the first second and then to draw the laoded GIF.

Before patch: [i.imgur.com/xDNsVg1.mp4](https://i.imgur.com/xDNsVg1.mp4)
After patch: [i.imgur.com/ioUvR02.mp4](https://i.imgur.com/ioUvR02.mp4)

### DIFF
--- a/Telegram/SourceFiles/boxes/edit_caption_box.cpp
+++ b/Telegram/SourceFiles/boxes/edit_caption_box.cpp
@@ -51,9 +51,13 @@ EditCaptionBox::EditCaptionBox(
 		dimensions = QSize(photo->width(), photo->height());
 		image = photo->large();
 	} else if (const auto document = media->document()) {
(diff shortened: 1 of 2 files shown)


john-preston:
@23rd What do you mean by custom thumbnails? When the video is sent with a thumbnail not being prepared from that video? It is senders fault really :(

john-preston:
Ok, thanks!

Issue #5740, opened 2019-02-27
The appdata doesn't have an OARS rating


LeandroStanger:
@grulja You can use https://github.com/flathub/flathub/wiki/AppData-Guidelines to help make sure you get everything.

grulja:
This should be reported to upstream.

Aokromes:
https://github.com/telegramdesktop/tdesktop/commit/f8740c5894b5737762dffe15fa892741fcbe0416

PR #5831, opened 2019-03-24
Update AppStream metadata with information about releases and keywords

grulja:


grulja:
This will solve https://github.com/telegramdesktop/tdesktop/issues/5740.

john-preston:
@grulja It solves only one particular version :( there was such pr already, but it is required for me to be able to fill changelog only in one place and then use in buildscripts, now it is changelogs.txt

grulja:
I was hoping you can do this one simple step manually, but I believe this can be also somehow scripted or a tool written to do that. I'll look what can be done about that. Having version and changelog in AppStream metadata is a good step towards users, they will see it in gnome-software or in KDE plasma-discover.

grulja:
I was thinking about modifying your python scripts, but given I have zero knowledge of python and limited time, I will just keep patching the appstream metadata in the Flatpak version of Telegram.

john-preston:
@grulja I hope to reach that. Still hope 😭

PR #5860, opened 2019-03-29
Do not upload wallpaper after updating

mymedia2:
* This behaviour may be unexpected for a user and cause to leak private pictures to the remote server.
 * The patch prevents the checkUploadWallPaper() method from being called during startup initialization. The user still has an opportunity to send his subsequent wallpapers to Telegram Cloud. To do this, he can click on the link "Choose from file".

### DIFF
--- a/Telegram/SourceFiles/window/themes/window_theme.cpp
+++ b/Telegram/SourceFiles/window/themes/window_theme.cpp
@@ -389,11 +389,7 @@ void ChatBackground::start() {
 }
 
 void ChatBackground::refreshSession() {
-	const auto session = AuthSession::Exists() ? &Auth() : nullptr;
-	if (_session != session) {
-		_session = session;
-		checkUploadWallPaper();
-	}
+	_session = AuthSession::Exists() ? &Auth() : nullptr;
 }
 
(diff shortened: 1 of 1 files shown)


john-preston:
How is it different from the user choosing a private picture in the new version?

mymedia2:
> How is it different from the user choosing a private picture in the new version?

If user clicks to the "Choose wallpaper" button by themselves, it expresses their own desire. But software is commonly updated automatically without user interaction. In that case, it is rather not good if something unexpected may occur. For instance, uploading a private image to a remote server might be considered undesirable.

john-preston:
@mymedia2 The user clicks "Choose from file" in the Settings anyway (maybe in the old version, maybe in the new). After he installed the new version he has no idea that "Choose from file" for the chat background works differently than the same button if he installed an old version.

If you think that uploading an image, that was chosen by clicking in Settings a "Choose chat background from file" button, to the server is undesirable — all users of new version get that behaviour anyway. If you think that it's fine — then there is no difference for the existing chat backgrounds, they were chosen the same way, by clicking in Settings a "Choose chat background from file" some time ago.

PR #5883, opened 2019-04-05
Correctly move out the value.

amosbird:


john-preston:
Those were already done, sorry.

PR #5901, opened 2019-04-08
Fixed "Show in Folder" feature on Gnome 3.30+

xvitaly:
Parameter --no-desktop is deprecated and should not be used.


john-preston:
@xvitaly Will it only fix the feature on Gnome 3.30+ or also break on some older Gnome?

john-preston:
(or not Gnome)

john-preston:
In which version of nautilus it started breaking the behaviour and in which it became not needed? Perhaps nautilus version should be queried before deciding whether it should be passed.

xvitaly:
> Will it only fix the feature on Gnome 3.30+ or also break on some older Gnome?

It should work fine, but I will test it on old Ubuntu version in VM.

xvitaly:
@john-preston Tested on Ubuntu 16.04. Works great.

john-preston:
Thanks!

PR #5943, opened 2019-04-21
Updated building-xcode.md. Moved number of threads into variable.

23rd:
This small PR moves existing thread parameters to one place, as it is done in cmake doc.


stek29:
@23rd are you going to merge it?

PR #5946, opened 2019-04-21
Correctly set _desktopFile to Telegram.desktop if Telegram.desktop was found

Johnnynator:


stek29:
Please create PR to dev, not to master

PR #5994, opened 2019-05-10
Allow GNU/Linux maintainers to forward launcher name

xvitaly:
Use optional compiler directive -DTDESKTOP_LAUNCHER_FILENAME to forward
required XDG desktop file name.

If not specified, fallback to default value will be used.


john-preston:
Thanks.

PR #6011, opened 2019-05-15
Updated building-cmake.md

23rd:
Added `-y` parameters to make installation of dependencies non-interactive.
Changed order of installation for correct linking of GCC.


john-preston:
Thanks!

PR #6025, opened 2019-05-18
Fixed building of PCH using Clang compiler

xvitaly:
Now Telegram Desktop can be built using Clang compiler.


john-preston:
Thanks!

PR #6028, opened 2019-05-19
Fixed AppData validation error

xvitaly:
Fixed AppData validation error: tags cannot be empty.

Before:
```
$ appstream-util validate-relax --nonet telegramdesktop.appdata.xml
telegramdesktop.appdata.xml: FAILED:
• tag-invalid           : Expected children for tag
Validation of files failed
```

After:
```
$ appstream-util validate-relax --nonet telegramdesktop.appdata.xml
telegramdesktop.appdata.xml: OK
```


john-preston:
@xvitaly I've applied a patch for flathub version that was filling out the releases tag :( But that's OK, I'll move the tag itself into the patch as well.

Issue #5858, opened 2019-03-28
Constant crashes on MacOS after latest 10.14.4 update


kalieye:
### Steps to reproduce
1. open telegram
2. use it/leave it on the background
3. boom goes the dynamite

### Expected behaviour
Didn't crash until the latest MacOS update

### Actual behaviour
Crashes every few minutes

### Configuration
**Operating system**: MacOS 10.14.4 (Mojave)

**Version of Telegram Desktop**: 1.6.3 (latest App Store version)

**Used theme**: Midnight

<details><summary><b>Logs</b>:</summary>
Insert log.txt here (if necessary)
</details>

i can't seem to find any logs either inside the contaier or in the crashlogs of osx, any suggestions on where to get them?

beeuhtricks:
Same here. Not quite every few minutes, but probably at least a dozen times a day.

mik9:
It crashes after OS switches between integrated and discrete gpu.

john-preston:
@mik9 Yes, it looks like that. I did have a crash on an external display connection. I'll test more today.

john-preston:
@all

This is a Qt crash that is caused (I think) by Apple bug being added in OpenGL support in macOS 10.14.4.

I could reproduce the crash. As I understand it happens in all Qt versions, at least here is Qt Creator crash report with the same stacktrace:

https://bugreports.qt.io/browse/QTCREATORBUG-22215

Qt team is investigating it and I hope they'll come up with a workaround. Maybe it can be reported to Apple as well, but I wouldn't count on Apple fixing it, they've deprecated OpenGL completely in macOS 10.14 and I'm afraid their priority for the bug will be low.

john-preston:
I can confirm that this is not a Qt bug, but macOS 10.14.4 bug in any OpenGL application.

Here I've attached a simplest OpenGL program in pure Objective-C. If you compile it in Xcode 10.2 and run on macOS 10.14.4 with two GPUs (integrated and discrete) it will crash if the laptop will switch from one to another.

[TestOpenGL.zip](https://github.com/telegramdesktop/tdesktop/files/3026999/TestOpenGL.zip)

john-preston:
I've submitted a bug report to Apple.

PR #6031, opened 2019-05-19
Re-enable integrated GPU on macOS.

f355:
This limits the fix introduced in a058e75a to just 10.14.4, as #5858 seems to be fixed in 10.14.5.


john-preston:
Yeah, thanks. Works awful, though. Flickering, stale window state (till repaint) etc. macOS ¯\_(ツ)_/¯

PR #6088, opened 2019-06-04
Disable auto download for Files in default setting

paulkim3151:
I work for a company that uses Telegram on many computers. Multiple files are uploaded to the group chat room, and all of the files, including those I don't need personally, are automatically downloaded, making  uncomfortable. Therefore, it would be better for job to not automatically download the files in default settings.

In addition, it would be more appropriate to disable auto-downloading, considering files such as unidentified executables.


john-preston:
@paulkim3151 We believe that the default option should still be autodownload being enabled. If you're uncomfortable with autodownload you can disable it in Settings.

paulkim3151:
Oh I see. Thanks for the comment.

PR #6109, opened 2019-06-07
Show name instead of phone in dialog title (#2927)

### LINKED ISSUE #2927: Revert to user's chosen name instead of phone number after removing from contacts
### Steps to reproduce
1. Add user as mutual contact
2. Open a chat with the user
3. Remove the contact

### Expected behaviour
After removing a user from contacts, the name displayed in messages and on the chat list should revert to the user's selected name. This is the current behavior in group chats. This behavior is permanent, as Telegram is designed to never remove a user's phone number once it is shared.

As a user, I have no desire to connect Telegram to my actual phone/desktop/whatever contacts, particularly because it prevents me from seeing the user's chosen name. It is frustrating that once someone else has added me as a contact, I can never see their proper name again, even if I remove the contact.

### Actual behaviour
When the user is removed from contacts, their name in direct chats and the chat list appears as their phone number. Group chats display the user's chosen name.

Below are screenshots of what the user looks like in the private chat and chat list, and then what the same user looks like in the user detail view. As you can see, it's quite inconsistent.

phone_number
detail_view

### Configuration
Windows 10 Pro, Telegram Desktop 1.0.2

rikakomoe:
Fix #2927.

### DIFF
--- a/Telegram/SourceFiles/data/data_peer.cpp
+++ b/Telegram/SourceFiles/data/data_peer.cpp
@@ -595,9 +595,7 @@ not_null<const PeerData*> PeerData::migrateToOrMe() const {
 const Text &PeerData::dialogName() const {
 	return migrateTo()
 		? migrateTo()->dialogName()
-		: (isUser() && !asUser()->phoneText.isEmpty())
-			? asUser()->phoneText
-			: nameText;
+		: nameText;
 }
 
 const QString &PeerData::shortName() const {


john-preston:
This will be improved in the next version, the name will be shown in the chats list and the phone number only in the chat title.

rikakomoe:
phone number showing in the chat title is annoying as well. 😕

ralesk:
@john-preston that makes no sense whatsoever :(

PR #6130, opened 2019-06-11
Re-format tdesktop/Telegram/SourceFiles/* using Resharper

rasoulcpu:


john-preston:
@rasoulcpu But why?

rasoulcpu:
> @rasoulcpu But why?

Simply because (in my opinion) it's more readable 😊
just i want to suggest you to change coding style conventions. and my PR is just a preview,thanks 🙏

PR #6131, opened 2019-06-12
Get language name and country name from QLocale (Linux)

vitlav:


john-preston:
@vitlav Platform::SystemCountry returns two-symbol country code, like US or UK, but QLocale::countryToString(locale.country()) returns UnitedStates.

vitlav:
> @vitlav Platform::SystemCountry returns two-symbol country code, like US or UK, but QLocale::countryToString(locale.country()) returns UnitedStates.
Need we ISO 639-1 two-letter codes in that place?

vitlav:
Well, if we need two-symbol code for a language name and two-symbol code for a country name, just split locale name (ru_RU, en_GB, en_US, el_GR) by lodash.

john-preston:
Thanks.

Issue #6157, opened 2019-06-20
Template Change


SmartManoj:
Colon and spacing changed

```### Configuration
**Operating system:**

**Version of Telegram Desktop:**

**Used theme**:
```
To
```### Configuration
**Operating system** : 

**Version of Telegram Desktop** : 

**Used theme** : 
```

Sasha-Sorokin:
Why do you need spaces before the colons?

SmartManoj:
In this `**Operating system:**`
No space after the star.
So if we type `**Operating system:**Win` bold does not work

SmartManoj:
> Why do you need spaces before the colons?

Is colon need bold also?

Sasha-Sorokin:
I mean, you don't need spaces after the asterisks (before colons):

```
**Operating system**: 

**Version of Telegram Desktop**: 

**Used theme**: 
```

---

Also you'd better open a pull request to change the Issue Template instead of creating a new issue:
https://github.com/telegramdesktop/tdesktop/blob/master/.github/ISSUE_TEMPLATE.md

PR #6159, opened 2019-06-21
Update ISSUE_TEMPLATE.md

SmartManoj:
Changed alignment
In this **Operating system:**, No space after the star.
So if we type **Operating system:**Win, bold does not work
Eg: **Operating system:**Win

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -15,9 +15,9 @@ Tell us what should happen
 Tell us what happens instead
 
 ### Configuration
-**Operating system:**
+**Operating system**:
 
-**Version of Telegram Desktop:**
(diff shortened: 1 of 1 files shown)


SmartManoj:
Also please update to the new issue template workflow.

SmartManoj:
#6157
https://github.com/telegramdesktop/tdesktop/issues/6157#issuecomment-504274091

23rd:
Why do you want to update the issue template if you ignore it and don't use in half of your issues anyway? 🤔

SmartManoj:
> Why do you want to update the issue template if you ignore it and don't use in half of your issues anyway? 🤔

I thought issue template is only for bugs

SmartManoj:
@stek29 
When we get new issue template workflow?

PR #6245, opened 2019-07-08
Fix compile and linking ffmpeg_utility.cpp

perestoronin:


john-preston:
This will crash.

john-preston:
```
QPixelLayout qPixelLayouts[QImage::NImageFormats];
const auto layout = &qPixelLayouts[QImage::Format_ARGB32];
const auto convert = layout->storeFromARGB32PM;
...
convert(blabla)
```

You create an empty array of things and then try to call a method with address from that not-filled-with-anything array.

xvitaly:
@john-preston Do not use Qt private API methods in your code!

john-preston:
@xvitaly This is irrelevant to the fact that this PR can't be merged.

PR #6299, opened 2019-07-15
Added some os and packaging platform

alsoGAMER:


john-preston:
Thanks. I don't think it is a good idea to include links to unofficial builds here, sorry.

PR #6417, opened 2019-08-18
Allow searching for messages by a user who has left a chat, fixes #5667

### LINKED ISSUE #5667: Can't search for messages of left users

udf:
This is PR #5728 but with a fix that allows searching for your own messages. Also [this request](https://github.com/telegramdesktop/tdesktop/pull/5728#pullrequestreview-212555077) has been done.

The original PR message from @TheWug follows:

-----

fixes #5667

Change `Dialog::SearchFromController` to inherit not from `ParticipantsBoxController`, which only shows the users in the group, but instead from `AddSpecialBoxController` which shows the group's users but falls back on global search as well.

I know you don't generally accept pull requests for features and calling this a bug is very borderline but I figured I'd post it anyway.

### DIFF
--- a/Telegram/SourceFiles/boxes/peers/add_participants_box.cpp
+++ b/Telegram/SourceFiles/boxes/peers/add_participants_box.cpp
@@ -445,7 +445,6 @@ void AddSpecialBoxController::loadMoreRows() {
 void AddSpecialBoxController::rowClicked(not_null<PeerListRow*> row) {
 	auto user = row->peer()->asUser();
 	switch (_role) {
-	case Role::Members: return;
 	case Role::Admins: return showAdmin(user);
 	case Role::Restricted: return showRestricted(user);
 	case Role::Kicked: return kickUser(user);
--- a/Telegram/SourceFiles/boxes/peers/add_participants_box.cpp
+++ b/Telegram/SourceFiles/boxes/peers/add_participants_box.cpp
@@ -312,6 +312,8 @@ void AddSpecialBoxController::prepare() {
 	delegate()->peerListSetSearchMode(PeerListSearchMode::Enabled);
 	auto title = [&] {
 		switch (_role) {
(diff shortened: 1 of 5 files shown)


john-preston:
Thanks!

bodqhrohro:
Unfortunately, this resolves the issue only partially. It's still not possible to search for the messages of accounts who don't have a username and of deleted accounts.

udf:
@bodqhrohro Not being able to search for accounts without a username is definitely a bug, does this happen regardless if the account is in the group?

As for deleted accounts, I'm not sure how this would be implemented. Allowing searching by ID would definitely work, however there's no easy way to get the ID of an account in tdesktop.

Aokromes:
kotatogram shows id of user, so maybe someone can cherry pick the commit implementing it.

bodqhrohro:
> @bodqhrohro Not being able to search for accounts without a username is definitely a bug, does this happen regardless if the account is in the group?

I re-checked: it works well with ordinary names, but fails on a name that consists only from an emoji.

Actually, I consider this implementation quirky: not only it's non-intuitive and still gives no clue what users have written in a group, but also brings all the flaws of the global search which is too fuzzy and excessive for this task. Still better than nothing, though, and thank you for that. Ideally, the list of writers instead of the list of participants should be displayed, but AFAIK, it's not implemented in Client API yet :/

PR #6442, opened 2019-08-26
Faster upload/download

SlavikMIPT:


john-preston:
It is not that simple :( On fast networks such constant changes will improve the speed, but on slow networks it will cause over loaded connections which will lead to constant reconnecting by timeouts and decrease in speed / sometimes not loading at all.

SlavikMIPT:
> It is not that simple :( On fast networks such constant changes will improve the speed, but on slow networks it will cause over loaded connections which will lead to constant reconnecting by timeouts and decrease in speed / sometimes not loading at all.

Well, so it can be to regulate the number of concurrent loading threads depending on the file size and network speed?

PR #6457, opened 2019-08-29
Added "Beginning" button to calendar

RadRussianRus:
This pull request adds functionality to jump to first message by pressing "Beginning" button when selecting date.

Calendar with "Beginning" button

Additionaly, you can press **Home** button on keyboard while calendar is open.


john-preston:
Thanks! The beginning date is by default not accurate, but I guess it doesn't matter, let everyone jump to the beginning of Telegram history and they'll see the first message anyway.

PR #6461, opened 2019-08-30
source file .tar.gz

TheoOliveira:
Could you  add this for instalation option. AppImage is fine but i rather use .tar.gz


Aokromes:
i doub you really wanted to do this:

 Conversation 0  Commits 11  Checks 0  Files changed 2,299

PR #6466, opened 2019-08-31
Make link preview title clickable

### LINKED ISSUE #6421: [Feature Request] Make title in link preview clickable
From user's perspective I see a problem when clicking on image from link preview. Sometimes it acts as "click to open URL", sometimes it acts as "click to open image in link preview" and I personally do not see a pattern (and honestly don't have to)

As a first step I suggest making title (in red circle) clickable to always open URL
изображение

A more complex solution is to either always open image or go to URL (the second one is better in case of "hidden links" in HTML symbols).

RadRussianRus:
This PR closes #6421.

You can now click link preview title to open the link that generated it. If site title does not exist, page title will be used for link instead.

Also you can select and copy site link along with page title and description.


john-preston:
Thanks!

Issue #1060, opened 2015-09-04
Shortcuts


auchri:
The goal of this issue is to add the following shortcuts:
## Usage
- [x] Lock telegram with <kbd>CTRL</kbd> + <kbd>L</kbd>
- [ ] Minimize to tray (<kbd>CTRL</kbd> + <kbd>X</kbd> if no text is selected, <kbd>CMD</kbd>  + <kbd>M</kbd> on mac) or open it, if it's already minimized (see #748, global hotkey)
- [x] Completely close telegram (<kbd>CTRL</kbd> + <kbd>Q</kbd>, <kbd>CTRL</kbd> + <kbd>W</kbd> for windows)
- [ ] Open emoji panel when pressing <kbd>TAB</kbd> in text field
- [ ] Insert last sent message into the text field with <kbd>CTRL</kbd> + <kbd>Up</kbd>, on mac <kbd>CMD</kbd> + <kbd>Up</kbd>
- [ ] Mac Fullscreen-Mode <kbd>CMD</kbd> + <kbd>CTRL</kbd> + <kbd>F</kbd> (See #1904)
- [ ] <kbd>Home</kbd> when message editor empty: go to beginning of message history.
- [ ] <kbd>End</kbd> when message editor empty: go to end of message history.
- [ ] <kbd>Shift</kbd> + Click in selection mode (see 4 in #2404)
- [ ] Switch between two last chats (see #3303)
- [ ] Log Out (see #4251)

## Voice messages
- [ ] <kbd>CTRL</kbd> + <kbd>R</kbd> to start recording
- [ ] <kbd>CTRL</kbd> + <kbd>R</kbd> to stop and send recording
- [ ] <kbd>CTRL</kbd> + <kbd>shift (⇧)</kbd> + <kbd>R</kbd> cancel recording

## Emoji

- [ ] <kbd>Tab</kbd> (Single press) - Open Emoji menu
- [ ] <kbd>Tab</kbd> (In Emoji menu _or ultraDoubleTabPressing_) - Open Stickers menu
- [ ] <kbd>Shift</kbd> + <kbd>**←**</kbd> (in Stickers menu, when focused at first sticker pack) - Open GIFs menu
- [ ] <kbd>Shift</kbd> + <kbd>**→**</kbd> (in GIFs menu) - Back to the stickers
- [ ] <kbd>Shift</kbd> + <kbd>**→**</kbd> (in Stickers menu) - Next sticker pack
[...]

PR #6487, opened 2019-09-04
Support Key_End in Chat history

BaseMax:
Hello,

I'm Max. (https://t.me/MAX_BASE)
This is a suggestion to telegram team, to support `End key` in the chat history.
> Note: Because we need to press multi-time PageDown key when we need to go down.

It requires to handle the `End` event in there. (Not complete yet)
Thank you for your attention.

Best,
Max Base

### DIFF
--- a/Telegram/SourceFiles/history/history_widget.cpp
+++ b/Telegram/SourceFiles/history/history_widget.cpp
@@ -5251,6 +5251,11 @@ void HistoryWidget::keyPressEvent(QKeyEvent *e) {
 		_scroll->keyPressEvent(e);
 	} else if (e->key() == Qt::Key_PageUp) {
 		_scroll->keyPressEvent(e);
+	// This is not recommended at this step, may be increased crash in the program.
+	// } else if (e->key() == Qt::Key_Home) { //0x01000010
+	// 	_scroll->keyPressEvent(e);
+	} else if (e->key() == Qt::Key_End) { //0x01000011
+		_scroll->keyPressEvent(e);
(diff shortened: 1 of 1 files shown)


RadRussianRus:
I think you forgot rest of your commit.

`Ui::ScrollArea::keyPressEvent` doesn't catch `End` key. Event is forwarded to `QScrollArea::keyPressEvent`, which is implemented in `QAbstractScrollArea::keyPressEvent`. And [description of this method in docs says](https://doc.qt.io/qt-5/qabstractscrollarea.html#keyPressEvent):
> It handles PageUp, PageDown, Up, Down, Left, and Right, and ignores all other key presses.

BaseMax:
> It handles PageUp, PageDown, Up, Down, Left, and Right, and ignores all other key presses.

Thank you.
Yeah, I did suggest it, It will require to develop.

Best,
Max

RadRussianRus:
Then it should go to issues, not to pull requests.

BaseMax:
> Then it should go to issues, not to pull requests.

Sorry for that, but my commits is not complete yet.
I hope someone continues it. (If it's good)

RadRussianRus:
Make pull request only when you complete your commits. Otherwise it should go to "issues".
Also, there is #1060, which has `End` shortcut as planned.

BaseMax:
> Also, there is #1060, which has `End` shortcut as planned.

Okay, Thank you for mention it.
So telegram team will work on it, and my pull not required!

Regards,
Max

PR #6501, opened 2019-09-07
be compatible with range-v3 0.9.x/1.0 branch

farseerfc:
This made 5 changes to the current code base to be compatible with higher versions of range-v3 library.
1. ranges::iterator_range was renamed to subrange, see https://github.com/ericniebler/range-v3/issues/766
2. PercentCounterItem need an operator== for ranges::sort for some reason
3. include <range/v3/to_container.hpp> changed to include <range/v3/range/conversion.hpp>
4. explicitly specify the upper bound in ranges::view::ints
5. change the deprecated ranges::to_ (with underscore) to ranges::to (without underscore)

### DIFF
--- a/.appveyor/install.bat
+++ b/.appveyor/install.bat
@@ -28,7 +28,10 @@ GOTO:EOF
     git clone -q --depth 1 --branch master https://github.com/telegramdesktop/dependencies_windows.git %LIB_DIR%
     cd %LIB_DIR%
 
-    git clone --depth 1 --branch 0.5.0 https://github.com/ericniebler/range-v3
+    git clone https://github.com/ericniebler/range-v3
+    cd range-v3
+    git checkout 0.5.0
+    cd ..
 
     if exist prepare.bat (
         call prepare.bat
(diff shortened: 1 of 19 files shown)


john-preston:
@farseerfc It fails to build on AppVeyor, you've left 0.5.0 tag there.

Does the new code build in Visual Studio 2019? (with at least one version of range-v3) Because I can't accept changes that break build on the main platform.

john-preston:
I'm fine with replacing 0.5.0 with something later, like 0.9.0, but I think it is still better to bind to some tagged release than to HEAD, the build could break any time.

Aokromes:
latest atm is 0.9.1 for sure bug fixes, maybe better than 0.9.0 https://github.com/ericniebler/range-v3/releases

farseerfc:
Sorry I am trying to setup a Visual Studio 2019 building environment on my spare machine but it's taking longer time than I thought. Will try to test this PR on that when I set it up.
I also think sticking to a tagged release like 0.9.1 will ease future efforts.

john-preston:
@farseerfc Are you sure it works in VS at all? Because I know that 0.5.0 was made to work in VS and then it was broken again, because VS still doesn't fully support expression sfinae and some things.

svenstaro:
@john-preston Is there anything left to do to merge this? It builds fine now apparently.

john-preston:
@farseerfc Thanks!
@svenstaro Done.

PR #6518, opened 2019-09-10
Fix last message tail drawing in theme preview

RadRussianRus:
If you set `msgInBg` color to semi-transparent, you'll see that tail of last message in theme preview will overlay message bubble. The problem is: if `QPainter->setClipping(false)` was called, then first clip instruction (in this case `setClipRegion` for cutting bubble corner) will be skipped.

Current state (1.8.7 beta):

After fix (this pull):


john-preston:
Thanks!

PR #6529, opened 2019-09-12
Add option to disable animated stickers

### LINKED ISSUE #6234: Animated Stickers Autoplay, And Can't Be Stopped
### Steps to reproduce
1.  Set "Autoplay GIFs" to Off under Settings >> Advanced
2.  Open a chat that contains an animated sticker

### Expected behaviour
Animated stickers should respect the Autoplay GIFs setting, or there should be another setting disable animated sticker autoplay.

### Actual behaviour
Animated stickers autoplay, and there appears to be no possible way to stop them or disable this behavior.

### Configuration
**Operating system:**  Windows 7, and Windows 10

**Version of Telegram Desktop:**  1.7.13

**Used theme**:  Default, I guess?

DASPRiD:
This PR adds a checkbox in the sticker options to disable animated stickers from playing entirely. Instead, it will halt on the first frame.

Fixes #6234

### DIFF
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -327,6 +327,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 "lng_settings_suggest_emoji" = "Suggest emoji replacements";
 "lng_settings_suggest_by_emoji" = "Suggest popular stickers by emoji";
 "lng_settings_loop_stickers" = "Loop animated stickers";
+"lng_settings_disable_animated_stickers" = "Disable animated stickers";
 "lng_settings_large_emoji" = "Large emoji";
 "lng_settings_view_emojis" = "View list";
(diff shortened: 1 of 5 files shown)


john-preston:
Sorry, but those changes won't be merged, as stated in [contributing guidelines](https://github.com/telegramdesktop/tdesktop/blob/dev/.github/CONTRIBUTING.md#what-contributions-are-accepted), at least now.

DASPRiD:
What *exactly* is the reason for not merging it? If i understand the guidelines correctly, it's because of the new UI element (checkbox in the settings)?

gmaksimov:
I think that when you add new feature, you should have backward compatibility. In UI it means have option to disable it.

Animated stickers are annoying (they steal attention) when they move as long as you have seen them already. There is config option that disables looping them but the problem is **every time** you open dialog **stickers are being played** one time.

I understand you position on adding new features or new UI elements. But config options are more about disabling new features than adding new UI elements :)

PR #6572, opened 2019-09-19
Fix "All users and admins" checkbox in admin log filter box not unchecking all admins

udf:
Here's a demonstration of the behaviour that this fixes: https://giant.gfycat.com/ImpressionableDefiniteFieldspaniel.webm

### DIFF
--- a/Telegram/SourceFiles/history/admin_log/history_admin_log_filter.cpp
+++ b/Telegram/SourceFiles/history/admin_log/history_admin_log_filter.cpp
@@ -284,9 +284,10 @@ void FilterBox::Inner::createAllUsersCheckbox(const FilterValue &filter) {
 	_allUsers = addRow(object_ptr<Ui::Checkbox>(this, tr::lng_admin_log_filter_all_admins(tr::now), filter.allUsers, st::adminLogFilterCheckbox), st::adminLogFilterSkip);
 	_allUsers->checkedChanges(
 	) | rpl::start_with_next([=](bool checked) {
-		if (checked && !std::exchange(_restoringInvariant, true)) {
+		if (!std::exchange(_restoringInvariant, true)) {
+			auto allChecked = _allUsers->checked();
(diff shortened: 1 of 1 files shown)


john-preston:
It was designed as a feature :( The `All actions` checkbox really means all the other checkboxes, but `All users and admins` checkbox means more. You may want to uncheck it to remove only users, but leave the admins.

But on the other hand if you want all admins without users you indeed can uncheck the `All users and admins` and then check all admins back. The important thing is that in case you do that you don't get `All users and admins` checked back automatically, how it should happen with `All actions`.

udf:
Maybe the wording should be changed? Right now unchecking `All users and admins` feels like it should uncheck, you know, all users and admins.

I agree that it'd be nice to switch between filtering on users and admins, but it's also nice to be able to filter to just one user (which would be achieved by deselecting everyone and selecting the user, like you'd do to filter on one action).

I think that having two checkboxes would work better, one for `All users` and another for `All admins`, but I'm not sure if that would look good or be intuitive to use.

----

Also, I noticed that the filter box only shows admins instead of all users. Which means the `All users and admins` checkbox is not very useful in its current state.

Qwerty-Space:
It's an unexpected outcome when unticking the checkbox, and I think that's important when it comes to UI.  It's more useful to be able to untick all users than to not do anything when unticking it.  Especially since every member isn't even in the checklist.

john-preston:
@udf Right now unchecking "all users and admins" removes all users :) (and admins except for checked below). So if you want to see actions only by admins, but not users, you uncheck it and leave all admins checked below. But I guess you want to see actions of one admin more frequently than actions of all admins without actions of users, so I think it will be better your way.

udf:
@john-preston Thanks!

It would be better if the list showed all members (admins on top, and then normal users). That would let you filter for any member's actions. I'll try that out later this week when I have some time.

We'd still be missing the filtering on either users or admins, but like you said that's probably not necessary.

PR #6578, opened 2019-09-21
Make theme preview fully translatable

RadRussianRus:
This PR makes theme preview fully translatable, like in other official clients.

## Tests

Custom Russian-based locale (all new keys translated):

Official Russian-based locale (no new keys):

English:

### DIFF
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -1570,6 +1570,30 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 "lng_theme_preview_users#one" = "{count} person is using this theme";
 "lng_theme_preview_users#other" = "{count} people are using this theme";
 
+"lng_theme_preview_dialog_name1" = "Eva Summer";
+"lng_theme_preview_dialog_text1" = "Reminds me of a Chinese proverb: the best time to plant a tree was 20 years ago.";
+"lng_theme_preview_dialog_name2" = "Alexandra Smith";
+"lng_theme_preview_dialog_text2" = "This is amazing!";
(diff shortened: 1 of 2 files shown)


john-preston:
@RadRussianRus This is not thread safe :( Theme preview is generated from a random background thread and it will crash if in that moment some langpack updates will happen, for example if an update with new langpack values will come from server.

RadRussianRus:
Hmm, interesting. But some values are already translated. Are they thread safe?

john-preston:
@RadRussianRus Like which ones?

RadRussianRus:
These two:


First is from `Generator::paintDialogs`, [line 575](https://github.com/telegramdesktop/tdesktop/blob/04b9d4bdb5a660fc35b1457cd5ec125784f135d0/Telegram/SourceFiles/window/themes/window_theme_preview.cpp#L575):
```c++
_p->drawText(phRect, tr::lng_dlg_filter(tr::now), QTextOption(st::dialogsFilter.phAlign));
```

And second is from `Generator::paintComposeArea`, [line 539](https://github.com/telegramdesktop/tdesktop/blob/04b9d4bdb5a660fc35b1457cd5ec125784f135d0/Telegram/SourceFiles/window/themes/window_theme_preview.cpp#L539):
```c++
_p->drawText(placeholderRect, tr::lng_message_ph(tr::now), QTextOption(st::historyComposeField.placeholderAlign));
```

john-preston:
@RadRussianRus No, that’s bad too :(

RadRussianRus:
I guess they should be hardcoded then. Alright, guess I'm closing this pull for now. Thank you!

john-preston:
@RadRussianRus No, it could be done right. They should be collected on the main thread and then passed down to the generator.

PR #6589, opened 2019-09-22
Use exact version of libxkbcommon in Travis build

RadRussianRus:
libxkbcommon won't build on Travis if there is no cache because of missing tag checkout.


john-preston:
Thanks.

PR #6612, opened 2019-09-30
Resize message with photo when bubbled

RadRussianRus:
This PR solves two problems by resizing message with photo when it's bubbled (i.e. have additional text like caption or author name).

## Problem 1: barely readable caption

It takes too many lines and it's hard to read this.


john-preston:
@RadRussianRus Yeah, some day it should be done like on iOS app, with nice blurred image filling the paddings.. But I guess this way it is better than nothing.

john-preston:
ups..

PR #6619, opened 2019-10-01
Remove replyTo from switchInlineBotButton in same peer

### LINKED ISSUE #6600: [Bug] switch_inline_query_current_chat in bot messages buttons set force_reply when clicked
Hello there,
I'm reporting an issue with Telegram Desktop on Windows (I've tried all the platforms where Telegram Official apps are supported and the issue is present only on Windows).

### Steps to reproduce
1. Add a bot in a group or supergroup
2. Bot post a message in this group with InlineKeyboardMarkup
3. One of the buttons has switch_inline_query_current_chat parameter
4. Click on this button

### Expected behaviour
The client should show me inline string for the bot in the group chat

### Actual behaviour
The client show me the inline query for the bot in the group chat BUT also set the ForceReply behavior on the bot message that has the inline button I clicked. I don't want reply to the message by default and other clients (in other platforms) don't do that!

### Configuration
Windows 10 Enterprise v 1809
Telegram Desktop 1.8.8
Standard Theme

MrDarkest:
Fix #6600


DO97:
@john-preston

john-preston:
That's strange that there was an explicit completely not needed code path 🤔 But I hope this way is indeed the correct way to handle those buttons.

PR #6621, opened 2019-10-01
Do not show edit timer for supergroup admins

RadRussianRus:
Since supergroup admins can edit their messages without limit, timer is not needed there.


RememberTheAir:
Isn't it possible only if you have the permission to pin message? 🤔

john-preston:
Thanks!

PR #6642, opened 2019-10-03
Show restricted date to user

RadRussianRus:
With this PR users can see their restriction date in blocked message field/inline bot suggestions/sticker box/GIF box.

Currently text tells only "You have been restricted" without date.

### DIFF
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -1768,6 +1768,13 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 "lng_restricted_send_inline" = "The admins of this group restricted you from posting inline content here.";
 "lng_restricted_send_polls" = "The admins of this group restricted you from posting polls here.";
 
+"lng_restricted_send_message_until" = "The admins of this group restricted you from writing here until {date}, {time}.";
+"lng_restricted_send_media_until" = "The admins of this group restricted you from posting media content here until {date}, {time}.";
+"lng_restricted_send_stickers_until" = "The admins of this group restricted you from posting stickers here until {date}, {time}.";
+"lng_restricted_send_gifs_until" = "The admins of this group restricted you from posting GIFs here until {date}, {time}.";
+"lng_restricted_send_inline_until" = "The admins of this group restricted you from posting inline content here until {date}, {time}.";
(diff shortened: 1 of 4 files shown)


john-preston:
@RadRussianRus Does it fit in all cases?

RadRussianRus:
Checked cases:
- [x] Sticker panel: there is more than enough space.
- [x] GIF panel: same as sticker panel.
- [x] Inline results: preview is resized to fit text.
- [x] Message field: it spans to next line, but it fits.

If there are any other test cases to check, tell me and I'll check them too.

Also I'd like to rebuild on top of [this commit](https://github.com/telegramdesktop/tdesktop/commit/60b39c8f7e1a3b7c6cd46f4c591a27ad4da7c5e4), since inline preview bug was fixed.

Screenshots:

RadRussianRus:
Oh, I guess it wasn't fixed.

RadRussianRus:
Okay, I've found a bug. But it seems that it affect original phrase too.

RadRussianRus:
Strange thing though: with drag'n'drop you can see full message, but when you click "Attach" button, it's not.

RadRussianRus:
Found the source of problem. For `chooseAttach` in `HistoryWidget` used `Ui::Toast::Show`, but others using `ShowErrorToast` that is configured with multiline.
https://github.com/telegramdesktop/tdesktop/blob/dcf79df0b2eb0d8bc3757f44c43e62aef9d4fdee/Telegram/SourceFiles/history/history_widget.cpp#L3222

Should I add fix to this pull?

john-preston:
@RadRussianRus Yes, please.

RadRussianRus:
After this fix it works as it should:

RadRussianRus:
Hmm, scheduled messages are still affected though.

RadRussianRus:
It was same problem though. Now it's fixed in scheduled messages too:

RadRussianRus:
Testing with forwards.

When restricted from sending messages:


When restricted from sending media:


When restricted from sending GIFs and stickers:




When restricted from sending polls:

RadRussianRus:
Time formats also work fine:

RadRussianRus:
If user was restricted forever, it doesn't show date, of course.

RadRussianRus:
Is there any other case to test?

Issue #6640, opened 2019-10-03
[Feature Request] AutoLock Time


banghasan:
### Is your feature request related to a problem?
I need to set autolock to active after certain time, eg after 15 minutes. But Telegram only give option to autolock afer 1 minute, 5 minutes, 1 hour, and 5 hours

![autolock](https://user-images.githubusercontent.com/5436959/66126521-d4c5a300-e613-11e9-8acc-3d9bc99c9fbd.png)

### Describe the solution you'd like
I'd like you to add more option for autolock time. Eg after 15 minutes, after 30 minutes

### Describe alternatives you've considered
As an alternative, you could also make it possible for user to input custom time for autolock.

BrainNull:
I need similar option for muting chats/channels/dialogs

> ### Is your feature request related to a problem?
> I need to set autolock to active after certain time, eg after 15 minutes. But Telegram only give option to autolock afer 1 minute, 5 minutes, 1 hour, and 5 hours
> 
> ![autolock](https://user-images.githubusercontent.com/5436959/66126521-d4c5a300-e613-11e9-8acc-3d9bc99c9fbd.png)
> 
> ### Describe the solution you'd like
> I'd like you to add more option for autolock time. Eg after 15 minutes, after 30 minutes
> 
> ### Describe alternatives you've considered
> As an alternative, you could also make it possible for user to input custom time for autolock.

Aokromes:
1/5/15/30/1h/2h/3h/6h/12h better imho

kirsan31:
Still actual.

banghasan:
Still waiting for an update for this feature

banghasan:
When will this feature be implemented? 😨

john-preston:
Closed by b3622b4

Should appear in the next release.

PR #6667, opened 2019-10-11
Add 15 min to autolock box isssue #6640

edrisranjbar:
as @BrainNull requested in issue #6640; I added this feature.
please check it and if it is ok, Merge this.

### DIFF
--- a/Telegram/SourceFiles/boxes/auto_lock_box.cpp
+++ b/Telegram/SourceFiles/boxes/auto_lock_box.cpp
@@ -23,7 +23,7 @@ void AutoLockBox::prepare() {
 
 	addButton(tr::lng_box_ok(), [this] { closeBox(); });
 
-	auto options = { 60, 300, 3600, 18000 };
+	auto options = { 60, 300, 900, 3600, 18000 };
 
 	auto group = std::make_shared<Ui::RadiobuttonGroup>(Global::AutoLock());
 	auto y = st::boxOptionListPadding.top() + st::autolockButton.margin.top();


23rd:
I don't think that's a good idea.

We can add this "15 minutes" item, but what if someone else opens a new issue and wants to add another time?
In my opinion, this problem should be solved by replacing radio button with time picker.
Anyway, I don't make the final decision.

BrainNull:
Timepicker is the best solution

edrisranjbar:
> I don't think that's a good idea.
> 
> We can add this "15 minutes" item, but what if someone else opens a new issue and wants to add another time?
> In my opinion, this problem should be solved by replacing radio button with time picker.
> Anyway, I don't make the final decision.

yes, I agree with time picker too.
if it is good, I can make it.

mymedia2:
Solution with a time picker or a combo box is more difficult to implement since an internal lib_ui has no suitable widget.

Aokromes:
1m/2m/5m/15m/30m
1h/2h/3h/6h/12h/24h better imho

edrisranjbar:
> 1m/2m/5m/15m/30m
> 1h/2h/3h/6h/12h/24h better imho

This will be a long list of options; I think for UI this wont be good enough.
Can we just add a few items?

john-preston:
No, this way the next guy will come around and ask for 10 minutes. Even the original linked issue #6640 asks for 30 minutes as well.

It could be done with a slider similar to one in Settings > Advanced > Manage local storage:

image

with a lot of values, like @Aokromes suggests (we need to keep old 5h there I think), something like:

1/2/5/10/15/20/30 minutes, 1/2/3/4/5/6/9/12/18/24 hours.

PR #6700, opened 2019-10-19
Works on Arch

amirashabani:
If proof is needed: 
tdesktop works on Arch Linux


Self-Perfection:
The list is incomplete anyway. Many other distributions include telegram-desktop as well. I'd rather replace manual enumeration of distros with TD with repology badge:
`[Packaging status](https://repology.org/project/telegram-desktop/versions)`
[Packaging status](https://repology.org/project/telegram-desktop/versions)

john-preston:
Yeah, this doesn't seem like it is going anywhere, too many distributions anyway.

PR #6704, opened 2019-10-19
Make build scripts a bit more configurable

mymedia2:
This adds some GYP variables (cpu_type, bitness, qt_loc, use_common_libs, etc.) to affect build configuration. All additional features are available when the corresponding flag is set. Default behavior is untouched (I hope so).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ env:
    - BUILD_VERSION="disable_network_proxy"
    - BUILD_VERSION="disable_desktop_file_generation"
    - BUILD_VERSION="disable_gtk_integration"
+   - BUILD_VERSION="use_packed_resources"
 
 matrix:
   fast_finish: true
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -43,6 +43,7 @@ OPENAL_PATH="$BUILD/openal-soft"
 OPENAL_CACHE_VERSION="4"
 
 GYP_DEFINES=""
+GYP_PACKED_RESOURCES="0"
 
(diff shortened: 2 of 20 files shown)


john-preston:
@mymedia2 I’m afraid it’ll have to wait till the next release, because it has a lot of gyp changes :( It should happen in October.

mymedia2:
@john-preston, okay, I'll refine the pull request to that time, at least it should pass CI

xvitaly:
I think all this GYP scripts should be removed completely. We can take plain Cmake build script from Kepka fork.

john-preston:
@xvitaly Does this config offer the same static builds with no more runtime dependencies than the current build on all three systems?

xvitaly:
> Does this config offer the same static builds with no more runtime dependencies than the current build on all three systems?

You can forward `-DUSE_BUNDLED=1` to use bundled dependencies.

vitlav:
> @xvitaly Does this config offer the same static builds with no more runtime dependencies than the current build on all three systems?
Can we have two build system? gyp based for static build for all three systems and cmake based for build with system libs?
For instance, I maintain cmake scripts for the official sources (this repo).

john-preston:
@vitlav We can. But someone should maintain the second.

Johnnynator:
@vitlav You are not the only one doing that. @msva for their gentoo overlay and I for @void-linux do currently also maintain a cmake based build system, but neither of us hold full copy right of it and therefore it can't be added into this repo.

john-preston:
Doesn't apply after the new submodules were introduced :(

john-preston:
This will become obsolete once we switch to CMake.

PR #6705, opened 2019-10-21
Properly escape markup text for libnotify

eworm-de:
Message text was passed to libnotify as is, but libnotify expects
properly escaped markup. Sending unscrapted characters results in
warnings:

xfce4-notifyd[2831]: Failed to set text 'foo & bar' from markup due to
error parsing markup: Error on line 1: Entity did not end with a semicolon;
most likely you used an ampersand character without intending to start
an entity — escape a mpersand as &amp;

This adds a call to g_markup_escape_text() to properly escape the text.


eworm-de:
@john-preston, you wrote the libnotify code, want to have a look?

john-preston:
@eworm-de This breaks native notifications on Ubuntu:

image

Without your patch it looks that way:

image

john-preston:
@eworm-de Actually, message text is not passed to libnotify as is:

https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp#L403-L409

eworm-de:
Possibly your notification daemon does not support the capability "actions"?
See  #6784 for the solution.

PR #6753, opened 2019-11-05
Fixes for build instructions

RadRussianRus:
### On Windows
1. Corrected commit to checkout for `ThirdParty/patches`: `b0ec5df4` was not found. Probably `b0ec5df` was implied as it was used on `Libraries/patches`.
2. Removed `git checkout tdesktop` for zlib since `desktop-app/zlib` doesn't have `tdesktop` branch.

### On Linux
1. Since googletest is empty, it should be cloned manually (with removal of empty folders beforehand).
2. CMake can't find `minidump_upload.m` because it doesn't exist. Actual file that needs to be searched is `minidump_upload.cc`. There is only one match and no false positives, so it should be safe to replace it with sed.

### DIFF
--- a/docs/building-msvc.md
+++ b/docs/building-msvc.md
@@ -32,7 +32,7 @@ Open **x86 Native Tools Command Prompt for VS 2019.bat**, go to ***BuildPath*** 
     cd ThirdParty
     git clone https://github.com/desktop-app/patches.git
     cd patches
-    git checkout b0ec5df4
+    git checkout b0ec5df
     cd ../
     git clone https://chromium.googlesource.com/external/gyp
     cd gyp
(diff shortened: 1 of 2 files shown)


docs/building-msvc.md
```diff
@@ -32,7 +32,7 @@ Open **x86 Native Tools Command Prompt for VS 2019.bat**, go to ***BuildPath***
     cd ThirdParty
     git clone https://github.com/desktop-app/patches.git
     cd patches
-    git checkout b0ec5df4
+    git checkout b0ec5df
```
Aokromes: indeed, correct full hash is
b0ec5df5a34276539ba86995943dafc876048e8c
very likely typo.

john-preston:
Thanks!

PR #6778, opened 2019-11-11
Fixed animated stickers cache

RadRussianRus:
When building TD 1.9+ from sources, animated stickers are breaking after first play cycle. This patch corrects typo in ffmpeg_utility, which fixes this problem.


john-preston:
Thanks!

PR #6782, opened 2019-11-12
Fix Linux notification markup support detection

hexchain:
`actions` and `body-markup` are never mutually exclusive. This corrects detection for `body-markup` if `action` is supported.


hexchain:
#6784

PR #6784, opened 2019-11-12
libnotify: capabilities are not exclusive

eworm-de:
Capabilities "actions" and "body-markup" can be available at the
same time.


john-preston:
Thanks!

Issue #3750, opened 2017-08-02
Linux: Clicking on settings tab hangs Telegram when no notification server is running


Kunagi7:
### Steps to reproduce
1. Stop any notification server (like xfce4-notifyd or the ones listed here: https://wiki.archlinux.org/index.php/Desktop_notifications )
2. Start Telegram Desktop
3. Open side drawer and click settings.

### Expected behaviour

The settings screen should open.

If Telegram Desktop requires a notification server and it doesn't find one it should display some kind of message indicating the issue.

### Actual behaviour

Telegram Desktop hangs until I start a desktop notification server.

I found this issue because I use i3 which doesn't have its own notification server.

### Configuration
**Operating system:**
OS: Fedora release 24 (Twenty Four) x86_64
**Version of Telegram Desktop:**
1.1.19
**Used theme**:
Breeze.
<details><summary><b>Logs</b>:</summary>

[2017.08.02 19:16:49] Getting DE, XDG_CURRENT_DESKTOP: ''

[2017.08.02 19:16:49] Getting DE, DESKTOP_SESSION: ''

[2017.08.02 19:16:49] Getting DE, KDE_SESSION_VERSION: ''

[2017.08.02 19:16:49] Getting DE, GNOME_DESKTOP_SESSION_ID: ''

[2017.08.02 19:16:49] Getting DE, KDE_FULL_SESSION: ''

[2017.08.02 19:16:49] DE: Other

[2017.08.02 19:16:49] Tray Icon: Try Qt = [TRUE], Prefer appindicator = [FALSE]

[2017.08.02 19:16:49] App Info: reading settings...

[2017.08.02 19:16:49] App Info: reading encrypted settings...

[2017.08.02 19:16:50] OpenAL Logging Level: (not set)

[2017.08.02 19:16:50] Audio Playback Devices: OpenAL Soft

[2017.08.02 19:16:50] Audio Playback Default Device: OpenAL Soft

[2017.08.02 19:16:50] Audio Capture Devices: Monitor of Built-in Audio Analog Stereo;Monitor 
Source of Simultaneous output to Built-in Audio Analog Stereo, Built-in Audio Digital Stereo (HDMI 3);Monitor of Built-in Audio Digital Stereo (HDMI 3)

[2017.08.02 19:16:50] Audio Capture Default Device: Monitor of Built-in Audio Analog Stereo

[2017.08.02 19:16:50] App Info: reading map...

[...]

Issue #6507, opened 2019-09-08
Clicking on any KDE notification trigger Telegram App


ACheshirov:
### Steps to reproduce
1. Clicking on any KDE notification (Geary's one for example) trigger Telegram app as well
![screenshot](https://user-images.githubusercontent.com/25287575/64489377-308e4d80-d25b-11e9-8f09-3bd166f58b03.png)
![screenshot2](https://user-images.githubusercontent.com/25287575/64489378-35eb9800-d25b-11e9-80f8-0a1944276423.png)
2. You can see that:
- on first screenshot Telegram app is staying in Tray menu.
- on second screenshot when I clicked on Geary notification, Telegram is triggered too (along with Geary app) and appeared to my taskbar.

Second scenario: Telegram is already open and just staying on taskbark, clicking on any notification it starts blinking as well.
I share two screenshots again
![screenshot_3](https://user-images.githubusercontent.com/25287575/64489477-536d3180-d25c-11e9-97c9-8c64df59a64f.png)
![screenshot_4](https://user-images.githubusercontent.com/25287575/64489478-5536f500-d25c-11e9-9a0e-9a9fdb0b63c0.png)

It's really strange behaviour

### Expected behaviour
Telegram should trigger on notification click only when that notification is related to it

### Configuration
OS: KDE neon User Edition 5.16 x86_64 
Kernel: 4.18.0-18-generic

**Version of Telegram Desktop:**
v 1.8.4

**Used theme**:
Night theme

D219:
Same problem on Debian with KDE and Kubuntu.

doctorx32:
Also noticed on Gnome desktop

Aokromes:
https://github.com/telegramdesktop/tdesktop/issues/1026

PR #6825, opened 2019-11-26
Native notifications on Linux without libnotify

### LINKED ISSUE #1026: Telegram comes to focus when clicking notifications from other applications
I actually don't know if this is Telegram or KDE **bug**, but when Telegram is launched and minimized to tray and a new e-mail arrives triggering a "New mail notification" and you push "Show mail" button -> Telegram desktop app is maximized. The same is when you get a download finished in FireFox and receive a notification with a button "Open" (to open a folder with the downloaded file). 
Push the button and for some reason Telegram app is maximized. I use gentoo. Telegram app is simply downloaded from its web-site.

So, **steps to reproduce** (always for me):
1. configure KDE notifications so that they have some buttons to push, e.g. "Show mail" for KMail.
2. Start Telegram and minimize it to tray.
3. Receive a button-containing-notification
4. Push a button in it.

**Expected result**: Show a KMail with new message.
**Actual result**: Telegram is maximized and focused. The new message is opened in a background.
**Reproduces** always.

I attach a screenshot with an example of the notifications and buttons.

telegram-send

ilya-fedin:
This change removes the need of gtk for native notifications on Linux
Closes #1026


ilya-fedin:
Added possible fix for #1026 and (duplicate) #6507 
But I have no idea for which notification daemon there was a hack that caused this problem

Aokromes:
@ilya-fedin edit 1st post and add a note "closes #1026" plz :)

ilya-fedin:
@Aokromes ok

ilya-fedin:
:thinking: travis-ci failed with -Werror=enum-compare

```
/home/travis/build/telegramdesktop/tdesktop/upstream/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp: In member function ‘QStringList Platform::Notifications::NotificationData::getServerInformation()’:

/home/travis/build/telegramdesktop/tdesktop/upstream/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp:90:67: error: comparison between ‘enum QVariant::Type’ and ‘enum QMetaType::Type’ [-Werror=enum-compare]

    if (serverInformationReply.arguments()[i].type() == QMetaType::QString) {
```

But Qt documentation recommends comparsion with QMetaType https://doc.qt.io/QT-5/qvariant.html#type

Should I add -Wno-enum-compare to build scripts?

23rd:
> Should I add -Wno-enum-compare to build scripts?

You can just replace `type()` with `userType()`.
So it will be:
```
if (serverInformationReply.arguments()[i].userType() == QMetaType::QString) {
```

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -7,390 +7,264 @@ For license and copyright information please follow this link:
...
-	NotificationData(const std::shared_ptr<Manager*> &guarded, const QString &title, const QString &body, const QStringList &capabilities, PeerId peerId, MsgId msgId)
-	: _data(Libs::notify_notification_new(title.toUtf8().constData(), body.toUtf8().constData(), nullptr)) {
-		if (valid()) {
-			init(guarded, capabilities, peerId, msgId);
-		}
-	}
-	bool valid() const {
-		return (_data != nullptr);
+	if(!_serverInformation.isEmpty()) {
+		_specificationVersion = parseSpecificationVersion();
```
23rd: According to this place, we don't really need to keep `_serverInformation`.
This list of strings is used only for parsing of specification version, so it can be passed to the method as a single argument.
Something like this.
```
_specificationVersion = parseSpecificationVersion(getServerInformation());
```
And then.
```
QVersionNumber NotificationData::parseSpecificationVersion(QStringList information) {
	if (information.size() >= 4) {
		return QVersionNumber::fromString(information[3]);
	} else {
		LOG(("Native notification error: server information should have 4 elements"));
	}

	return QVersionNumber();
}
```
As a result, we can safely remove `_serverInformation`.

(By the way, the current code style prefers the `std::vector<QString>` over the `QStringList`.

ilya-fedin: Fixed :ok_hand:

ilya-fedin:
I remembered that this PR also fixes #3750, because now there is no call to GetServerName in Supported(), instead the interface is checked for validity. I don’t know if I need to add this to the first post, because issue is auto-closed.

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -7,390 +7,274 @@ For license and copyright information please follow this link:
...
 	}
 
-	~NotificationData() {
-		if (valid()) {
-//			if (_handlerId > 0) {
-//				Libs::g_signal_handler_disconnect(Libs::g_object_cast(_data), _handlerId);
-//			}
-//			Libs::notify_notification_clear_actions(_data);
-			Libs::g_object_unref(Libs::g_object_cast(_data));
+	if (std::find(capabilities.begin(), capabilities.end(), qsl("body-markup")) != capabilities.end()) {
```
23rd: All such lines can be replaced with the range-v3.
Something like this.
```
if (ranges::find(capabilities, qsl("body-markup")) != capabilities.end()) {
```
Also to shorten the code we can move the `capabilities.end()` to a new variable. For example.
```
const auto endCap = end(capabilities);
if (ranges::find(capabilities, qsl("body-markup")) != endCap) {
...
if (ranges::find(capabilities, qsl("actions")) != endCap) {
...
etc
```

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -7,390 +7,274 @@ For license and copyright information please follow this link:
...
+			}
+			copyFrom = i + 1;
+		}
+	}
+	if (copyFrom > 0) {
+		result.append(data + copyFrom, textSize - copyFrom);
+		return result;
+	}
+	return text;
+}
```
23rd: I'm not sure, but why we can't just use [toHtmlEscaped()](https://doc.qt.io/qt-5/qstring.html#toHtmlEscaped)?

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -7,390 +7,274 @@ For license and copyright information please follow this link:
...
-//				Libs::g_signal_handler_disconnect(Libs::g_object_cast(_data), _handlerId);
-//			}
-//			Libs::notify_notification_clear_actions(_data);
-			Libs::g_object_unref(Libs::g_object_cast(_data));
+	if (std::find(capabilities.begin(), capabilities.end(), qsl("body-markup")) != capabilities.end()) {
+		if (!subtitle.isEmpty()) {
+			_body = qstr("<b>") + escapeHtml(subtitle) + qstr("</b>") + '\n' + escapeHtml(msg);
+		} else {
+			_body = escapeHtml(msg);
+		}
```
23rd: This can be simplified.
```
_body = subtitle.isEmpty()
	? escapeHtml(msg)
	: QString("<b>%1</b>\n%2")
		.arg(escapeHtml(subtitle))
		.arg(escapeHtml(msg));
```

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -7,390 +7,274 @@ For license and copyright information please follow this link:
...
+}
 
-	bool hasPoorSupport() const {
-		return _poorSupported;
-	}
-	bool hasActionsSupport() const {
-		return _actionsSupported;
-	}
+bool NotificationData::show() {
+	QStringList actionsStringList(QList<QString>::fromVector(QVector<QString>::fromStdVector(_actions)));
```
23rd: Okay, this is insane.
Perhaps I was wrong and it will be better to store `_actions` as `QStringList` in this particular case.

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: Hmm, it's code from old implementation

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: Fixed

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: Fixed

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: Fixed

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: Fixed

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -7,390 +7,236 @@ For license and copyright information please follow this link:
...
+	QString imageKey;
+
+	if (!_specificationVersion.isNull()) {
+		if (_specificationVersion.majorVersion() == 1 && _specificationVersion.minorVersion() >= 2 || _specificationVersion.majorVersion() > 1) {
+			imageKey = "image-data";
+		} else if (_specificationVersion.majorVersion() == 1 && _specificationVersion.minorVersion() == 1) {
+			imageKey = "image_data";
+		} else if (_specificationVersion.majorVersion() == 1 && _specificationVersion.minorVersion() < 1 || _specificationVersion.majorVersion() < 1) {
+			imageKey = "icon_data";
+		} else {
```
23rd: 1.
These huge conditions can be easily shortened with a new variables.
```
const auto major = _specificationVersion.majorVersion();
const auto minor = _specificationVersion.minorVersion();
if (major == 1 && minor >= 2) {
...
```

2.
It would be better to have a clearly `if ((a && b) || c)` condition instead of unclearly `if (a && b || c)`.

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -7,390 +7,236 @@ For license and copyright information please follow this link:
...
-		LOG(("LibNotify Error: could not get server name!"));
-		return QString();
-	}
-	if (!name) {
-		LOG(("LibNotify Error: successfully got empty server name!"));
-		return QString();
+	if (serverInformationReply.type() == QDBusMessage::ReplyMessage) {
+		for (int i = 0; i < serverInformationReply.arguments().size(); ++i) {
+			if (static_cast<QMetaType::Type>(serverInformationReply.arguments()[i].type()) == QMetaType::QString) {
+				serverInformation.push_back(serverInformationReply.arguments()[i].toString());
```
23rd: Can be simplified with the `for ( a : b )` loop.
For example.
```
for (const auto &arg : serverInformationReply.arguments()) {
	if (static_cast<QMetaType::Type>(arg.type()) == QMetaType::QString) {
		serverInformation.push_back(arg.toString());
```

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -7,390 +7,236 @@ For license and copyright information please follow this link:
...
-		}
-	}
-	if (copyFrom > 0) {
-		result.append(data + copyFrom, textSize - copyFrom);
-		return result;
-	}
-	return text;
-}
+NotificationData::NotificationData(const std::shared_ptr<Manager*> &guarded, const QString &title, const QString &subtitle, const QString &msg, PeerId peerId, MsgId msgId)
+: _notificationInterface("org.freedesktop.Notifications", "/org/freedesktop/Notifications", "org.freedesktop.Notifications")
```
23rd: I would suggest you to move the repeated constant values to the separate constants.
Something like this.
```
namespace {

const auto kService = QString("org.freedesktop.Notifications");
const auto kPath = QString("/org/freedesktop/Notifications");

} // namespace
```

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: fixed

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: fixed

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: fixed

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -7,390 +7,244 @@ For license and copyright information please follow this link:
...
-			GError *error = nullptr;
-			Libs::notify_notification_close(_data, &error);
-			if (!error) {
-				return true;
+	if (!capabilities.empty()) {
+		QString capabilitiesString;
+
+		for (const auto &capability : capabilities) {
+			if (!capabilitiesString.isEmpty()) {
+				capabilitiesString += qstr(", ");
```
23rd: This is a useless job for the Release build.

ilya-fedin: Maybe it is better to check is debug logs are enabled than this is a debug build? This should be very helpful if there will be issues with notifications.

23rd: > This should be very helpful if there will be issues with notifications.

Well, if this is useful for problems with notifications, we will not get this information from the users, as no one uses the Debug version. I think for a one-time log entry this will be fine to use `LOG()`.

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -7,390 +7,244 @@ For license and copyright information please follow this link:
...
 
-	if (auto capabilities = Libs::notify_get_server_caps()) {
-		for (auto capability = capabilities; capability; capability = capability->next) {
-			auto capabilityText = QString::fromUtf8(static_cast<const char*>(capability->data));
-			_capabilities.push_back(capabilityText);
-		}
-		Libs::g_list_free_full(capabilities, g_free);
+const QDBusArgument &operator>>(const QDBusArgument &argument, NotificationData::ImageData &imageData) {
+	argument.beginStructure();
+	argument >> imageData.width >> imageData.height >> imageData.rowStride >> imageData.hasAlpha >> imageData.bitsPerSample >> imageData.channels >> imageData.data;;
```
23rd:
```suggestion
	argument >> imageData.width >> imageData.height >> imageData.rowStride >> imageData.hasAlpha >> imageData.bitsPerSample >> imageData.channels >> imageData.data;
```

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -7,390 +7,244 @@ For license and copyright information please follow this link:
...
 
-	struct QueuedNotification {
-		PeerData *peer = nullptr;
-		MsgId msgId = 0;
-		QString title;
-		QString body;
-		bool hideNameAndPhoto = false;
-	};
+	if (capabilitiesReply.isValid()) {
+		return capabilitiesReply.value().toVector().toStdVector();
```
23rd: In my opinion, we can remove the blank elements from the vector right away.

```suggestion
		auto capabilities = capabilitiesReply.value();
		capabilities.removeAll(QString());
		return capabilities | ranges::to_vector;
```

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -7,390 +7,244 @@ For license and copyright information please follow this link:
...
+		QString capabilitiesString;
+
+		for (const auto &capability : capabilities) {
+			if (!capabilitiesString.isEmpty()) {
+				capabilitiesString += qstr(", ");
 			}
 
-			logError(error);
+			capabilitiesString += capability;
 		}
```
23rd: This loop can be replaced with `std::accumulate`.
For example.
```
const auto capabilitiesString = std::accumulate(
	capabilities.begin(),
	capabilities.end(),
	QString{},
	[](auto &s, auto &p) { return s += (p + qstr(", ")); }).chopped(2);
```

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
23rd: Moving such consts into an anonymous namespace is welcome.
An example from the project code.
https://github.com/telegramdesktop/tdesktop/blob/0c0c8f3584b500ef053bb1a4d9165cdbd9e18394/Telegram/SourceFiles/window/themes/window_theme.cpp#L36-L42

23rd:
By the way, all new project files follow the style with the column limit of 78.
It's quite difficult to work with long lines. =(
So, I think it would be great to split up them.

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -7,390 +7,244 @@ For license and copyright information please follow this link:
...
-		}
-		if (_capabilities.contains(qsl("body-markup"))) {
-			_markupSupported = true;
-		}
-	} else {
-		LOG(("LibNotify Error: could not get capabilities!"));
+auto NotificationDaemonRunning = false;
+bool Supported() {
+	static auto Checked = false;
+	if (!Checked) {
```
23rd: What is the reason to keep the naked `NotificationDaemonRunning` variable here? You can move it to the `Supported()` like you did with the `static auto Checked`.

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -7,390 +7,244 @@ For license and copyright information please follow this link:
...
+	if (serverInformation.size() >= 4) {
+		return QVersionNumber::fromString(serverInformation[3]);
+	} else {
+		LOG(("Native notification error: server information should have 4 elements"));
+	}
 
-	using Notifications = QMap<PeerId, QMap<MsgId, Notification>>;
-	Notifications _notifications;
+	return QVersionNumber();
+}
```
23rd: Nitpick: There is no reason to keep this function as a class method and it can be safely moved to anonymous namespace.
```
namespace {

auto ParseSpecificationVersion(const std::vector<QString> &serverInformation) {
	if (serverInformation.size() >= 4) {
		return QVersionNumber::fromString(serverInformation[3]);
	} else {
		LOG(("Native notification error: server information should have 4 elements"));
	}

	return QVersionNumber();
}

} // namespace
```

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -7,390 +7,244 @@ For license and copyright information please follow this link:
...
-	}
-#endif // !TDESKTOP_DISABLE_GTK_INTEGRATION
-	return nullptr;
-}
+void NotificationData::setImage(const QString &imagePath) {
+	QString imageKey;
+
+	if (!_specificationVersion.isNull()) {
+		const auto majorVersion = _specificationVersion.majorVersion();
+		const auto minorVersion = _specificationVersion.majorVersion();
```
23rd:
```suggestion
		const auto minorVersion = _specificationVersion.minorVersion();
```

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: As far as I understand, the point here is that the check is performed only once - at startup.
`static auto Checked` was in `Supported()` already: https://github.com/telegramdesktop/tdesktop/blob/0c0c8f3584b500ef053bb1a4d9165cdbd9e18394/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp#L267
Windows version has similar code: https://github.com/telegramdesktop/tdesktop/blob/0c713a930a40900c979205d60521532a2e67377d/Telegram/SourceFiles/platform/win/notifications_manager_win.cpp#L306-L312

Telegram/SourceFiles/platform/linux/notifications_manager_linux.h
```diff
@@ -23,16 +28,62 @@ inline bool SkipToast() {
 inline void FlashBounce() {
 }
 
-void Finish();
+class NotificationData : public QObject {
+	Q_OBJECT
```
23rd: I think we can safely remove `Q_OBJECT` to avoid an unnecessary extra MOC. 
```suggestion
```

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: Which blank elements? o_O

Telegram/SourceFiles/platform/linux/notifications_manager_linux.h
ilya-fedin: No, we can't
https://github.com/telegramdesktop/tdesktop/blob/3a0c597682890ab15b3b2c61a02adf09093c989f/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp#L66
https://github.com/telegramdesktop/tdesktop/blob/3a0c597682890ab15b3b2c61a02adf09093c989f/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp#L90

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
23rd: You have the following lines in your code.

https://github.com/telegramdesktop/tdesktop/blob/3a0c597682890ab15b3b2c61a02adf09093c989f/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp#L43-L45

So you are probably assuming that there may be blank elements?

Telegram/SourceFiles/platform/linux/notifications_manager_linux.h
23rd: Did you try?

ilya-fedin: Yes, I added Q_OBJECT only after a lot of tries without it

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
23rd: Yes, and you can see that these variables are at least in the anonymous namespace.

https://github.com/telegramdesktop/tdesktop/blob/0c713a930a40900c979205d60521532a2e67377d/Telegram/SourceFiles/platform/win/notifications_manager_win.cpp#L296-L313

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: No, it is a list of daemon capabilities for debug purposes.
It is also a port of https://github.com/telegramdesktop/tdesktop/blob/0c0c8f3584b500ef053bb1a4d9165cdbd9e18394/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp#L371

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: Okay, I'll put it in an anonymous namespace

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
23rd: Oops, I misread the condition, sorry.

Telegram/SourceFiles/platform/linux/notifications_manager_linux.h
23rd: Okay.

23rd:
Hmm, do we really need to create the `QDBusInterface` in each new `NotificationData`? Can we create a single interface and use it for all future notifications? Can the provided information be changed during the tdesktop execution?

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: fixed

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: fixed

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: fixed

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: fixed

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: fixed

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: fixed

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
ilya-fedin: fixed

ilya-fedin:
> By the way, all new project files follow the style with the column limit of 78.
> It's quite difficult to work with long lines. =(
> So, I think it would be great to split up them.
> (Except for comments with links.)

fixed

> Hmm, do we really need to create the `QDBusInterface` in each new `NotificationData`? Can we create a single interface and use it for all future notifications? Can the provided information be changed during the tdesktop execution?

I checked - it works without that. But it is still necessary to request capabilities and specification version every time, because user can replace notification daemon (or it can be replaced by system upgrade, because notification daemon doesn't works all the time and only starts upon notification request by dbus activation).

john-preston:
Thanks a lot!

PR #6826, opened 2019-11-28
Proposal:Add Chat category

annymosse:
add Chat category for more classification and will appear under Chat category on any Linux distro
more info http://standards.freedesktop.org/menu-spec/menu-spec-1.0.html#category-registry


john-preston:
Ok, thanks.

PR #6831, opened 2019-11-29
Update building-cmake.md

amosbird:
missing cd


john-preston:
Thanks, but due to moving from GYP to CMake the whole instructions will need to be updated.

john-preston:
🤔 or not.

PR #6881, opened 2019-12-16
Replace min width of 380px width 360px

jsparber:
It would be awesome to have Telegram Desktop on the [librem5](https://puri.sm/products/librem-5/). Since the app is already adaptive there isn't much to change, but the min width of 380px is slightly to wide. Therefore this changes the width to 360 which doesn't impact the functionality of Telegram Desktop, but allows it to run on the phone without shooting out of the screen.

### DIFF
--- a/Telegram/SourceFiles/window/window.style
+++ b/Telegram/SourceFiles/window/window.style
@@ -10,14 +10,14 @@ using "ui/widgets/widgets.style";
 using "history/history.style";
 using "boxes/boxes.style"; // UserpicButton
 
-windowMinWidth: 380px;
+windowMinWidth: 360px;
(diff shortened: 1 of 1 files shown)


john-preston:
@jsparber A lot of places I've taken the min width into account, I'm not sure nothing will be broken. Even most boxes are 364px.

jsparber:
> @jsparber A lot of places I've taken the min width into account, I'm not sure nothing will be broken. Even most boxes are 364px.

Hmm, i couldn't find any break with this change.

john-preston:
Ok, we can try.

john-preston:
Sorry, I'll revert this PR, many visual glitches appeared, like that one:

image

PR #6905, opened 2019-12-29
Remove windows XP-Vista support from readme

Aokromes:
(already removed on code)


john-preston:
It will be more than that, there will be "old version for Windows XP / Windows Vista, old version for OS X 10.6-10.7, old version for OS X 10.8-10.9, up to date support for Windows 7+, OS X 10.10-10.11, macOS 10.12+".

PR #6917, opened 2019-12-31
build fix

Aokromes:


john-preston:
Now CMake is used instead of GYP, I just didn't remove GYP files yet. And didn't update instructions 🙈 Sorry.

PR #6920, opened 2019-12-31
v1.9.2: Build fix

sina-ht:
Revert codegen to the commit without timestamp code.
Add new/removed source files to/from Telegram/gyp/telegram/sources.txt

Signed-off-by: Hiroshi Takekawa <sian.ht@gmail.com>

### DIFF
--- a/Telegram/gyp/telegram/sources.txt
+++ b/Telegram/gyp/telegram/sources.txt
@@ -6,6 +6,10 @@
 <(src_loc)/api/api_single_message_search.h
 <(src_loc)/api/api_text_entities.cpp
 <(src_loc)/api/api_text_entities.h
+<(src_loc)/api/api_self_destruct.cpp
+<(src_loc)/api/api_self_destruct.h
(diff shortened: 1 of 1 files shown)


john-preston:
Sorry, guys, this is not correct, I just need to update instructions to use CMake.

sina-ht:
> Sorry, guys, this is not correct, I just need to update instructions to use CMake.

Thank you for your quick reply!

I just wanted developers to know some fixes are needed.
I'd appreciate if you understood the tagged commits should be build-able.

john-preston:
@sina-ht Yeah, I see. They are buildable, but not like it is said in instructions 😭 You can try running "Telegram/configure" (.bat or .sh) and then build in "out" folder, if CMake runs successfully.

sina-ht:
> @sina-ht Yeah, I see. They are buildable, but not like it is said in instructions  You can try running "Telegram/configure" (.bat or .sh) and then build in "out" folder, if CMake runs successfully.

Thank you again!  It's a great tip.

PR #6945, opened 2020-01-02
[] lib_tgvoip.cmake: Match against x86 arches.

maxice8:
This avoids matching against other 32bit arches like armv7/armhf that
doesn't have msse2 but are 32bit and need other 32bit-specific flags to
allow for compilation like -g0.


john-preston:
Thanks!

PR #6947, opened 2020-01-02
Cmake: Also use /bin subdir for Ninja if the executable does not have…

Johnnynator:
… a suffxi (e.g. Linux)


john-preston:
Ok, thanks.

john-preston:

PR #6950, opened 2020-01-02
include cmake/varaibles.cmake only after project() is called

maxice8:


john-preston:
@maxice8 I'm afraid this is not possible, because mac os deployment target should be specified before the project is called.

maxice8:
can the build_linux32 be moved to somewhere else that can be called after ? use CMAKE_SIZEOF_VOID_P is the best option.

Or we can match CMAKE_SYSTEM_PROCESSOR and hope we don't get false positives + update whenever new arches appear.

john-preston:
And right now what works wrong? Maybe just match "64" in `uname -m`? Or it doesn't work already somewhere?

john-preston:
Ok, I'll do it your way.

PR #6980, opened 2020-01-08
Update cloud_lang.strings

jrthsr700tmax:
add Chinese (Taiwan) (zh_TW)


Aokromes:
plz, can you merge your 2 pull requests on one? :)

john-preston:
Did you check that this changes anything at all? This list is used "by country code" with two lettered code added to "cloud_lng_passport_in_" prefix, so there is no sense in adding this here.

PR #6985, opened 2020-01-09
Fixed TDESKTOP_LAUNCHER_FILENAME option

xvitaly:
Fixed TDESKTOP_LAUNCHER_FILENAME option. Missed after switching from GYP to CMake.


john-preston:
Thanks.

PR #6989, opened 2020-01-12
Renamed xdg metadata filenames to match application name

sntxerror:
XDG metadata filenames are now written as documented in freedesktop docs.

1. Fixes bug in xdg compatible DEs, when app was not listed in applications menus and finders.
2. Fixed AppStream legacy compatibility. [](url)

See [File naming](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html), [Filesystem locations](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html) and [Introduction](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html)


sntxerror:
Looks like dev build is broken now, due to edits in `../../Telegram/SourceFiles/history/view/media/history_view_video.h` location.

xvitaly:
> See File naming
> The name of the desktop entry should follow the "reverse DNS" convention: it should start with a reversed DNS domain name controlled by the author of the application, in lower case.

sntxerror:
So, you propose `org.telegram.TelegramDesktop.desktop`? Telegram Inc domains aren't controled, operated or owned by community and it looks like we can not use official domain. So while filename proposed by me matches major criteria for DBUS valid name, and while using just appname in CamelCase alone fixes the bugs in DEs I suppose it is correct to use `TelegramDesktop.desktop` as filename. 

But of course, if we do not care about domain name ownership in terms of desktop entry files I can easily add reverse domain name to the filename.

What do you think?

sntxerror:
Or as aco-author of application maybe I should use domain name controlled by me? :) Jokin.

Let's decide. Are we adding oficcial domain, which we do not control, just in case or we are happy to go with bug fixed even withoout domain?

xvitaly:
> So, you propose org.telegram.TelegramDesktop.desktop?

Currently proposed changes will break lots of things:
1. tdesktop code contains hard-coded `telegramdesktop.desktop` names (can be overrided in packaged builds by specifying `TDESKTOP_LAUNCHER_FILENAME` CMake option;
2. some modern GNU/Linux desktop environments expect launcher names from hard-coded list. For example KDE [expect this](https://github.com/KDE/purpose/blob/9fd3a91d55274860b4f6bf145af9a17302ba1003/src/plugins/telegram/telegramplugin.json#L137) launcher names: `org.telegram.desktop.desktop`, `telegramdesktop.desktop` and `telegram-desktop.desktop`.

john-preston:
I don't want to break anything :( I don't really understand what is going on here (being not a Linux distribution maintainer), but guys tell me this is not a required change.

Johnnynator:
> I don't want to break anything :( I don't really understand what is going on here (being not a Linux distribution maintainer), but guys tell me this is not a required change.

So far this change would only break KDE Purpose integration (Share menu in a few applications). I'm not aware of anything else that would brake by an arbitrary change of the .desktop file naming.

https://github.com/KDE/purpose/blob/1afcf7101db7530acf77bdee224cab18ec7b98a8/src/plugins/telegram/telegramplugin.cpp#L58-L63

xvitaly:
I implemented automatic installation of desktop file in #7046. Now maintainers can forward `-DTDESKTOP_LAUNCHER_BASENAME=org.telegram.desktop` build option and CMake will install desktop and appdata files with correct name in correct location. This PR can be closed.

john-preston:
@Johnnynator Well, I don't want this to break as well. At least without serious visible benefits for the users with accepting the change.

PR #7026, opened 2020-01-20
Use QStringList::join to print notifications capabilities

mymedia2:
This makes the code a bit simpler and compatible with Qt 5.9 since the QString::chopped method is not used.


john-preston:
Thanks.

PR #7046, opened 2020-01-22
Implemented installation support for GNU/Linux

xvitaly:
Now packaged version can be installed after build with `make install` or `ninja install`.


Telegram/CMakeLists.txt
```diff
@@ -1185,3 +1185,17 @@ if ((NOT disable_autoupdate OR NOT LINUX) AND NOT build_macstore AND NOT build_w
         set_target_properties(Packer PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${output_folder})
     endif()
 endif()
+
+if (DESKTOP_APP_USE_PACKAGED)
```
mymedia2: I think this `if` condition is unnecessary because the `install` command should not break bundled build.

Telegram/CMakeLists.txt
```diff
@@ -1185,3 +1185,17 @@ if ((NOT disable_autoupdate OR NOT LINUX) AND NOT build_macstore AND NOT build_w
...
+    install(FILES "Telegram/Resources/art/icon48.png" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/48x48/apps" RENAME "telegram.png")
+    install(FILES "Telegram/Resources/art/icon64.png" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/64x64/apps" RENAME "telegram.png")
+    install(FILES "Telegram/Resources/art/icon128.png" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/128x128/apps" RENAME "telegram.png")
+    install(FILES "Telegram/Resources/art/icon256.png" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/256x256/apps" RENAME "telegram.png")
+    install(FILES "Telegram/Resources/art/icon512.png" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/512x512/apps" RENAME "telegram.png")
+    install(FILES "lib/xdg/telegramdesktop.desktop" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/applications" RENAME "${TDESKTOP_LAUNCHER_BASENAME}.desktop")
+    install(FILES "lib/xdg/telegramdesktop.appdata.xml" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/metainfo" RENAME "${TDESKTOP_LAUNCHER_BASENAME}.appdata.xml")
+endif()
```
mymedia2: What about the `tg.protocol` file from the `lib/xdg` directory? Is it needed for KDE?

xvitaly: No. It deprecated and should not be used. KDE use mime-handlers from desktop file.

Telegram/CMakeLists.txt
xvitaly: It will not break anything until user manually run `make install`. But ofc I can add a new option for this.

john-preston:
Thanks!

PR #7062, opened 2020-01-23
Desktop environment detection fixes

ilya-fedin:
This PR
* adds flatpak to PreferAppIndicatorTrayIcon since Qt has problems with SNI implementation in flatpak
* removes Awesome since Qt icon works well with it and Awesome users use startx in general, that doesn't set XDG_CURRENT_DESKTOP/DESKTOP_SESSION anyway
* removes XFCE since Qt icon works well with it
* removes MATE since it is not GNOME and works well with Qt icon also
* adds the check for a path in DESKTOP_SESSION
* adds pavucontrol-qt to OpenSystemSettings

### DIFF
--- a/Telegram/SourceFiles/platform/linux/linux_desktop_environment.cpp
+++ b/Telegram/SourceFiles/platform/linux/linux_desktop_environment.cpp
@@ -7,6 +7,8 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 */
 #include "platform/linux/linux_desktop_environment.h"
 
+#include "platform/linux/specific_linux.h"
+
 #ifndef TDESKTOP_DISABLE_DBUS_INTEGRATION
 #include <QDBusInterface>
 #endif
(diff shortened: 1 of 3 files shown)


Telegram/SourceFiles/platform/linux/linux_desktop_environment.cpp
```diff
@@ -114,11 +115,12 @@ Type Get() {
...
 
 bool TryQtTrayIcon() {
-	return !IsPantheon() && !IsAwesome();
+	return !IsPantheon();
 }
 
 bool PreferAppIndicatorTrayIcon() {
-	return IsXFCE() || IsUnity() || IsUbuntu() ||
+	return (InSandbox() && !IsKDE()) ||
+		IsUnity() || IsUbuntu() ||
```
john-preston: Common style now is something like this:
```
return (InSandbox() && !IsKDE())
    || IsUnity()
    || IsUbuntu()
#ifndef TDESKTOP_DISABLE_DBUS_INTEGRATION
    || (IsGnome() && QDBusInterface("org.kde.StatusNotifierWatcher", "/").isValid());
#else
    || IsGnome();
#endif
```

ilya-fedin: fixed

john-preston:
Thanks! I hope it won't work worse anywhere.

PR #7078, opened 2020-01-26
Fix opening hashtag when in archive

RadRussianRus:
This PR fixes "hashtag bug".

How to reproduce:
1. Open archived chats
2. Click on any hashtag in chat or channel

What should happen:

What actually happens:

Additionally, if you click "X" button near channel or chat name, you'll get this:


john-preston:
Thanks!

PR #7080, opened 2020-01-26
Perform test for additional libatomic

mymedia2:
Telegram [uses][1] long atomic types, so at least on 32-bit ARM, MIPS, and PowerPC linking against libatomic is needed.

 [1]: https://github.com/telegramdesktop/tdesktop/blob/2c64676415a8ddc8340f9337bff8691aed47cdf2/Telegram/SourceFiles/export/output/export_output_stats.h#L28

### DIFF
--- a/Telegram/CMakeLists.txt
+++ b/Telegram/CMakeLists.txt
@@ -20,6 +20,7 @@ add_subdirectory(lib_lottie)
 add_subdirectory(lib_qr)
 add_subdirectory(codegen)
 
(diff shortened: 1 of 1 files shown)


john-preston:
@mymedia2 Wow, maybe I need to remove that, I don't think it works lock free in 32 bit windows / linux as well.

john-preston:
Anyway, thanks. Let it work that way for now.

PR #7087, opened 2020-01-27
Use QDBusConnection::connect to fix signal connection on old distributions

ilya-fedin:


john-preston:
Thanks, this fixes click-on-notifications for me.

Issue #7115, opened 2020-01-30
[Linux] [snap] Telegram tasktray icon incorrect display after update to 1.9.9


franksmcb:
### Steps to reproduce
Have snapd install Telegram Desktop 1.9.9

### Expected behaviour
Tasktray icon should display correctly with Telegram icon and show message indicator
### Actual behaviour
Tasktray icon does not display proper icon and does not message indicator

### Configuration
**Operating system:**
Ubuntu MATE 20.04, 19.10, 18.04
**Version of Telegram Desktop:**
1.9.9

With 1.9.9
![2020-01-30_10-42_1](https://user-images.githubusercontent.com/44311495/73475864-1c252b80-4389-11ea-83d4-db725f688118.png)
Before 1.9.9 
![tg-old-jpg](https://user-images.githubusercontent.com/44311495/73475865-1c252b80-4389-11ea-8036-85cc7084a7fa.jpg)

**Used theme**:

<details><summary><b>Logs</b>:</summary>
Insert log.txt here (if necessary)
</details>

samael:
Ubuntu Mate 18.04. Same problem

ilya-fedin:
Can you click tray by right button and open about dialog, then screenshot it?
Example:
![изображение](https://user-images.githubusercontent.com/17829319/73505450-72668e80-43ca-11ea-823d-d569747410a3.png)

franksmcb:
![image](https://user-images.githubusercontent.com/44311495/73508385-0455a380-4399-11ea-8b11-b1e6156f051d.png)

ilya-fedin:
This is what I suspected :(
Ubuntu MATE uses unity's indicator with a very limited support of SNI
Replacing this applet with MATE's Notification Area applet should fix the issue.

ilya-fedin:
https://bugs.launchpad.net/ubuntu/+source/sni-qt/+bug/1600136

ilya-fedin:
There are a hack in Qt, but looks like snaps doesn't have access to the process list :(
https://github.com/qt/qtbase/blob/v5.9.5/src/platformsupport/themes/genericunix/dbustray/qdbustrayicon.cpp#L196

Aokromes:
so, 3rd party bug?

ilya-fedin:
+/-, ideally ubuntu maintainers should add mate to the list in that file (and fix all qt applications in his repo/snap), but it should be possible to workaround on our side.

franksmcb:
[...]

Issue #5452, opened 2018-12-08
Opportunity to change font in tray icon popup menu manually (Gentoo, XFCE4, Awesome WM)


buriedaliveIX:
### Steps to reproduce
1.  Launch Telegram as regular
2.  Right mouse button click on Telegram tray icon
3.  Look at the some ugly font

### Expected behaviour
If there was an opportunity to change that font, but not yet

### Actual behaviour
I see very ugly font
I launched Telegram as FC_DEBUG=5 ./Telegram > ~/tg_output.log
As far as I have been able to find out this 'ugly font for me' called Becker
piece of output file tg_output.log

Substitute Edit hinting Assign true

Append list before  [marker]
Append list after  True(w)
FcConfigSubstitute editPattern has 27 elts (size 32)
	family: "Becker"(s)
	style: "Medium"(w)
	slant: 0(i)(s)
	weight: 100(i)(s)
	width: 100(i)(s)
	pixelsize: 12(f)(s)
	foundry: "unknown"(w)
	antialias: True(w)
	hintstyle: 3(i)(s)
	hinting: True(w)
	verticallayout: False(s)
	autohint: False(s)
	globaladvance: True(s)
	**file: "/usr/share/fonts/freefonts/becker.pfb"(s)**
	index: 0(i)(s)
	outline: True(w)
	scalable: True(w)

$ ls /usr/share/fonts/freefonts/
agateb.pfb    brushstr.pfb  elann___.pfb   ftpstrps.pfb  muriel.pfb    shalosti.pfb
agatei.pfb    calig___.pfb  encodings.dir  garmndmi.pfb  oldcnd_r.pfb  shf_____.pfb
agate.pfb     capri.pfb     engravrl.pfb   genoaita.pfb  oldext_r.pfb  slogan.pfb
ans.pfb       carrickc.pfb  erasblkb.pfb   genoarom.pfb  old_town.pfb  spooky.pfb
baskvlb.pfb   cheq.pfb      erasb.pfb      gocmase_.pfb  oldtwn_r.pfb  tekton.pfb
baskvli.pfb   classhvy.pfb  eraselt.pfb    gocmseq_.pfb  oswaldbl.pfb  tempofon.pfb
baskvl.pfb    cooprblk.pfb  erasltra.pfb   goudyob.pfb   parkave.pfb   tempoita.pfb
becker.pfb    coronet.pfb   erasmed.pfb    goudyoi.pfb   postanti.pfb  tribeca.pfb
[...]

Issue #4339, opened 2018-01-23
Maximize Telegram from system tray shortcut


Jacobtey:
### Steps to reproduce
1. Start Telegram
2. Press Alt+F4 to minimize Telegram to system tray
3. Can't find a shortcut to maximize Telegram from system tray

### Expected behaviour
Some shortcut to maximize Telegram from system tray

### Actual behaviour
No shortcut to maximize Telegram from system tray

### Configuration
**Operating system:**
Kernel: _4.14.14-1-MANJARO x86_64_ Desktop: _KDE Plasma 5.11.5_ Distro: _Manjaro Linux Hakoila 17.1.2_

**Version of Telegram Desktop:**
1.2.6

**Used theme**:
custom myself called rsm

<details><summary><b>Logs</b>:</summary>
Insert logs here (if necessary)

</details>

aiamuzz:
Its more than a year and no response to this request ... this would have been a useful feature !!!

I use telegram on linux ... things are so bad with me that a single left click on the telegram app icon in system tray does not bring up the application window ... the only way is to right click the systray icon and choose from the menu to bring up the application on the desktop.

Aokromes:
https://github.com/telegramdesktop/tdesktop/commit/3b4dfa138176b4d1e3cb14be1be638ebe2f52713

Issue #4059, opened 2017-11-11
LMC on tray icon act like RMC


hxss:
### Steps to reproduce
1. Close Telegram window to tray
2. Left click on Telegram tray icon

### Expected behaviour
Telegram window opens up.

### Actual behaviour
The Menu pops up as if right clicked.

### Configuration
**Operating system:**
4.13.11-1-ARCH x86_64

**Version of Telegram Desktop:**
1.1.23 Installed from [AUR](https://aur.archlinux.org/packages/telegram-desktop/).

</details>

stek29:
Please try executable from desktop.telegram.org (you can install it from AUR: telegram-desktop-bin)

hxss:
Failure while downloading https://updates.tdesktop.com/tlinux/tsetup.1.1.23.tar.xz

hxss:
@stek29 same behavior with telegram-desktop-bin

stek29:
What DE do you use?

hxss:
XFCE 4.12

haevalencia:
But that's the normal behavior of the status icon. Another thing would be if you are using a DE with support for AppIndicator or KStatusNotifier Item.

hxss:
@haevalencia this is not a normal behavior. This is first app in my tray, that show menu on left click. All other apps toggle they window(xfce4-terminal, deluge, blueberry, vlc(qt), keepassxc(qt)).

haevalencia:
@hxss I will be clearer. This is the _normal behavior_ for the Telegram Desktop status icon.
One of the reasons for GNOME developers to want to remove support from the status icon is that all applications implement it in a different way and therefore behave in a different way.

Felixoid:
What about making this behavior changeable by settings?

Felixoid:
Isn't it issue related to #3175?

hxss:
@Felixoid Yes, related. This is about one problem. But #3175 was closed without answer or solution published.

auchri:
@hxss Really?

> john-preston closed this issue in c1aa1c5 9 months ago

hxss:
@auchri What do you mean?

Felixoid:
I guess, @hxss means, that issue closed with solution

hxss:
What solution are you talking about? What I need to do to solve this problem? I'm now use v1.2.1 and LMC still open menu instead toggle window.

Felixoid:
[...]

PR #7131, opened 2020-02-03
Use LXQt's StatusNotifierItem implementation instead of appindicator

ilya-fedin:


magnus-gross:
What is the advantage of using this?

ilya-fedin:
> What is the advantage of using this?

appindicator depends on gtk and doesn't have the full support of SNI. This change fixes all the tray issues in flatpak and provides a full control on using the API, including a proper support of indicator-application-service and using of the system hidpi svg icons without converting (enables by an environment variable currently)
+ I added the use of Ui::PopupMenu for XEmbed tray, which solves the issue with theming of the tray menu

ilya-fedin:
In other words, this fixes:
https://github.com/telegramdesktop/tdesktop/issues/7115
https://github.com/telegramdesktop/tdesktop/issues/3216 (when the environment variable is set)
https://github.com/telegramdesktop/tdesktop/issues/5452 (Ui::PopupMenu uses open sans, as the whole Telegram)
https://github.com/telegramdesktop/tdesktop/issues/4059 and it's duplicate https://github.com/telegramdesktop/tdesktop/issues/4339

john-preston:
Ok, let's try that.

ugtar:
Unfortunately since this change (i think) we no longer have unread message counts in the app icon in ubuntu's (18.04) dock (or generic dash to dock), unless there is some workaround

ilya-fedin:
> Unfortunately since this change (i think) we no longer have unread message counts in the app icon in ubuntu's (18.04) dock (or generic dash to dock)

Why? :open_mouth:

ugtar:
well i could be wrong, but since upgrading to 1.9.9 i've noticed the unread badge doesn't appear in the dash icon anymore, I'm just guessing it has something to do with the change in notification mechanism

ilya-fedin:
> but since upgrading to 1.9.9

this pr was merged after 1.9.9. Do you using telegram from snap?

ugtar:
@ilya-fedin yes i use the snap. you're right it's not specifically this pr. sorry to make noise here. i thought there were several commits related to handling notifications and using qt abstractions rather than native calls in general, etc, but maybe i'm mistaken. It could be the notification badge thing is just a regular regression. in that case maybe i'll open an issue

ugtar:
i should point out though, that in terms of #7115 which this commit should fix, while i don't have an issue seeing the icon, as mentioned in that bug the unread count badge is also missing.  I'm currently running snap from the edge channel (g6cf9157fb), which includes this commit, and it doesn't seem to have fixed that at least (the badge is still missing from the icon)

ilya-fedin:
> yes i use the snap

can you tell which name has the desktop file of the snap?

ugtar:
bleh sorry, i see i misread--#7115 is talking about the task tray but i'm looking at the dash2dock icon. please excuse me i'll open an issue. sorry for the noise

PR #7135, opened 2020-02-04
Remove glib dependency from Qt

ilya-fedin:
This enables the use of [Qt's own event dispatcher](https://github.com/qt/qtbase/blob/dev/src/corelib/kernel/qeventdispatcher_unix.cpp) instead of [glib one](https://github.com/qt/qtbase/blob/dev/src/corelib/kernel/qeventdispatcher_glib.cpp)
See also: https://github.com/desktop-app/cmake_helpers/pull/20


23rd:
What about `building-cmake.md`?

PR #7146, opened 2020-02-06
Check action id when invoked

kbroulik:
Makes it more resilient


john-preston:
I hope it won't break on some non-so-common notification systems.

PR #7161, opened 2020-02-08
Respect user settings "Send by ..."

### LINKED ISSUE #6634: Respect user set "send" hotkey over entire app
At present when the Settings/ Chat setting/ Messages/ Send by... hotkey is set to Ctrl+Enter (and new line by Enter) it works only for "true" message sending while a very similar functions like Message Forward or Image Attachment annotation/comment behaves reversed to the user set settings (Enter/newline does Send).

Subj.

kirsan31:
Respect user settings "Send by ..." for:

- forward dialog
- send file dialog
- edit caption dialog
- schedule messages
- new channel dialog
- group description edit dialog 
- rate call dialog
- report bot dialog
- support mode

Fixes #6634.

Change Enter only to Enter and Ctrl+Enter in create poll dialog.

### DIFF
--- a/Telegram/SourceFiles/boxes/add_contact_box.cpp
+++ b/Telegram/SourceFiles/boxes/add_contact_box.cpp
@@ -494,6 +494,7 @@ void GroupInfoBox::prepare() {
 		_description->setInstantReplaces(Ui::InstantReplaces::Default());
 		_description->setInstantReplacesEnabled(
 			_navigation->session().settings().replaceEmojiValue());
+		_description->setSubmitSettings(_navigation->session().settings().sendSubmitWay());
 
 		connect(_description, &Ui::InputField::resized, [=] { descriptionResized(); });
 		connect(_description, &Ui::InputField::submitted, [=] { submit(); });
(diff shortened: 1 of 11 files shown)


john-preston:
Some fields were not supposed to be multiline frequently even if they support line breaks. I use Ctrl+Enter sending, but I'm pretty sure that question in polls creation should submit the entered question by Enter.

I'm also not sure about quick reply in the notifications, because there you usually write small answer and want to send it fast, also hitting Enter in an empty field just marks the message as read right now.

The rest I think will work fine with the suggested change.

kirsan31:
> Some fields were not supposed to be multiline frequently even if they support line breaks. I use Ctrl+Enter sending, but I'm pretty sure that question in polls creation should submit the entered question by Enter.
> 
> I'm also not sure about quick reply in the notifications, because there you usually write small answer and want to send it fast, also hitting Enter in an empty field just marks the message as read right now.
> 
> The rest I think will work fine with the suggested change.

May be add `Ui::InputField::SubmitSettings::Both` to polls? As for now there are `Enter` only (by default)...

john-preston:
@kirsan31 yeah, ok.

kirsan31:
Done.

john-preston:
Thanks!

PR #7180, opened 2020-02-11
Fix crash when tray icon is disabled on macOS

ilya-fedin:


Telegram/SourceFiles/platform/mac/main_window_mac.mm
```diff
@@ -556,7 +556,9 @@ void ForceDisabled(QAction *action, bool disabled) {
 			trayIcon->setContextMenu(trayIconMenu);
 		}
 	} else {
```
mymedia2: you can just remove this redundant `{ }` block, joining with `else if`

john-preston:
Thanks!

PR #7182, opened 2020-02-11
If Notifications service isn't registered, check if it could be activated

kbroulik:
DBus supports launching services on demand the first time a call is made to them.
In case the service isn't currently running, check whether it would be activated
when we try to call it.
Especially in case of session restoration the application might be launched before
the shell's notification service is fully up and running.

### DIFF
--- a/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
@@ -14,6 +14,8 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 
 #ifndef TDESKTOP_DISABLE_DBUS_INTEGRATION
 #include <QtCore/QVersionNumber>
+#include <QtDBus/QDBusConnection>
+#include <QtDBus/QDBusConnectionInterface>
 #include <QtDBus/QDBusReply>
 #include <QtDBus/QDBusMetaType>
 #endif
@@ -23,6 +25,10 @@ namespace Notifications {
 
 #ifndef TDESKTOP_DISABLE_DBUS_INTEGRATION
 namespace {
+
+constexpr auto kDBusService = "org.freedesktop.DBus"_cs;
+constexpr auto kDBusObjectPath = "/org/freedesktop/DBus"_cs;
+constexpr auto kDBusInterface = kDBusService;
 
 constexpr auto kService = "org.freedesktop.Notifications"_cs;
 constexpr auto kObjectPath = "/org/freedesktop/Notifications"_cs;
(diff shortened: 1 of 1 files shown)


Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -316,13 +322,40 @@ const QDBusArgument &operator>>(const QDBusArgument &argument,
...
-		kObjectPath.utf16(),
-		kInterface.utf16()
-	).isValid();
+	enum Status {
+		Unknown = -1,
+		NotSupported,
+		Supported
+	};
+
+	static Status status = Unknown;
```
john-preston: Please look at how it is done here:

https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/SourceFiles/platform/linux/specific_linux.cpp#L272-L294

Something like:
```
bool Supported() {
    static const auto Result = [] {
        if (blabla) {
            return true;
        }
        blabla;
        return false;
    }();
    return Result;
}
```

john-preston:
@kbroulik And this really happens and the fix works? @ilya-fedin says he has such running-on-demand service and it works already, as if isValid() call activates it.

ilya-fedin:
Yes, mate has a dbus-activable notification daemon and current code detects it.
Are there cases where this doesn't work? :thinking:

kbroulik:
Odd. With my Telegram snap (1.9.9.4) it doesn't work, it just assumes no service was found and doesn't use native notifications in this case. With a test program with `QDBusInterface` (which one should never use) and a Telegram git build it launches the service as expected.

ilya-fedin:
DE?

kbroulik:
KDE Plasma. Oh, could maybe the snap sandbox be messing with dbus activation?

ilya-fedin:
Maybe...

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -316,13 +322,30 @@ const QDBusArgument &operator>>(const QDBusArgument &argument,
...
+			return true;
+		}
+
+		// Check if service would be activated when we call it
+		QDBusMessage message = QDBusMessage::createMethodCall(kDBusService.utf16(),
+			kDBusObjectPath.utf16(),
+			kDBusInterface.utf16(),
+			QStringLiteral("ListActivatableNames"));
+
+		QDBusMessage reply = QDBusConnection::sessionBus().call(message);
```
ilya-fedin: Maybe QDBusReply is better?
https://github.com/KDE/knotifications/blob/master/src/notifybypopup.cpp#L171

kbroulik:
Ok, it appears to be a Snap sandboxing "issue".
The `Introspect` call `QDBusInterface` (which is why one should never use it) isn't allowed making the DBus-activation fail because effectively no call was ever made to the service.

PR #7229, opened 2020-02-17
Revert tray icon size to 22 on Linux (except KDE) and fix tray counter disabling in KDE

ilya-fedin:


john-preston:
Thanks!

varlesh:
worked fine :+1:

PR #7233, opened 2020-02-18
Changed libvdpau source url

RadRussianRus:
Anongit of FreeDesktop is not working now, and because of that libvdpau won't clone.

Since FreeDesktop uses GitLab instance for their projects, this PR changes libvdpau source url to their GitLab.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -147,9 +147,8 @@ jobs:
         run: |
           cd $LibrariesPath
 
-          git clone git://anongit.freedesktop.org/vdpau/libvdpau
+          git clone https://gitlab.freedesktop.org/vdpau/libvdpau.git --depth=1 -b libvdpau-1.2
           cd libvdpau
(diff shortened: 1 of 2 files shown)


.github/workflows/linux.yml
```diff
@@ -147,7 +147,7 @@ jobs:
         run: |
           cd $LibrariesPath
 
-          git clone git://anongit.freedesktop.org/vdpau/libvdpau
+          git clone https://gitlab.freedesktop.org/vdpau/libvdpau.git
```
mymedia2: There is no need to clone the entire repository. Use `--depth=1` and `-b libvdpau-1.2` for checkout to desired Git tag.

RadRussianRus: Done.

ilya-fedin: @mymedia2 Then it’s probably should be replaced everywhere for consistency.

john-preston:
Thanks!

PR #7257, opened 2020-02-22
Fix builds for linux

mymedia2:
* Remove intermediate build directories in Linux workflow.
 * Use fixed version v3.16.3 of CMake in Snap workflow.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -170,7 +170,7 @@ parts:
   cmake:
     source: "https://gitlab.kitware.com/cmake/cmake.git"
     source-depth: 1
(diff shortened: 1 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -170,7 +170,7 @@ parts:
   cmake:
     source: "https://gitlab.kitware.com/cmake/cmake.git"
     source-depth: 1
-    source-branch: master
+    source-branch: v3.16.3
```
ilya-fedin: This is a tag, not a branch

john-preston:
Thanks!

PR #7270, opened 2020-02-24
Enable native Wayland support

### LINKED ISSUE #2353: Telegram can't run natively on Wayland
### Steps to reproduce
1. Launch Telegram Desktop under a Wayland session.
2. Use any method to see if it's running via XWayland or on Wayland natively, e.g. [this one](https://medium.com/@bugaevc/how-to-easily-determine-if-an-app-runs-on-xwayland-or-on-wayland-natively-8191b506ab9a#.1cz6e64aq).
### Expected behaviour

Telegram should use Wayland natively, as Qt supports Wayland backend.
### Actual behaviour

Telegram uses X, so it has to be forwarded via XWayland, which causes performance and compatibility issues.
### Configuration

**Operating system:** Fedora 24, GNOME on Wayland session

**Version of Telegram Desktop:** 0.10.1
### Logs

Manually asking Telegram to run on Wayland results in an error:

```
$ QT_QPA_PLATFORM=wayland-egl /opt/telegram/Telegram 
This application failed to start because it could not find or load the Qt platform plugin "wayland-egl"
in "".

Available platform plugins are: xcb.

Reinstalling the application may fix this problem.
<traceback>
```

Suggesting that the bundled Qt is only built with X support.

Relevant Qt docs: https://wiki.qt.io/QtWayland

ilya-fedin:
See also https://github.com/desktop-app/cmake_helpers/pull/31
Fixes #2353


john-preston:
Thanks!

PR #7285, opened 2020-02-25
Fix escaping in linux launcher creating

ilya-fedin:


Telegram/SourceFiles/settings/settings_codes.cpp
```diff
@@ -113,7 +113,7 @@ auto GenerateCodes() {
 		});
 	});
 	codes.emplace(qsl("registertg"), [](::Main::Session *session) {
-		Platform::RegisterCustomScheme();
+		Platform::RegisterCustomScheme(true);
 		Ui::Toast::Show("Forced custom scheme register.");
 	});
```
23rd: Perhaps it would be better to add the `#ifndef TDESKTOP_DISABLE_REGISTER_CUSTOM_SCHEME` condition.
```suggestion
#ifndef TDESKTOP_DISABLE_REGISTER_CUSTOM_SCHEME
	codes.emplace(qsl("registertg"), [](::Main::Session *session) {
		Platform::RegisterCustomScheme(true);
		Ui::Toast::Show("Forced custom scheme register.");
	});
#endif // !TDESKTOP_DISABLE_REGISTER_CUSTOM_SCHEME
```

ilya-fedin: Fixed

john-preston:
Thanks!

PR #7321, opened 2020-03-03
Fix notification daemon detection on KDE

### LINKED ISSUE #5590: Native Notification option gone if Telegram autostarts
### Steps to reproduce
1. Autostart Telegram
2. Receive a message

### Actual behaviour
Telegram draws it's own notification instead of using the native notification system. 
Also, the native notification toggle in the Telegram settings is gone.

### Expected behaviour
Telegram should continue to use native notifications if that option was previously set.
The option should not disappear either.

### Workaround
Fully quit and restart telegram manually.

### Configuration
**Operating system:** Antergos Linux (Arch Based)

**Desktop Environment:** KDE Plasma 5.14.5

1.5.4

**Used theme**:
On Telegram?
Arc-2 Bright

### LINKED ISSUE #7308: There is no option "Use native notifications" in KDE if the Telegram is loaded before the Plasma because of Autostart (Racing condition)
There is no option "Use native notifications" in KDE if the Telegram is loaded before the Plasma because of Autostart (Racing condition)

Telegram 1.9.19 beta
Linux OpenSUSE Tumbleweed
KDE 5

Steps to reproduce:
killall plasmashell
run telegram
run plasmashell

Possible solution:
Check the running Plasma when the user opens the Settings/Notifications window.

image
image

ilya-fedin:
Fixes #5590

### DIFF
--- a/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
@@ -143,13 +143,9 @@ NotificationData::NotificationData(
 	if (capabilities.contains(qsl("actions"))) {
 		_actions << qsl("default") << QString();
 
-		_dbusConnection.connect(
-			kService.utf16(),
-			kObjectPath.utf16(),
-			kInterface.utf16(),
-			qsl("ActionInvoked"),
-			this,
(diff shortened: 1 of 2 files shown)


john-preston:
Thanks!

PR #7340, opened 2020-03-07
Convert TDESKTOP_DISABLE_DESKTOP_FILE_GENERATION to runtime environment variable

### LINKED ISSUE #306: [Linux] Stop creating ".desktop" file automatically!

ilya-fedin:
This removes TDESKTOP_DISABLE_DESKTOP_FILE_GENERATION macro that has no sence since schema registration won't work without desktop file generation.
Inested, added an environment variable with the same name which closes #306


john-preston:
Thanks!

PR #7351, opened 2020-03-10
Minimize dbus requests

ilya-fedin:


Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -94,7 +106,14 @@ bool Inhibited() {
...
 		return reply.value().toBool();
-	} else if (reply.error().type() != QDBusError::InvalidArgs) {
+	} else if (reply.error().type() == QDBusError::ServiceUnknown
+		|| reply.error().type() == QDBusError::InvalidArgs) {
+		InhibitedNotSupported = true;
+	} else {
+		if (reply.error().type() == QDBusError::AccessDenied) {
+			InhibitedNotSupported = true;
+		}
+
```
23rd: Didn't test.
```suggestion
	}
	constexpr auto dbusErrors = {
		QDBusError::ServiceUnknown,
		QDBusError::InvalidArgs,
		QDBusError::AccessDenied,
	};
	const auto error = reply.error();
	InhibitedNotSupported = ranges::contains(dbusErrors, error.type());
	LOG(("Native notification error: %1").arg(error.message()));
```

ilya-fedin: I'll try. But with QDBusError::ServiceUnknown and QDBusError::InvalidArgs there are shouldn't be any error, since it is means that really unsupported.
AccesDenied, through, means that there are some sandbox and InhibitedNotSupported is used to prevent log flooding, otherwise errors will be on each notifification.

ilya-fedin: It works, have added to PR

john-preston:
Thanks!

PR #7365, opened 2020-03-12
Set a list of architectures that snap package runs on

ilya-fedin:


ilya-fedin:
Looks like a bug in snapcraft :see_no_evil:

ilya-fedin:
Okay, then without `build-error: ignore`

snap/snapcraft.yaml
```diff
@@ -6,6 +6,13 @@ base: core18
 grade: stable
 confinement: strict
 
+architectures:
+  - build-on: amd64
+  - build-on: i386
+  - build-on: arm64
+  - build-on: armhf
+  - build-on: ppc64el
```
23rd: Will it work this way?
```suggestion
architectures: [amd64, i386, arm64, armhf, ppc64el]
```

ilya-fedin: No. Look at example 6:
https://snapcraft.io/docs/architectures

23rd: Okay.

john-preston:
Thanks!

PR #7367, opened 2020-03-12
Add tag `launchable` into linux appdata

ilya-fedin:
This fixes icon displaying in some linux store apps and info updating on flatpak site (currently flatpak maintainer fixed it with downstream patch)


john-preston:
Thanks.

PR #7370, opened 2020-03-12
Add TDESKTOP_FORCE_PANEL_ICON environment variable for possibility to use system icon even if it is not detected

ilya-fedin:


john-preston:
Thanks!

PR #7378, opened 2020-03-14
Move all global menu code into TDESKTOP_DISABLE_DBUS_INTEGRATION

ilya-fedin:


john-preston:
I hope I've fixed it. Otherwise one more commit will be necessary.

PR #7379, opened 2020-03-14
Add DESKTOP_APP_DISABLE_DBUS_INTEGRATION to matrix of Linux GitHub action defines

ilya-fedin:


23rd:
Yeah, I have thought about that too since the code has a ton of exceptions for Linux.

PR #7383, opened 2020-03-14
Fallback to launcher basename if FLATPAK_ID variable is empty

ilya-fedin:


Erick555:
You may just hardcode `org.telegram.desktop` instead or relying on `FLATPAK_ID` being available or not.

PR #7420, opened 2020-03-20
bugfix: incorrect erase element

TrueBober:


john-preston:
Thanks! This is a serious bug.

PR #7469, opened 2020-03-28
Add AV1 support to snap

ilya-fedin:


snap/snapcraft.yaml
```diff
@@ -191,13 +191,35 @@ parts:
       - libtinfo5
     prime: [-./*]
 
+  nasm:
+    source: https://repo.or.cz/nasm.git
```
23rd: Sadly `repo.or.cz` is quite unstable for CI. =(

ilya-fedin: I can't find any mirror, but there are tarballs: https://www.nasm.us/pub/nasm/releasebuilds/2.14.02/

PR #7487, opened 2020-03-29
set msbuild platform to x86

jeffythedragonslayer:
On my 64-bit machine it tries to look for a x64 platform and fails, this forces it to x86 and build succeeds.  I'm having other build problems with NASM at the moment though so I'm not sure if this needs to be applied in more places.


john-preston:
It should be that way by default if you launch the commands from the correct command prompt (x86 Native Tools Command Prompt).

PR #7593, opened 2020-04-08
Make Telegram Desktop usable on phones

a-wai:
Phones running Linux, such as the Librem5 and PinePhone, use a 360x720
resolution for the display. This causes Telegram Desktop to overflow by
20px, which leads to usablility issues.

Setting the minimum width to 360px is enough to make it scale properly
and retain full usability.

### DIFF
--- a/Telegram/SourceFiles/window/window.style
+++ b/Telegram/SourceFiles/window/window.style
@@ -10,7 +10,7 @@ using "ui/widgets/widgets.style";
 using "history/history.style";
 using "boxes/boxes.style"; // UserpicButton
 
-windowMinWidth: 380px;
+windowMinWidth: 360px;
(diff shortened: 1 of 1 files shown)


Aokromes:
this was added before and reverted because it caused problems.

Aokromes:
i may suggest you to use sub 100% scaling from commandline.

john-preston:
@a-wai Indeed, you may try 95% scale factor (or a bit lower), it will make minimal window width 361px.

I tried to change it to 360px without any other changes and there appeared a lot of visual glitches on 360px window.

a-wai:
Alright, thanks for taking the time to reply, setting the scale to 90-95% does the trick indeed.

PR #7603, opened 2020-04-09
Update refresh.py

andrew000:


john-preston:
@andrew000 Why do you want to update it with such a minor change? No one even uses this script.

Issue #7611, opened 2020-04-10
Windows Notifications Pop-Up While in Full-Screen Mode


karkason:
### Steps to reproduce
1. Open a YouTube video in full screen.
2. Receive a message in Telegram Desktop.

### Expected behaviour
There shouldn't be any pop-up that interferes with the full-screen content.

### Actual behaviour
There is a pop-up.

### Configuration
**Operating system:** Windows 10.

**Version of Telegram Desktop:** 2.0.1

**Used theme**: Normal.

It's important in particular during presentations when you don't want the notification to pop up.

MasterGroosha:
You can right-click Telegram icon in Windows tray and click "Disable notifications", which will disable them on this app alone. It's like "Do not disturb" mode which won't affect Telegram on your other devices.

Aokromes:
https://github.com/telegramdesktop/tdesktop/issues/2290

PR #7612, opened 2020-04-10
Windows - Not showing notifications if currently in Full Screen mode …

karkason:
…(i.e. presentation or youtube video)

Resolves my #7611 issue :)

### DIFF
--- a/Telegram/SourceFiles/platform/win/notifications_manager_win.cpp
+++ b/Telegram/SourceFiles/platform/win/notifications_manager_win.cpp
@@ -455,6 +455,19 @@ void Manager::Private::clearNotification(PeerId peerId, MsgId msgId) {
 	}
 }
 
+namespace {
+bool isFullScreen() {
+	QUERY_USER_NOTIFICATION_STATE userState = {};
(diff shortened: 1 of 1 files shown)


ilya-fedin:
There are already SkipToast and SkipAudio methods

john-preston:
@karkason But it is already done here:

https://github.com/telegramdesktop/tdesktop/blob/1e5aa2a78d0a0ba06d95c63e746d7b795fe838b8/Telegram/SourceFiles/platform/win/notifications_manager_win.cpp#L714-L724

If you've checked your changes, can you check why they work differently from that is already there in SkipToast method?

karkason:
> @karkason But it is already done here:
> 
> https://github.com/telegramdesktop/tdesktop/blob/1e5aa2a78d0a0ba06d95c63e746d7b795fe838b8/Telegram/SourceFiles/platform/win/notifications_manager_win.cpp#L714-L724
> 
> If you've checked your changes, can you check why they work differently from that is already there in SkipToast method?

You’re right. I’ll look into it!

john-preston:
Please reopen if there is new information.

PR #7623, opened 2020-04-12
Fixes for linux action:

ilya-fedin:
* Disable building of unneeded openal tools and tests
* Disable ffmpeg linkage with unneeded libraries
* Disable unneeded dtd validation for libwayland
* Omit Qt flags that set to default values
* Fix prefix usage
* Build dependencies in release mode to reduce build size


john-preston:
Thanks!

PR #7629, opened 2020-04-13
Use OpenAL without direct channels

ilya-fedin:


john-preston:
Ok, let's check this.

Issue #5840, opened 2019-03-25
Some ogg/opus files are not recognized as "audio files" 


strannik-j:
### Steps to reproduce
1. Send ogg or opus audio file more than 10 Mb size to chat.
2. Check what category that file recieved.
3. Check if file was placed inside player playlist or played as singular file.

### Expected behaviour
File placed in "audio files" category as well as added to playlist.

### Actual behaviour
Some audio files are categorized as just files.

### Configuration
**Operating system:** Manjaro Linux x64

**Version of Telegram Desktop:** 1.6.2

**Used theme**: Default

Example of file that was categorized as just "file":
[test.ogg](https://yadi.sk/d/GbKflB4lZNmixA)
[test.opus](https://yadi.sk/d/HEX93dtw_DvAhA)

rackelhahn69:
I've got the same problem for mp3 files as well. The problem has been persisting for quite a while. Sometimes it works on my laptop, but not on my desktop even though both are running the most up-to-date version of Manjaro x64 and telegram-desktop from the community repository. I think there is a package/are packages missing...

Nachtalb:
There are hundreds of audio codecs available. To support every single one is not easy. And eg. `ogg` is a container and not a codec or stream. The ogg file which you provided contains a video and an audio stream. The video stream is corrupted though.

I have removed the video stream from your file and voila it works as an audio file in Telegram.

![image](https://user-images.githubusercontent.com/9467802/56358275-7b40f780-61de-11e9-8a10-8e8d86d5076a.png)
The second one is the fixed one.

Your original `test.ogg`
```
$ ffmpeg -i test.ogg -hide_banner
[ogg @ 0x55accd891880] 8192 bytes of comment header remain
[ogg @ 0x55accd891880] invalid pts -104
Input #0, ogg, from 'test.ogg':
  Duration: 00:08:24.27, start: 0.000000, bitrate: 215 kb/s
    Stream #0:0: Audio: vorbis, 44100 Hz, stereo, fltp, 224 kb/s
    Metadata:
      ARTIST          : Ария
      ALBUM           : Мания Величия
      track           : 03
      TITLE           : Волонтёр
      GENRE           : Heavy Metal
      DATE            : 1985
    Stream #0:1: Video: mjpeg, yuvj444p(pc, bt470bg/unknown/unknown), 300x300 [SAR 72:72 DAR 1:1], 90k tbr, 90k tbn, 90k tbc
    Metadata:
[...]

Issue #1754, opened 2016-03-10
Add playback support for more audio formats


diazbastian:
Tested on:
Telegram Desktop 0.9.20 dev
Telegram for Android 3.6.1 (753)
Telegram for iOS 3.6

The following audio formats works for Telegram for mobile devices(Android/iOS) and not for TDesktop:
- PCM/WAV
- MKA container (vorbis/aac codec) - Perhaps other codecs already supported like flac, mp3, etc (no tested)
- WMA (normal and lossless)

Cheers

susnux:
And also tdesktop depends on opus and ffmpeg but it can not play opus audio files.
Adding this would be awesome!

diazbastian:
@susnux 
I also use opus codec to my songs, however, Telegram uses opus codec for encoding voice messages. I think that would cause a conflict. I prefer support is of the same audio formats that Android and iOS.
Regards

haevalencia:
:+1:  I keep my recordings on opus from other apps, but can't be played on telegram desktop by technical issues not much to do.

haevalencia:
Today ask directly to the TSI and they said Telegram supports the audio formats supported by the mobile OS. Maybe add the codec and container comparable on both platforms?

stephaaan:
Can confirm that MP3 doesn't work as well - not played (it only tries to and fails). Tried with multiple files and samples :(

haevalencia:
@telegramdesktop What is the status of analysis to add more support for other audio formats as said in #1745 ?
I think the new version of Telegram include an integrated video player. It would be good that at least Telegram Desktop equated support for playing audio files. Cheers.

PeterMX:
I want support for opus, I have a lot of my music in opus format

diazbastian:
@PeterMX audio using opus codec with *.ogg extensión work fine on in-app audio player

Socialdarwinist:
After encoding a short clip I wanted to use as a meme, I only succeeded after several tries because of Telegram’s limited support – using it on Arch Linux.
Matroška and WebM were not playable at all inside the Telegram client, just MP4, and when I uploaded MP4, the sound did not play if it wasn’t encoded in AAC, for example if encoded with Opus or libav AC-3. It should be a thing to use MKV and WebM, they are only container formats (leaving the same file size for the resulting file), and AC-3 is allegedly not covered by patents since 2017, and Opus is from the free software environment.

tdesktop-closer:
Hey there!

[...]

PR #7664, opened 2020-04-19
Add .opus and .oga files as song formats

VictorienXP:
Telegram can already play .ogg Vorbis and Opus files correctly on Desktop and Android.
But .opus and .oga files sent from the Desktop app won't get marked as audio files.

This pull request will make any .opus and .oga files actually marked as music files that can be read from Telegram directly.

I've tested it and files sent from the patched client will work correctly on the current Desktop app and Android app.

Annotation 2020-04-19 185426

Screenshot_20200419-185530

Related issues: #5840 #1754


john-preston:
Ok, thanks. I hope they play on iOS as well.

PR #7726, opened 2020-04-26
Restore setFamily in crash report window

ilya-fedin:
Since setFamily in QApplication::setFont was removed


john-preston:
Thanks!

PR #7752, opened 2020-04-29
Updated supported systems

balboag:
XP and Vista are no longer supported. Win 7 is the min version.

Currently there are two installers for Mac: 
- macOS 10.12 - 10.15 
- Mac OS X 10.10 - 10.11

Since 10.12, the Mac system started to be called **macOS**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,9 @@ The source code is published under GPLv3 with OpenSSL exception, the license is 
 
 ## Supported systems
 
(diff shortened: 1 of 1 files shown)


john-preston:
Well, there are legacy builds for Windows XP / Vista, for OS X 10.6-10.7 and for OS X 10.8-10.9 :)

john-preston:
Sorry, I've updated the README with the correct data with links now.

PR #7759, opened 2020-04-30
Exclude files that aren't necessary for actions

RadRussianRus:
This PR excludes some files in actions to reduce unnecessary builds.


john-preston:
Thanks.

PR #7779, opened 2020-05-02
Fix decoration applying and trigger repainting on update

ilya-fedin:
For some reason this is needed for newer Qt in flatpak


john-preston:
Thanks!

PR #7783, opened 2020-05-02
Do not ignore changes for docs needed for build

RadRussianRus:
Since commits checkout are taken from the docs, docs should not be ignored by actions.


john-preston:
Thanks.

PR #7799, opened 2020-05-04
Move "X groups in common" in shared media to bottom

### LINKED ISSUE #5476: "Groups in common" should be placed after all the shared media items
### Steps to reproduce
1. open a profile of an user you have some groups in common with and also shared voice messages

### Expected behaviour
"groups in common" should be the last item, placed after all the media types since it's a bit different

### Actual behaviour
["groups in common" is placed before "voice messages"](https://i.imgur.com/fuscLJS.jpg)

### Configuration
**Version of Telegram Desktop:** 1.5.1

RadRussianRus:
This PR moves "X groups in common" button in profile after media types, since "common groups" is not a media type itself. Fixes #5476.

Before:

After:


john-preston:
Thanks.

PR #7810, opened 2020-05-06
Remove old `cd` to platforminputcontexts plugins

RadRussianRus:
This PR fixes CMake docs. There is no need to clone hime, nimf and fcitx to `qtbase/src/plugins/platforminputcontexts`, and `cd` to this folder was removed in CI, but wasn't removed in docs.


john-preston:
Sorry, you were second.

RadRussianRus:
Doesn't matter, glad it's fixed now.

PR #7822, opened 2020-05-06
feat: add setting checkbox to change messages width wide/tight

### LINKED ISSUE #2060: Allow to resize message block in chat window

kudlatyamroth:
Fixes #2060

This PR add setting checkbox in Chat Settings section.
If checkbox is checked, messages in chat doesn't have limit on width beside chat own width.
If checkbox is unchecked, messages width behave like now.

The default is unchecked to preserve default settings.

This is how it looks like when it is enabled:
chat_settings

chat_messages

P.S. If something is wrong, sorry because my last time with C++ was 15 years ago... Now it looks strange to me...

### DIFF
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -359,6 +359,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 "lng_settings_bg_show_all" = "Show all themes";
 "lng_settings_bg_tile" = "Tile background";
 "lng_settings_adaptive_wide" = "Adaptive layout for wide screens";
+"lng_settings_wide_msg_width" = "Wide message width";
 
 "lng_settings_section_call_settings" = "Calls Settings";
 "lng_settings_call_section_output" = "Speakers and headphones";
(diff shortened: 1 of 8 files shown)


kudlatyamroth:
updated to latest dev :)

danwalmsley:
This is a much needed feature, probably off by default, but much needed for dev teams that use telegram

john-preston:
I've done something similar for the case from your screenshot but without additional settings.

Note that per [contributing guidelines](https://github.com/telegramdesktop/tdesktop/blob/dev/.github/CONTRIBUTING.md#what-contributions-are-accepted) we generally don't merge PRs that introduce new UI elements, including settings.

With the current solution done in https://github.com/telegramdesktop/tdesktop/commit/07e3671ca8e61e2d9a972c19c520d16077b89aa0 the monospace blocks (both `pre` and `code` if they take the whole paragraph) may extend the message bubble beyond st::maxMsgWidth limit.

kudlatyamroth:
@john-preston but PRs with disabling st::maxMsgWidth for all messages were declined, so i hope so that this with config will have more chance to be merged.
As we know that mac client don't have this limit...
Wouldn't be great to have that option, so users can choose if they want full width or limited?

grokys:
It's a shame this wasn't accepted. I'd really like messages to take more than ~15% of my screen width:

image

kudlatyamroth:
@grokys if you want i created AUR package: https://aur.archlinux.org/packages/telegram-desktop-wide/

grokys:
Thanks @kudlatyamroth but I'm on Windows. Maybe I'll have to do my own build :/

Issue #1548, opened 2016-01-23
Telegram causes music in rhythmbox (gstreamer applications) to play faster!?


cookiengineer:
When I startup telegram, the music in my applications play around 1.5 times the usual speed. I have not yet any kind of hint what causes the bug to appear, I just know that it _always_ happens when starting Telegram.

Hardware used: Intel NUC (i5 and Intel Audio dedicated hardware, no additions).
OS used: UbuntuGNOME 15.10

Does anybody have an idea how to debug such a thing? Is it most likely pulseaudio-related?

cookiengineer:
I can now reproduce the bug, but I have no clue how to figure out what causes it. strace is so big that it's a couple hundred megabytes big with all the pulseaudio stuff going on.

Steps to reproduce (any Ubuntu with GNOME or GTK3 installed (also XFCE)):

- Start rhythmbox or any other gstreamer media player
- Start playing back a music track
- Open telegram binary

This will crash the pulseaudio frequency (or timer!?), so track will be played back around 1,5x times faster.

Fixes:

- When Bug occurs, go to top right corner in GNOME shell
- Open Settings from there
- Open Sound Settings

Now the bug is fixed. Can be reproduced afterwards following the steps above (telegram must be closed and re-opened, also from the systray bar).

tdesktop-closer:
Hey there!

We're automatically closing this issue since there was no activity in this issue since 372 days ago. We therefore assume that the user has lost interest or resolved the problem on their own. Closed issues that remain inactive for a long period may get automatically locked.

Don't worry though; if this is in error, let us know with a comment and we'll be happy to reopen the issue.

Thanks!

<sub>(Please note that this is an [automated](https://github.com/auchri/github_auto_closer) comment.)</sub>

cookiengineer:
This bug has not been magically fixed. It is still active and still telegram is causing change of frequency on devices that have not 44kHz by default.

stek29:
Is this Telegram issue for sure?
Also, can you test other OpenAL based apps?

cookiengineer:
I've been using the same identical setup for over 5 years now. It is telegram, for absolutely nothing-else-breaks sure. Every other application, even games, work perfectly without any issue.

Telegram keeps always resetting the frequency of the sound device, which it should not do. Eversince I reported this bug, this issue stays the same and is still causing troubles for devices that have no 44.1kHz frequency.

[...]

PR #7826, opened 2020-05-07
Check only if at least one audio device is exist on startup

ilya-fedin:
This makes https://github.com/telegramdesktop/tdesktop/issues/1548 don't affect on startup, but only when capture feature is used


john-preston:
Thanks.

PR #7828, opened 2020-05-07
Update README.md

QuestYouCraft:
view changelog.txt


damnfella:
Hi, could you add a note about support of older systems until a certain tdesktop version, so that they could use that one, if at all they need to?

QuestYouCraft:
Yes, I could.

john-preston:
Sorry, I've updated the README with the correct data with links now.

PR #7864, opened 2020-05-10
Fix call window hiding when compositing is not supported

### LINKED ISSUE #6675: The Call Pop-up does not get hidden/closed on Wayland after accepting a call
### Steps to reproduce
1. Run Telegram on a Wayland compositor (only tested it with GNOME 3.34) with the environment variable `QT_QPA_PLATFORM=wayland` set;
2. Make a call;
3. When the call starts, click on other app window so that Telegram loses focus.

### Expected behaviour
The call pop-up should disappear after losing focus.

### Actual behaviour
The call pop-up does not completely disappear after losing focus, it is still visible on the workspace overview (KDE and GNOME have this feature where all active workspaces and opened windows on them are shown at once), and most likely, at the same time, positioned below a currently active window.

### Configuration
**Operating system:** Fedora Silverblue 31

**Version of Telegram Desktop:** 1.8.14 via [Flatpak](https://github.com/flathub/org.telegram.desktop)

**Used theme:** The vanilla Night theme

### LINKED ISSUE #7862: Closing context popup while sub-context menu is open causes Telegram to crash
### Steps to reproduce
1. Select text in the message box and right click it;
2. Hover over either "Formatting" or "Spelling" to open a corresponding sub-context menu;
3. Move the mouse cursor to Telegram window, without folding the sub-context menu, and press the left mouse button.

### Expected behaviour
The context popup should disappear. 

### Actual behaviour
The `telegram-desktop` process crashes and dumps core, and `The Wayland connection experienced a fatal error: Protocol error` is printed out when the client is run from a CLI. 

Clicking outside Telegram window closes the context popup, without triggering a crash.

### Configuration
**Operating system:** Fedora Silverblue 32 (GNOME 3.36 on **Wayland**)

**Version of Telegram Desktop:** 2.14

**Installation source (Linux Only)** - `flatpak` (+ `$ sudo flatpak override --env=QT_QPA_PLATFORM= org.telegram.desktop` to enable the **Wayland** backend for the flatpak)

**Used theme**: The vanilla night theme

```
May 10 16:49:54 fedora systemd-coredump[42337]: Process 42230 (telegram-deskto) of user 1000 dumped core.
                                                
Stack trace of thread 2:
#0  0x00007f055c05d605 n/a (/usr/lib/x86_64-linux-gnu/libc-2.30.so + >
#1  0x0000003000000018 n/a (n/a + 0x0)
```

</details>

ilya-fedin:
Fixes #6675

### DIFF
--- a/Telegram/SourceFiles/calls/calls_panel.cpp
+++ b/Telegram/SourceFiles/calls/calls_panel.cpp
@@ -463,6 +463,8 @@ void Panel::toggleOpacityAnimation(bool visible) {
 			_visible ? 1. : 0.,
 			st::callPanelDuration,
 			_visible ? anim::easeOutCirc : anim::easeInCirc);
+	} else if (!isHidden() && !_visible) {
+		hide();
 	}
 	if (isHidden() && _visible) {
 		show();


john-preston:
Thanks!

PR #7893, opened 2020-05-14
Add possibility to control external upater flag with a config in /etc

ilya-fedin:


Telegram/SourceFiles/core/launcher.cpp
```diff
@@ -98,13 +98,29 @@ void ComputeDebugMode() {
...
 
 void ComputeTestMode() {
-	if (QFile(cWorkingDir() + qsl("tdata/withtestmode")).exists()) {
+	if (QFile::exists(cWorkingDir() + qsl("tdata/withtestmode"))) {
 		cSetTestMode(true);
 	}
 }
 
+void ComputeExternalUpdater() {
+	QFile file(qsl("/etc/tdesktop/externalupdater"));
```
mymedia2, line 107: It would be nice to get rid of hard-coded path. What about `QStandardPaths::AppConfigLocation`?

ilya-fedin: Then it would be `/etc/xdg/Telegram Desktop`, not a good option

PR #7896, opened 2020-05-14
Disable fallback session management

### LINKED ISSUE #948: Add option "Start in system tray" in "Settings" window

ilya-fedin:
Fixes #948


john-preston:
Thanks!

PR #7898, opened 2020-05-15
Fix call window size and notification positioning on Sway

### LINKED ISSUE #7591: Non-native notifications don't work on tiling window managers (Linux only)
Telegram re-implements notifications rather than use the OS one, but on Linux, these don't render properly on tiling window managers.

### Steps to reproduce
1. On a clean installation, just ask someone to send you a message.

### Expected behaviour
Ideally, Telegram should just use the OS's notifications. There's really no reason to override the user's established preference by default.

Regardless of the default, the notifications should render properly on tiling window managers.

### Actual behaviour
Notifications are a full-sized window, with their content inside a corner of this otherwise transparent window.

There were two windows before taking this screenshot, each taking up 50% screen space. Telegram's notification window is a third window, so now they each take up 33% scree space. You might get the impression that there's empty space in the middle -- the window is actually taking up all of that space, but only drawing on parts of it.

2020-04-07T19:36:59,087346415+02:00

For the following screenshot, I had my window layout set to tabbed, at which point, Telegram created a new window which, again, takes up as much space as it can (the background is just a transparent non-clickthrough area of this notification window).

2020-04-07T19:37:04,175094266+02:00

Here's a final screenshot, capturing *only* the space occupied by the currently active window:

2020-04-07T19:36:49,818025566+02:00

### Configuration
**Operating system:** ArchLinux, running swaywm as a tiling window manager.
**Version of Telegram Desktop:** 2.0.1
**Used theme**: Tinted

### LINKED ISSUE #7595: Call window won't render properly on Linux
### Steps to reproduce
1. Initiate a call

### Expected behaviour
A normal window that can adapt its content to its size should be shown. If something with fixed-size is expected, it should probably just be part of the Telegram main window (like audio clips).

### Actual behaviour
Telegram creates a strange window; it doesn't seem to hint to the WM that it's a modal window, so it's treated as a normal window. On a tiling WM, this means it'll grow to occupy all available space:

2020-03-25T21:18:53,937572672+01:00

If I make if a floating window, it apparently takes the size hints, and grows smaller. However, note that the encryption-verification emojis don't show -- they're left outside the window's visible space:

2020-03-25T21:20:22,597470151+01:00

It's very likely that the issue is due to some hints missing -- floating window managers are a bit more forgiving with these, but tiling window managers behave quite differently depending on them.

It should be kept in mind that actually hinting that this should be a modal window would be problematic too -- modal windows are usually quickly dismissed (eg: a confirmation or file picker), but this window wouldn't, have having it float on top of the screen for the duration of the call means it covers up whatever other app one tries to use during the call.

### Configuration
Operating system: ArchLinux, running swaywm as a tiling window manager.
Version of Telegram Desktop: 2.0.1
Used theme: Tinted

ilya-fedin:
Fixes #7591
Fixes #7595

### DIFF
--- a/Telegram/SourceFiles/calls/calls_panel.cpp
+++ b/Telegram/SourceFiles/calls/calls_panel.cpp
@@ -299,7 +299,8 @@ QImage Panel::Button::prepareRippleMask() const {
 }
 
 Panel::Panel(not_null<Call*> call)
-: _call(call)
+:RpWidget(App::wnd())
+, _call(call)
 , _user(call->user())
 , _answerHangupRedial(this, st::callAnswer, &st::callHangup)
 , _decline(this, object_ptr<Button>(this, st::callHangup))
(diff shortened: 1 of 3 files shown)


WhyNotHugo:
Arifacts aren't available for this branch?
(wanted to quickly test how the notifications behave and provide some feedback).

ilya-fedin:
> Arifacts aren't available for this branch?

I already sent you an artifact in the issue yesterday, it was compiled from this branch

WhyNotHugo:
Notifications don't really work as expected: they show up in the middle of the screen (each covering the previous one).

They also grab keyboard focus immediately.

ilya-fedin:
> Notifications don't really work as expected: they show up in the middle of the screen (each covering the previous one).

Yeah, that's all i can do

john-preston:
Thanks!

WhyNotHugo:
Thanks for addressing this :)

PR #7962, opened 2020-05-29
Update building-msvc.md

SlavikMIPT:
Using old version of OpenAL causes aliasing distortion in high-frequency band. I've built the client using v1.19 version of OpenAL
Check out my [investigation](https://telegra.ph/Using-old-version-of-OpenAL-causes-aliasing-distortion-in-high-frequency-band-05-29):


SlavikMIPT:
Sorry - I didn't check newest version - this is already fixed

PR #7991, opened 2020-06-03
Fix empty dates JSON serialization

aprosvetova:
Empty dates are serialized as `1970-01-01T03:00:00` now.
I think it's not that easy to parse such dates.

I'm not quite familiar with Qt, C++ and TDesktop codebase, so I didn't even test this, but it should obviously work.


john-preston:
I think they should not appear now, I filter out all empty `edited` dates in 2.1.10 version.

Issue #7808, opened 2020-05-05
[Feature Request] UX improvement on the buttons in typing field.


aemxn:
### Is your feature request related to a problem?
Not a problem per se, but rather a UX improvements when typing a message.

To put it simply, when using a mouse to navigate on the screen, we sometimes accidentally hovered on the emoji button which makes it pop a dialog message containing emojis/stickers and whatnot. When our initial intention was just to press the send message button, this popup is kinda distracting and sometimes when miss-clicked, will send wrong message (stickers, gif, emoji).

![ux1][ux1]

### Describe the solution you'd like

Replace the whole button to show an **arrow** instead, which when clicked, it will bring out the buttons. This will not only reduces space, but will improve greatly on users focus on writing message.

![ux2][ux2]

[ux1]:https://i.imgur.com/BVhUG9K.png
[ux2]: https://i.imgur.com/6i0lgu0.jpg

ilya-fedin:
#1333? :thinking:

martagolimowska:
Cghdgjhchj

ilya-fedin <notifications@github.com> schrieb am Mi., 6. Mai 2020, 05:08:

Archydos:
Why hasn't this issue been dealt with? It's been ongoing since 2015 if I'm not mistaken.

aemxn:
@Archydos maybe it's not as important or there's a conflicting rules on Telegram Desktop UX.

It's not an issue if you are using a maximized window, but for some people (like me) always like to use the smallest window to use Telegram Desktop.

**Maximized** -- fixed:

![Imgur](https://i.imgur.com/NJAObfj.png)

**Smallest window** -- popped up:

![Imgur](https://i.imgur.com/PBfTXOc.png)

reallyuniquename:
Still valid.

aemxn:
Updated first post with a gif.

bn-l:
What I love about this issue (still not fixed since 2015--the issue was simply closed) is that when you're typing and your mouse moves accidentally over the fucking emoji button, if you move it away and then back a bit too soon the window will reappear.

This is because the element goes stays transparent for a while (still taking hover events) before actually disappearing. If you're distracted you can do this a few times without realizing what's going on fully.

6 years. Can we at least have an option to remove this button?

reallyuniquename:
Related PR #8009.

bn-l:
@john-preston HELLO!?!?!?!

john-preston:
[...]

Issue #1333, opened 2015-11-26
Hover-activated buttons yes/no?


ralesk:
There's this thing with the attachment and emoticon buttons that they activate on hover (bringing up the little windows)... and to be quite honest I find that a really weird behaviour.

So I'd just like to ask the others here in the issue-land whether it's just me or they too are bothered by it.

It doesn't seem like a hard thing to change either way (I mean, to activate them on click like every other button/menuitem/whatever).

VitRom:
+1, it's VERY annoying.
Especially taking in account a wide-spread reading habit to point with a mouse onto a key areas. One tiny move over the last message timestamp and a huge popup blows in eyes.

amiremami:
I agree with @ralesk and @VitRom . We worked for years with the point and click system, it's really annoying that hovers make the tooltip appears accidently. I was thinking maybe it's better to show the emoticons tooltip as a fixed sidebar, same as this:

![dfgfdg](https://cloud.githubusercontent.com/assets/15929497/11879979/cc1ff232-a4fe-11e5-965e-056c15f4dd5e.png)

This will prevent accident hover thing, and also increases speed in searching and choosing the sticker we want to send.

ralesk:
Still wishing for the hover popups to be turned into click-activated ones.  Any feedback from developers on this, and other users?

ralesk:
I still regularly bump into not wanting to open the emoji popup but managing to and ending up almost clicking an emoji.

amiremami:
After working a lot with Telegram, I get used to it and now I'm too lazy for clicking on emoji button for opening it! 😄

VitRom:
A two kind of jokes are, one with a smile(s), and another one for those who understands that a joke's salt not in the smile(s).
Hover-activation is for the first only.
Er... :-)

ralesk:
My biggest problem with all the UI bugs I reported is that nobody from Telegram bothers to respond to it.

stek29:
#1751 btw

ralesk:
Thanks @stek29. Amazing response time there too.

gottesman:
The user's list from a group, with less than 100 members, just pops up when I move the mouse to the Back Button when I want to go back to main screen. That's pretty annoying, because I end up viewing a user's profile if I just insta-click.

![image](https://cloud.githubusercontent.com/assets/8980291/24885070/c1ce0fa8-1e4c-11e7-9489-1388c6fec3c2.png)

[...]

PR #8009, opened 2020-06-06
Do not pop up emoji tabbed panel and media menu on mouse over

mymedia2:
I suggest to not spawn popups on simple mouse over. When a user just moves a mouse, they does not expect to raise any windows. In Telegram, this may lead to really annoying missclicking when the user wishes to send a text or choose a message for reply.

I was going to leave activation a third panel by double left click on emoji button. Although, I did not find an `addDoubleClickHandler` or similar in `Ui::AbstractButton`. So for now I use single right click to toggle the third panel in history widget. You may consider this as a start approach.

### DIFF
--- a/Telegram/SourceFiles/chat_helpers/tabbed_panel.cpp
+++ b/Telegram/SourceFiles/chat_helpers/tabbed_panel.cpp
@@ -231,10 +231,6 @@ void TabbedPanel::leaveEventHook(QEvent *e) {
 	return TWidget::leaveEventHook(e);
 }
 
-void TabbedPanel::otherEnter() {
-	showAnimated();
-}
-
 void TabbedPanel::otherLeave() {
 	if (preventAutoHide()) {
 		return;
@@ -402,9 +398,7 @@ void TabbedPanel::showStarted() {
 }
 
 bool TabbedPanel::eventFilter(QObject *obj, QEvent *e) {
-	if (e->type() == QEvent::Enter) {
-		otherEnter();
-	} else if (e->type() == QEvent::Leave) {
+	if (e->type() == QEvent::Leave) {
 		otherLeave();
 	}
 	return false;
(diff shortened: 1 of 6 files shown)


magnus-gross:
I really don't want to dishonor the work you did for this PR, but I actually really like the existing mouseover feature and would dislike if it was changed. I can't remember ever misclicking due to this feature.

aemxn:
Just wanted to link a specific issue discussion regarding this:

- https://github.com/telegramdesktop/tdesktop/issues/7808
- https://github.com/telegramdesktop/tdesktop/issues/1333

john-preston:
Sorry, it works as it is supposed to.

reallyuniquename:
@john-preston does it mean there are no plans to implement at least some improvement like popup delay (even ~200ms would be enough)? 

Right now it's like you have to play a cursor maze game if you want to touch anything in the right bottom corner of UI and not trigger emoji panel.

magnus-gross:
@reallyuniquename I really cannot understand your cursor maze game... Are you perhaps a touchpad user with a very bad touchpad? I never had any problems with the popup myself, so I am curious why it seems to be such a problem for some people.

reallyuniquename:
@magnus-gross it really depends on your workflow. I mainly have two issues. If you run Telegram window maximized system toolbars happen to be right under emoji button, it's almost pixel perfect, I kid you not. I trigger emoji popup like 90% of the time when I touch anything in that corner. Likewise when you finish formatting Telegram message using mouse the shortest cursor path to the `Send` button goes right through emoji button.

aemxn:
Maximized Telegram window have the ability to put the emoji panel at the right side of the window -- click the emoji button on maximized -- this will solve the problem, though I doubt people are using maximized window all the time. 

I think the issue is with people who uses the smallest Telegram window size, the popup will happen if you hover around the bottom right corner.

Kirrrr:
Double it.
The option to show emoji popup only after button click wolud be brilliant.

PR #8031, opened 2020-06-09
Update libwayland

ilya-fedin:


imjasonmiller:
Thank you @ilya-fedin and @john-preston! It works again on my end.

PR #8038, opened 2020-06-10
Use gtk clipboard when available to avoid pasting problems

### LINKED ISSUE #5730: tdesktop freezes when pasting an image from the clipboard
### Steps to reproduce
1. Copy an image to the clipboard with `xclip`
2. Try to paste the image in a chat with Ctrl+V

### Expected behaviour
A popup should allow me to send the image compressed or not.

### Actual behaviour
tdesktop freezes with no messages on the terminal, as seen bellow

replay

### Configuration
**Operating system:** Arch Linux

**Version of Telegram Desktop:** 1.5.15

**Used theme**: Custom

There isn't any problem if I select the image through the file picker.

Here's the image I tried to paste, in case it's of any use:

print image

### LINKED ISSUE #6952: Telegram crashes other applications on paste of image data (Linux)
I noticed this problem persisting on many updates for Telegram, but since the latest version (or update to Linux Mint 18, which I had to do to get the latest version to run) it seems to happen way more frequently.
I was unable to find a bug report on this issue, even tho it seems to be an older one.

Basically what happens is, you copy an image to clipboard, paste it to telegram and telegram keeps running as if nothing happened while the other program crashes.
I had it happen mostly with firefox before, tho very rarely, once every two weeks or so. Since TG kept running, I supposed it was a bug in firefox, even tho pasting anywhere else worked as expected.
Since system/TG update, it happened with gimp too, and very frequently there - happened on 4 consecutive tries. Two different programs having the same issues interacting with Telegram (but no other programs) seems to point at an issue with Telegram.

### Steps to reproduce
1. Open Gimp (2.8 used here)
2. Open an image and select a random portion of it, copy to clipboard
_EDIT: it seems the problem occurs way more often, though not exclusively if memory serves well, if you actually select all (ctrl-a) instead of  random portion_
_EDIT 2: Just confirmed it also happens with mere selections_
3. Change to Telegram and hit paste.
4. if problem did not occur, repeat 2, 3 with different parts of the image
Note, it often works, if TG ask you to send as image or file, the bug did not occur

### Alternative steps to reproduce
I'm not sure why, but using this method with this exact image I could reproduce it this way on nearly every try, while it needed quite a few tries of the other method.
The image is a selected part of a screenshot from youtube, that I wanted to send to a friend. It is not specifically engineered to cause this issue.
_EDIT: through more testing, it seems the cause for being able to reproduce it way more with this image was not the image itself, but that this method used select-all. I'll leave it in for now as I can't be sure if that was the only reason this method of reproduction was so successful._
1. Open Gimp (2.8 used here)
2. Open image https://i.imgur.com/1ADl4cs.png in your browser, copy paste it to gimp
2. In gimp, hit select all, and copy
3. Change to Telegram and hit paste.
4. if problem did not occur simply hit abort in the telegram dialog and hit paste again, no need to copy anew.

I found no relieable way to reproduce it in firefox (or another browser). Copying this same image in google chrome and pasting in Telegram just does nothing (no crash either), so that may be a different bug (pasting it in, say, gimp worked as expected). Copying this image in firefox and pasting it in telegram worked as expected.
But as mentioned, I had the bug with firefox only very rarely.

### Expected behaviour
TG asking you to send as image or file, like it does in case the bug doesn't occur.

### Actual behaviour
The source of the image crashes.

### Configuration
Mint 18, freshly updated from Mint 17.3.
Gimp 2.8.16

Telegram Desktop v. 1.9.3

**Used theme**:
Classic

As for logs, Gimp reports this to the console at crash time:

The program 'gimp' received an X Window System error.
This probably reflects a bug in the program.
The error was 'BadWindow (invalid Window parameter)'.
  (Details: serial 194616 error_code 3 request_code 18 minor_code 0)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the --sync command line
   option to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)

(script-fu:10291): LibGimpBase-WARNING **: script-fu: gimp_wire_read(): error

The serial number keeps changing, the error/request/minor code does not

[screen recording of the bug in action.zip](https://github.com/telegramdesktop/tdesktop/files/4017795/screen.recording.of.the.bug.in.action.zip)

### LINKED ISSUE #7223: Cannot paste image on Telegram Desktop on Linux
### Steps to reproduce
1. Right click on an image in Chrome and select Copy image
2. Control + V into Telegram

### Expected behaviour
Send image dialog shoud appear and the image can be sent.

### Actual behaviour
Usually nothing, sometimes after pressing CTRL + V dozens of times it will paste the image. Sometimes it pastes a link even though I selected Copy image and not Copy image address (checked multiple times).

Note: This was described here: https://github.com/telegramdesktop/tdesktop/issues/1892 and here: https://www.reddit.com/r/ManjaroLinux/comments/bm2soh/telegramdesktop_does_no_paste_images_from/

### Configuration
**Operating system:** Linux Mint 19.3 x64 (Cinnamon edition)

**Version of Telegram Desktop:** 1.9.13

**Used theme**: N/A

[log.txt](https://github.com/telegramdesktop/tdesktop/files/4210300/log.txt)
</details>

### LINKED ISSUE #7292: Sending images from clipboard on Linux messes up colors
### Steps to reproduce
1. Use recent `imagemagick` on Linux (tested with v7.0.9-26).
2. Take a screenshot with `import -window root png:- | xclip -selection clipboard -t image/png`
3. Save the screenshot from clipboard to local file for later comparison: `xclip -selection clipboard -t image/png -o > original.png`
4. Paste the screenshot from clipboard to Telegram client, and use "send as file" ("send as image" has the same issue).
5. Download the sent image (file) from Telegram client.
6. Compare previously produces original file with downloaded file.

### Expected behaviour
Files are the same.

### Actual behaviour
Files are different, and the images actually look different on some clients (tested on Telegram Android, Telegram Web).

### Configuration
**Operating system:** Arch Linux

**Version of Telegram Desktop:** 1.9.14 / 1.9.19

**Used theme**: dark

</details>

Screenshot illustrating how the problem looks on Android:

image

Above is the screenshot sent directly from clipboard, below is the same screenshot saved to file and sent from that file.

Actual files involved in the example:
image_2020-02-26_21-21-25
original

ilya-fedin:
Fixes #5730, fixes #6952, fixes #7223, fixes #7292


foxpy:
I am all hands for this PR, have been using patched version for several hours and it works great.

PR #8106, opened 2020-06-20
Disable restart on session start explicitly

ilya-fedin:
Looks like `QApplication::setFallbackSessionManagementEnabled(false)` is not enough

### DIFF
--- a/Telegram/SourceFiles/core/sandbox.cpp
+++ b/Telegram/SourceFiles/core/sandbox.cpp
@@ -27,6 +27,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 #include "facades.h"
 #include "app.h"
 
+#include <QtGui/QSessionManager>
 #include <QtGui/QScreen>
 
 namespace Core {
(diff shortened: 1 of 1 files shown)


Telegram/SourceFiles/core/sandbox.cpp
```diff
@@ -132,6 +133,12 @@ int Sandbox::start() {
 		});
 	});
 
+	// https://github.com/telegramdesktop/tdesktop/issues/948
+	// and https://github.com/telegramdesktop/tdesktop/issues/5022
+	connect(this, &QGuiApplication::saveStateRequest, this, [](QSessionManager &manager) {
+		manager.setRestartHint(QSessionManager::RestartNever);
+	}, Qt::DirectConnection);
```
23rd, line 140:
```suggestion
	const auto restartHint = [](QSessionManager &manager) {
		manager.setRestartHint(QSessionManager::RestartNever);
	};
	connect(
		this,
		&QGuiApplication::saveStateRequest,
		this,
		restartHint,
		Qt::DirectConnection);
```

PR #8182, opened 2020-06-30
Remove redundant move in storage_media_prepare.cpp

mymedia2:


john-preston:
@mymedia2 #8178 was earlier, can you leave only the redundant move part?

mymedia2:
@john-preston I've converted the PR to simple fix of -Wredunant-move warning.

john-preston:
Thanks!

PR #8206, opened 2020-07-03
Update building-msvc.md

Neurotoxin001:
Add some more requirements because without them users will get errors

### DIFF
--- a/docs/building-msvc.md
+++ b/docs/building-msvc.md
@@ -49,9 +49,6 @@ Add **GYP** and **Ninja** to your PATH:
 * Press **Edit**
 * Add ***BuildPath*\\ThirdParty\\gyp** value
 * Add ***BuildPath*\\ThirdParty\\Ninja** value
-* Add ***BuildPath*\\ThirdParty\\Python27** value
-* Add ***BuildPath*\\ThirdParty\\yasm** value
-* Add ***BuildPath*\\ThirdParty\\jom** value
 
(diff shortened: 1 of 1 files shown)


john-preston:
If this line is done when you build:

https://github.com/telegramdesktop/tdesktop/blame/dev/docs/building-msvc.md#L56

There is no need to add Python27, jom and yasm to PATH.

Neurotoxin001:
> There is no need to add Python27, jom and yasm to PATH.

I got errors like this if I don't add them to PATH (on Windows 10 LTSC 1809 at least)
image

And
> "python" not recognized as a command

john-preston:
@Neurotoxin001 Well, then you didn't run the command I mentioned above in the root of the build path - it adds python and yasm and jom to PATH.

Sventra:
I had no issues with the build by following all the steps in the documentation

Neurotoxin001:
> UPD2: But there is another question then — why gyp and Ninja not in this first line with all other PATH's?

I got a problem because I don't have a Git and 

> git clone --recursive https://github.com/telegramdesktop/tdesktop.git

is going right after first line with "SET PATH". After installing Git I start from the second line. That's why I have errors in next steps.

john-preston:
@Neurotoxin001 Ok, so you can just leave the "install git" point and remove the path points, because if you perform all the commands one after another you won't have any errors.

john-preston:
Ok, thanks!

PR #8244, opened 2020-07-13
Move QTextItemInt symbols to lib_ui

ilya-fedin:
Depends on https://github.com/desktop-app/lib_ui/pull/20


john-preston:
I've merged the lib_ui PR, please update the submodules (you can include cmake_helpers update as well).

PR #8255, opened 2020-07-15
Added '-name' command-line argument for custom Qt application name.

### LINKED ISSUE #6951: [Feature Request] Add -class option to set WM_CLASS
I use multiple telegram accounts and in ubuntu gnome, they're grouped in the dock by WM_CLASS of the telegram and because of that grouping accessing them by clicking in the dock is difficult 

I would like to have -class option so I add it to my .desktop file to set the WM_CLASS at startup

XamarinDeveloper:
Will set WM_CLASS option in some desktop environments.
Fixes #6951

### DIFF
--- a/Telegram/SourceFiles/core/launcher.cpp
+++ b/Telegram/SourceFiles/core/launcher.cpp
@@ -22,6 +22,8 @@ namespace {
 
 uint64 InstallationTag = 0;
 
+QString ApplicationName = qsl("TelegramDesktop");
+
 class FilteredCommandLineArguments {
 public:
 	FilteredCommandLineArguments(int argc, char **argv);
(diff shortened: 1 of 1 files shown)


john-preston:
Is the app custom built in those environments? Maybe it should be done in the build time?

XamarinDeveloper:
> Is the app custom built in those environments? Maybe it should be done in the build time?

No, it's for the runtime you can see the details in the related issue

> There's no need in such argument, you can set custom class name via RESOURCE_NAME environment variable.

Wow, thanks, I did not know that and this really should've been said in the issue thread, anyway there still is the need of -class option to set the RESOURCE_NAME environment variable at runtime. I've written a shell file for it but it's not very convenient.

If I do that, will it be approved? or there is no way this option is gonna be implemented?

ilya-fedin:
> anyway there still is the need of -class option to set the RESOURCE_NAME environment variable at runtime.

Why just not toadd the variable to a .desktop file?

XamarinDeveloper:
> Why just not toadd the variable to a .desktop file?

I still need to learn a lot :man_facepalming:
I'll close the issue as well, Thank you so much

PR #8263, opened 2020-07-17
Update telegramdesktop.desktop (add Russian translation of the comment)

OlesyaGerasimenko:
Added Russian translation of the comment

### DIFF
--- a/lib/xdg/telegramdesktop.desktop
+++ b/lib/xdg/telegramdesktop.desktop
@@ -2,6 +2,7 @@
 Version=1.0
 Name=Telegram Desktop
 Comment=Official desktop version of Telegram messaging app
+Comment[ru]=Официальная версия мессенджера Telegram для настольного ПК
 TryExec=telegram-desktop
 Exec=telegram-desktop -- %u
 Icon=telegram


john-preston:
Is this really done that way? Maybe if most (or better all) of current supported languages were added, at least it/de/nl/pt_BR/es/ko.

jmallach:
It would be a lot better to use itstool or similar to manage this, like many eg. GNOME packages do. Editing the desktop file directly is recipe for disaster.

http://itstool.org/

GNOME uses meson for their build system for most modules, but I'm sure there are examples out there of itstool integration in CMake.

ilya-fedin:
> It would be a lot better to use itstool or similar to manage this, like many eg. GNOME packages do.

I see only that it genrates PO files, can't find any mention about .desktop file on documentation page

ilya-fedin:
> GNOME uses meson for their build system for most modules

What KDE uses for this? KDE uses cmake.

jmallach:
Meh, I somehow assumed tdesktop was using PO files.

KDE seems to use PO files, and glancing over the krita repo, they seem to have some machinery that updates the desktop file with its translations in batch. https://invent.kde.org/graphics/krita/-/commits/master/krita/org.kde.krita.desktop. I have no idea how KDE handles its translations these days, though.

PR #8296, opened 2020-07-23
Remove redudant library order hacks

ilya-fedin:


23rd:
The `cmake` submodule is already updated?

ilya-fedin:
Yes

Issue #6020, opened 2019-05-16
Change data export directory format


lapo-luchini:
### Steps to reproduce
1. export data

### Expected behaviour
Create a directory naming scheme that has a proper ordering on multiple exports.

### Actual behaviour
Using `DataExport_dd_mm_yyyy` the time order of exports gets mixed up.

I'd suggest to switch to ISO8601, i.e. `DataExport_yyyy_mm_dd` so that time order is the same as ASCII order.

adripo:
I also need this date format to keep folders organized. Cannot order them properly in windows.

PR #8308, opened 2020-07-26
Export chats with ISO 8601 dates

magnus-gross:
Instead of ChatExport_06_07_2020 we get ChatExport_2020-07-06, which conforms to ISO 8601 and has the advantage that it cannot be misinterpreted in countries where it is common to start with the month instead of the day.

Also sorting directories by name now also means automatically being chronologically sorted.

See #6020

### DIFF
--- a/Telegram/SourceFiles/export/output/export_output_abstract.cpp
+++ b/Telegram/SourceFiles/export/output/export_output_abstract.cpp
@@ -33,11 +33,9 @@ QString NormalizePath(const Settings &settings) {
 	}
 	const auto date = QDate::currentDate();
 	const auto base = QString(settings.onlySinglePeer()
-		? "ChatExport_%1_%2_%3"
-		: "DataExport_%1_%2_%3"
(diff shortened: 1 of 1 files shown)


afwn90cj93201nixr2e1re:
I guess it's produce more problems now (not pr, but not resolving global offset), coz we can't recognize is that UTC based or not, not really informative, could we change it to UTC based? Coz if someone gonna share backup's other ppls can't recognize on which time zone is based on. 

@john-preston

vimpostor:
@afwn90cj93201nixr2e1re  What are you talking about? This PR does not change anything about being UTC based. In fact, this whole part of the code does not have anything to do with UTC. The directory name only contains the date, not the time. It only changes the order from DD-MM-YY to YY-MM-DD, nothing else.

Did you perhaps mean to create an issue? I can't really follow your problem.

afwn90cj93201nixr2e1re:
but yes, pr titles says different.

PR #8362, opened 2020-08-01
Update fcitx5-qt

ilya-fedin:


ilya-fedin:
Moved into https://github.com/telegramdesktop/tdesktop/pull/8402

PR #8368, opened 2020-08-02
Use GDBus to send notifications to avoid Q_OBJECT

ilya-fedin:


magnus-gross:
What is wrong with `Q_OBJECT`?

ilya-fedin:
It is better to ask @john-preston, he said me that Q_OBJECT and MOC are unpleasant in tdesktop when I implemented QtDBus-based NotificationData and finally I found a way how to drop it from NotificationData

xvitaly:
OMG. You remove QtDbus and use ugly GTK version? This is so bad. Next time start rewriting the whole tdesktop to GTK.

ilya-fedin:
> and start using the ugly GTK implementation

glib is not GTK, glib is used even by Qt itself. It provides base methods for *nix such as getting home directory, working with .desktop files, file associations, etc.

xvitaly:
> glib is not GTK, glib is used even by Qt itself.

OK, the ugly glib plain C implementation instead of well-tested native QtDbus.

ilya-fedin:
> well-tested native QtDbus.

I don't know why you call QtDBus well-tested, it uses the old libdbus library that is almost abandoned.

xvitaly:
> I don't know why you call QtDBus well-tested, it uses the old libdbus library that is almost abandoned.

Qt5Dbus is a part of Qt, whatever it uses under the hood.

ilya-fedin:
> Qt5Dbus is a part of Qt

You may see that tdesktop prefers non-Qt implementations rather than Qt ones for anything else.
E.g. lib_rpl instead og Qt signals, lib_crl instead of QThread, manual work with WinAPI instead of QtWinExtras, manual work with Mac APIs instead of QtMacExtras, manual getting of executable path instead of QCoreApplication::applicationFilePath, system() instead of QProcess

xvitaly:
> You may see that tdesktop prefers non-Qt implementations rather than Qt ones for anything else.

And this is not good at all.

ilya-fedin:
> And this is not good at all.

¯\_(ツ)_/¯

magnus-gross:
I suspect that `Q_OBJECT` is not wanted, because it has quite the heavy overhead, though using glib might have some too.
Also I see that tdesktop is still using the old Qt signal and slot syntax (i.e. `SIGNAL()` and `SLOT()` macros) instead of the new syntax, so perhaps that was a reason back then to abandon these cursed Qt signals.

Also @xvitaly there is no reason to force Qt library functions everywhere, sometimes they are not that well-tested. See for e.g. why QtLottie was abandoned for rlottie.

xvitaly:
> See for e.g. why QtLottie was abandoned for rlottie.

QtLottie was too buggy and unstable due to an early development stage.

ilya-fedin:
Btw, QtDBus has not only MOC cons, but functional too. For example, when I implemented [flatpak's autostart API](https://github.com/telegramdesktop/tdesktop/blob/361b99b0c9c9f69bc8479a4413dbe62113405734/Telegram/SourceFiles/platform/linux/specific_linux.cpp#L78-L109), I wasn't able to connect to a signal that doesn't exist on connect moment. However this is required to work with flatpak APIs.
With gdbus I will be able to just copy the [implementation from libportal](https://github.com/flatpak/libportal/blob/bff3289781ef27388e75f772d5fe9dee20fdfb53/libportal/background.c#L177-L186)

I guess QtDBus has this limitation since it based on the old libdbus library

john-preston:
Thanks!

PR #8428, opened 2020-08-15
Fix #8427: folder exclusion filters should always take priority over inclusion

### LINKED ISSUE #8427: `NoArchive` folder filter is handled inconsistent with mobile iOS app when a chat is both included and excluded
### Steps to reproduce
1. Choose an arbitrary chat X.
2. Archive the chat X.
3. Create a folder which "includes" the chat X (as an individual chat, not by chat type), but "excludes" all archived chats. 
4. Check whether the folder shows the chat X in tdesktop.
4. Check whether the folder shows the chat X in mobile app (I've used iOS).

### Expected behaviour
Behavior should be consistent between tdesktop and iOS app: either chat X is consistently shown both in tdesktop and mobile app, or the folder is consistently empty.

### Actual behaviour
Behavior is inconsistent: tdesktop 2.3 shows the chat X (i.e. its exclusion by "archived" does not work), but iOS app 7.0 does not (i.e. it excludes all archived chats, even if they're explicitly listed in the "archived" section).

### Configuration
**Operating system:** Windows 10 Pro 2004 (build 19041.450)

**Version of Telegram Desktop:** 2.3

**Used theme**: the default one

### Extra thoughts

I believe the correct behavior is the one where "exclude" rules have priority (it also coincides with iOS app's logic). Usecase: I archive/unarchive a chat on a regular basis and I would like it to be hidden/shown in all folders where it participates. E.g. some chats are for non-urgent work, so I archive them from "Work" folder while on vacation.

### Possible fix

Fix should probably land around `ChatFilter::contains` around [`data_chat_filters.cpp:215`](https://github.com/telegramdesktop/tdesktop/blob/887b6629f66e3e15786938f3bf7593bca39eaa31/Telegram/SourceFiles/data/data_chat_filters.cpp#L215): mass exclusion policies (`NoMuted`, `NoRead`, `NoArchived`) are checked for chats included by a mass inclusion policy (`Bots`, `Contacts`, etc), but not for chats included individually (`_always`). Note that I have zero experience in tdesktop code, and there may be other places need fixing, e.g. `ChatFilters::archiveNeeded` and its users, maybe?

yeputons:
NOTE: completely untested, but may fix #8427.

Previously if you had a chat individually included in a folder (via `_always`) it could not be excluded by a mass exclusion filter (e.g. `NoArchived`). Now these filters are always applied.

Maybe a small refactoring is a good idea: check all exclusion filters right after `if (_never.contains(history))`, and then simply `return (_flags & flag) || _always.contains(history);` in the very end. That will make the logic much clearer: apply all exclusions, and then include only some.

### DIFF
--- a/Telegram/SourceFiles/data/data_chat_filters.cpp
+++ b/Telegram/SourceFiles/data/data_chat_filters.cpp
@@ -198,8 +198,7 @@ bool ChatFilter::contains(not_null<History*> history) const {
 	if (_never.contains(history)) {
 		return false;
 	}
-	return false
-		|| ((_flags & flag)
+	return ((_flags & flag) || _always.contains(history))
 			&& (!(_flags & Flag::NoMuted)
 				|| !history->mute()
 				|| (history->hasUnreadMentions()
@@ -211,8 +210,7 @@ bool ChatFilter::contains(not_null<History*> history) const {
 				|| history->hasUnreadMentions()
 				|| history->fakeUnreadWhileOpened())
 			&& (!(_flags & Flag::NoArchived)
-				|| (history->folderKnown() && !history->folder())))
-		|| _always.contains(history);
+				|| (history->folderKnown() && !history->folder()));
 }
 
 ChatFilters::ChatFilters(not_null<Session*> owner) : _owner(owner) {


john-preston:
@yeputons But this is not how it is supposed to work :( maybe it'll change in the future, but now the rules for single chats, selected explicitly, always have more weight than category rules (unread / not archived / channels / contacts).

yeputons:
Oh, that's interesting. Is there a doc to check against? Then the iOS client is the one needing a fix. I've filed a bug there as well.

PR #8530, opened 2020-09-01
Separate dependencies and Telegram build to jobs

RadRussianRus:
This PR addresses the problem with rebuilding on CI.

## About the problem
When CI action fails for whatever reason, next time it will be fully rebuilt, both dependencies and Telegram. This leads to increased build times due to unnecessary rebuilds of dependencies.

## Proposed solution
My currently proposed solution is to split dependencies and Telegram build to separate jobs. That way even if Telegram build will fail, dependencies will be cached and reused on next run.

#### Why not use `continue-on-error`?
`continue-on-error` will make job marked successfully done even when it fails, so finding errors will be harder.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -42,10 +42,23 @@ on:
       - 'Telegram/configure.bat'
       - 'Telegram/Telegram.plist'
 
+env:
+  GIT: "https://github.com"
+  QT: "5_12_8"
+  QT_PREFIX: "/usr/local/desktop-app/Qt-5.12.8"
+  OPENSSL_VER: "1_1_1"
+  OPENSSL_PREFIX: "/usr/local/desktop-app/openssl-1.1.1"
+  CMAKE_VER: "3.17.0"
+  UPLOAD_ARTIFACT: "false"
+  ONLY_CACHE: "false"
+  MANUAL_CACHING: "6"
+  DOC_PATH: "docs/building-cmake.md"
+  AUTO_CACHING: "1"
+
 jobs:
 
-  linux:
-    name: Ubuntu 14.04
+  dependencies:
+    name: "Ubuntu 14.04: dependencies"
     runs-on: ubuntu-latest
     container: ubuntu:trusty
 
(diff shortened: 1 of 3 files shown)


.github/workflows/linux.yml
```diff
@@ -411,6 +372,21 @@ jobs:
           cd ..
           rm -rf libxkbcommon
 
+      - name: Libwayland.
```
ilya-fedin, line 375: Why did you move this?

.github/workflows/linux.yml
```diff
@@ -42,32 +42,30 @@ on:
...
-    name: Ubuntu 14.04
+  dependencies:
+    name: "Ubuntu 14.04: dependencies"
     runs-on: ubuntu-latest
     container: ubuntu:trusty
 
     strategy:
       matrix:
         defines:
           - ""
```
ilya-fedin, line 68: You can remove it since it is not used

.github/workflows/linux.yml
```diff
@@ -111,10 +109,10 @@ jobs:
           sudo update-alternatives --config gcc && \
           sudo add-apt-repository --remove ppa:ubuntu-toolchain-r/test -y
 
+      # We don't need recursive clone to build dependencies.
       - name: Clone.
         uses: actions/checkout@v2
         with:
-          submodules: recursive
           path: ${{ env.REPO_NAME }}
```
ilya-fedin, line 116: And this too

RadRussianRus: No, it's needed. Since we're generating cache key from current workflow file, we need at least shallow clone of repo.

.github/workflows/linux.yml
RadRussianRus: I thought of it as a placeholder, like on other workflows, but since there are not "Read defines" for them, you're probably right. Removed in 92b261b.

.github/workflows/linux.yml
RadRussianRus: Oops, missed that part. Fixed in 8d35320.

.github/workflows/win.yml
```diff
@@ -399,6 +374,154 @@ jobs:
...
+          echo Find any version of Python 2.
+          for /D %%a in (C:\hostedtoolcache\windows\Python\2.*) do (
+            SET PY2=%%a\x64
+          )
+          if [%PY2%] == [] (
+            echo Python 2 is not found.
+            exit 1
+          )
+          echo Found %PY2%.
+          echo ::set-env name=PY2::%PY2%
```
23rd, line 453:
```suggestion
```

23rd:
I don't like to see so much repeating code.

I have opened a new pull request to your branch, but it's not enough.
For example, each job installs the same packages now. This should be reduced. Each job should only install a necessary set of packages.

RadRussianRus:
> I don't like to see so much repeating code.
> 
> I have opened a new pull request to your branch, but it's not enough.
> For example, each job installs the same packages now. This should be reduced. Each job should only install a necessary set of packages.

Agree, that should be reduced as much as possible. I'll test later on how this could be reduced.

VChernomYA (approved):

VChernomYA (approved):

.github/workflows/linux.yml
VChernomYA: https://github.com/telegramdesktop/tdesktop/commit/8d353202a4970e5c12e0050531662904b37f1bb3

john-preston:
@RadRussianRus Is this still relevant? Or could it be closed?

RadRussianRus:
I guess it should be closed since CI have changed a lot. I'll open a new repo if needed.

PR #8547, opened 2020-09-04
Cache make install calls in macos action

ilya-fedin:


23rd:
Why do we have to cache a folder with the opus source code?
The opus repository is small and can be cloned pretty fast. You can use `git clone -b v1.3 --depth=1 $GIT/xiph/opus` as it was used in `linux.yml`.

ilya-fedin:
fixed

Issue #8626, opened 2020-09-21
tg:// link in firefox focuses telegram but does nothing


txtsd:
### Steps to reproduce
1. Click on a tg:// link in firefox to join a group

### Expected behaviour
Telegram should focus and join the group

### Actual behaviour
Telegram focuses, but the group is neither opened nor joined

### Configuration
**Operating system:** Arch Linux

**Version of Telegram Desktop:** 2.3.2

**Installation source (Linux Only)** - the official website / GitHub releases / flatpak / snap / distribution package: https://www.archlinux.org/packages/community/x86_64/telegram-desktop/

**Used theme**: Night Mode

<details><summary><b>No Logs Yet</b>:</summary>
Insert log.txt here (if necessary)
</details>

ilya-fedin:
Are you sure that link that you trying to open is supported in tdesktop?

ilya-fedin:
They disabled the functionality to install URL handler, so it is expected if it doesn't work
https://github.com/archlinux/svntogit-community/blob/packages/telegram-desktop/trunk/PKGBUILD#L43

txtsd:
> Are you sure that link that you trying to open is supported in tdesktop? Mobile clients have links that aren't supported by tdesktop afaik.

Yes, it opens in Bettergram correctly.

> They disabled the functionality to install URL handler, so it is expected if it doesn't work
> https://github.com/archlinux/svntogit-community/blob/packages/telegram-desktop/trunk/PKGBUILD#L43

So if I toggle that flag it'll work as expected?

ilya-fedin:
> So if I toggle that flag it'll work as expected?

Yes, it should

ilya-fedin:
But it updates the handler only when detects app update, so you will should trigger handler registration by entering `registertg` like a cheat code in settings (it won't do anything in the current build due to the flag)

txtsd:
What does `registertg` actually do? Does it add an entry in mimeapps.list so xdg-open can correctly open `x-scheme-handler/tg` links?

ilya-fedin:
> What does `registertg` actually do?

calls glib and it does some magic

txtsd:
Elaborate please

ilya-fedin:
> Elaborate please

I don't know what glib does under the hood exactly, but it works and other applications are using it too

ilya-fedin:
Maybe this define should be removed :thinking:

txtsd:
That's a good idea.

[...]

Issue #7305, opened 2020-02-28
tg:// not working on many devises on desktop


mishell-trickster:
I try to open chat via tg://resolve?domain=mishell_zero

It should open chat, but open only telegram app(not target chat). 

Telegram version: 1.9.14
OS: Linux, Windows, MacOS
Theme: Default

I tried this on 6 devices. 1 linux, 1 mac, 4 windows. Only on 2 windows laptops its work. Another devices just open app. 

This is bug? What should I do?

ilya-fedin:
Try to open settings and write (as a cheat code, without a text field) `registertg`

mishell-trickster:
> Try to open settings and write (as a cheat code, without a text field) `registertg`

Dont work, I see same bug. And this is not fix formally, because another people dont know about this method.

ilya-fedin:
I don't know about windows/macos, but maybe I can help with linux.
Do you using official binary or snap/flatpak/distribution package?
If the first, can you attach ~/.local/share/applications/telegramdesktop.desktop?

Aokromes:
@mishell-trickster

mishell-trickster:
> I don't know about windows/macos, but maybe I can help with linux.
> Do you using official binary or snap/flatpak/distribution package?
> If the first, can you attach ~/.local/share/applications/telegramdesktop.desktop?

I use official binary.
This is my desktop file:

_[Desktop Entry]
Version=1.0
Name=Telegram Desktop
Comment=Official desktop version of Telegram messaging app
TryExec=/home/mishell/Desktop/Telegram
Exec=/home/mishell/Desktop/Telegram -- %u
Icon=telegram
Terminal=false
StartupWMClass=TelegramDesktop
Type=Application
Categories=Chat;Network;InstantMessaging;Qt;
MimeType=x-scheme-handler/tg;
Keywords=tg;chat;im;messaging;messenger;sms;tdesktop;
X-GNOME-UsesNotifications=true_

ilya-fedin:
`xdg-mime query default x-scheme-handler/tg`

Aokromes:
@mishell-trickster

Aokromes:
no answer for over a month, closing.if this bug is still valid open a new ticket using updated version.

Issue #3674, opened 2017-07-18
Tg links don't open on Telegram (linux)


svhnl:
When I open a tg link, firefox says that the address wasn't understood and that it wasn't linked to an application. When I took a look in the applications tab, there was no tg link at all. I already re-installed telegram

I use telgram on linux mint 18.1 64 bit  with Telegram 1.1.10 
<details><summary><b>Logs</b>:</summary>
Insert logs here (if necessary)

</details>

stek29:
Are you using TDesktop from official site? Also, how are you launching it? Do you have `tg://` in your .desktop file?
And what's your DE? What does xdg-utils show for tg?

Bluscream:
I have the same problem i just get "The address wasn’t understood".

![image](https://user-images.githubusercontent.com/3318223/29995628-907e3d3a-8fdd-11e7-95cd-7c17f415f9cd.png)

svhnl:
I use the official telegram desktop from the official website I launch it like it came installed. My DE is KDE plasma. Your last question I don't get.

cono:
@svhnl it should be covered by:
```
% grep MimeType /usr/share/applications/telegram-desktop.desktop
MimeType=x-scheme-handler/tg;
```

If you have this line in the Desktop file, then it means that this is not problem of the Telegram, its more about your DE and XDG settings.

Refer to this documentation: https://wiki.archlinux.org/index.php/default_applications

pasharik:
Hi,
I have similar issue. Kubuntu 16.04
```
xdg-mime query default x-scheme-handler/tg
telegram.desktop
```

I've created telegram.desktop manually:
```
[Desktop Entry]
Type=Application
Name=Telegram
Exec=/opt/telegram/Telegram -- %u
Icon=telegram
Terminal=false
MimeType=x-scheme-handler/tg;
X-KDE-Protocols=tg
```

d0tfile:
It is not Telegram problem. In Firefox go to about:config and create new bool "network.protocol-handler.expose.tg" with value "False".
http://kb.mozillazine.org/Register_protocol#Linux

Issue can be closed.

dagelf:
This is a Telegram problem - the shortlinks should be joinable with the web version, but they are not. Any idea how to convert a tg.me link to a url that can be used by web.telegram.org?

atommixz:
works for me in firefox:
about:config
create key:value (boolean)
network.protocol-handler.expose.tg false
restart browser

[...]

Issue #2630, opened 2016-11-12
Telegram for linux is not able to open tg links


sprnza:
Hi there! My problem is the Telegram desktop client for linux is not opening tg links from Firefox. The Telegram client is being opened but there is no an appropriate chat/bot opened. My telegramdesktop.desktop file has this line

`Exec=telegram-desktop -- %u`

if I run

`telegram-desktop -- tg:link`

the link is being opened fine. Is it just my installation problem or what?

stek29:
What do you see when you execute `xdg-mime query default x-scheme-handler/tg` ?

sprnza:
`speranza@~$ xdg-mime query default x-scheme-handler/tg
telegramdesktop.desktop
`

sprnza:
The app is being launched but it doesn't open up the link.

john-preston:
@sprnza Can you launch it by a link and then look at ~/.TelegramDesktop/log.txt (or ~/.local/TelegramDesktop/log.txt), the arguments should be listed there.

sprnza:
```
[2016.11.12 16:00:17] Launched version: 10011, alpha: [FALSE], beta: 0, debug mode: [FALSE], test dc: [FALSE]
[2016.11.12 16:00:17] Executable dir: /usr/bin/, name: telegram-desktop
[2016.11.12 16:00:17] Initial working dir: /home/speranza/
[2016.11.12 16:00:17] Working dir: /home/speranza/.TelegramDesktop/
[2016.11.12 16:00:17] Arguments: "/usr/bin/telegram-desktop" "tg://resolve?domain=mpscmaterial"
```

john-preston:
@sprnza Where did the "--" argument go? :/

sprnza:
@john-preston Good question.

sprnza:
```
speranza@~$ locate telegramdesktop.desktop
/usr/share/applications/telegramdesktop.desktop
speranza@~$ cat /usr/share/applications/telegramdesktop.desktop |grep Exec
Exec=telegram-desktop -- %u
```

AndydeCleyre:
@sprnza double check that you don't have a broken telegram launcher hiding in `~/.local/share/applications`.

sprnza:
@AndydeCleyre, yep "locate telegramdesktop.desktop" in the previous post searches through the whole system

AndydeCleyre:
[...]

PR #8627, opened 2020-09-21
Get rid of redudant cmake options

ilya-fedin:


john-preston:
Thanks!

AndydeCleyre:
I usually build with `TDESKTOP_DISABLE_REGISTER_CUSTOM_SCHEME` -- can someone tell me why it's become redundant?

ilya-fedin:
@AndydeCleyre there's no need to disable scheme registration anymore

AndydeCleyre:
@ilya-fedin Thank you but can you elaborate a bit? You mean it will never try to do scheme registration?

ilya-fedin:
> Thank you but can you elaborate a bit? You mean it will never try to do scheme registration?

it will, but it's ok. It won't introduce a second menu item.

AndydeCleyre:
So if a user has a different app associated with `tg://` URIs, then installs this, this won't steal that association?

ilya-fedin:
> So if a user has a different app associated with `tg://` URIs, then installs this, this won't steal that association?

It will.

AndydeCleyre:
So how can we continue to prevent this app from doing that? On a linux system, AFAIU, all that should be needed is the 

```ini
MimeType=x-scheme-handler/tg;
```

line in the `.desktop` file, and anything further will be invasive and stepping on users' and admins' toes. In other words, the app itself should *advertise* its ability to deal with those URIs, not *choose for a user or system* that this is the best choice.

ilya-fedin:
>  the app itself should _advertise_ its ability to deal with those URIs, not _choose for a user or system_ that this is the best choice

This behavior was chosen by @john-preston, if you don't like it, you should discuss with him about that. Personally, I think there should be a dialog like firefox has, but I guess preston won't be thrilled with this idea :)

AndydeCleyre:
I thought the choice until this PR was to offer `TDESKTOP_DISABLE_REGISTER_CUSTOM_SCHEME` as a way to disable the invasive behavior. Is there an issue or other reference for the reversal of that choice?

ilya-fedin:
> I thought the choice until this PR was to offer `TDESKTOP_DISABLE_REGISTER_CUSTOM_SCHEME` as a way to disable the invasive behavior.

As far as I know, it was there only due to the fact that it creates a second menu entry for packaged installation. Since it compatible with packaged build now, there no need for it anymore.

>  Is there an issue or other reference for the reversal of that choice?

No, this was discussed privately. The main reason was the fact that people installing tdesktop via package managers creating issues about the non-working handler. So, we decided to remove the flag and if some maintainer really want to remove handler installation, he should patch the source code and only he will be completely responsible for this choice.

AndydeCleyre:
> The main reason was the fact that people installing tdesktop via package managers creating issues about the non-working handler.

Can you reference such an issue? I found:

- #2630 -- but that was a multiple-handler-entry issue
- #3674 -- but that's **not** using a package manager and is about firefox, **not** xdg
- #7305 -- but that's **not** using a package manager and the reporter went MIA
- #8626 -- never properly debugged, possibly a firefox issue, not enough info

So that last one, #8626, is the only one I could find matching your description. Are you saying "the main reason" for this change was that one bug report that was never thoroughly investigated? Or am I missing reports in my search?

ilya-fedin:
> Are you saying "the main reason" for this change was that one bug report that was never thoroughly investigated?

Yeah, such issues appear from time to time and reporters just disappear usually, so we decided just to ensure that telegram is default, just like on other platforms.

Just like I said earlier, if you don't like the behavior, you should discuss that with preston, not with me.

AndydeCleyre:
I cannot find a second issue as you describe. Are you *sure*

> such issues appear from time to time

?

As I said, I only found one such issue, and it was never thoroughly investigated.

> if you don't like the behavior, you should discuss that with preston, not with me.

I don't mean to target you, I just thought this was the most relevant place for the inquiry. 

Pinging @john-preston

ilya-fedin:
> Are you _sure_

No, I'm not sure.

ilya-fedin:
I understand that you just want the old behavior, but this issue is cross-platform, you could have multiple Telegram clients on Windows or macOS and tdesktop will overwrite your default. There is no reason that Linux should be handled differently.

AndydeCleyre:
I cannot speak for the best ways to handle protocol assignment on those other platforms.

It seems to me that this new behavior removes the option for protocol assignment to be handled *correctly* on linux, not simply *differently*.

I have yet to see any evidence or explanation of the `TDESKTOP_DISABLE_REGISTER_CUSTOM_SCHEME` behavior being incorrect or inferior, while the opposite and newly mandatory behavior is demonstrably problematic in that it clobbers user preferences, requiring manual intervention on every update, and is in principle disrespectful to the user.

ilya-fedin:
> while the opposite and newly mandatory behavior is demonstrably problematic in that it clobbers user preferences, requiring manual intervention on every update, and is in principle disrespectful to the user

This issue is not specific to Linux and should be addressed in a cross-platform manner.

ilya-fedin:
> I have yet to see any evidence or explanation of the `TDESKTOP_DISABLE_REGISTER_CUSTOM_SCHEME` behavior being incorrect or inferior, while the opposite and newly mandatory behavior is demonstrably problematic in that it clobbers user preferences, requiring manual intervention on every update, and is in principle disrespectful to the user.

Well, I can say that this is just your opinion. There's another opinion: automatic default chaging is useful for who want just to start the app and use it. We can argue about that for long, but I tired of it and want to stop the discussion.

AndydeCleyre:
The fact that it clobbers user preferences is not an opinion.

> automatic default chaging is useful for who want just to start the app and use it

But it's not been demonstrated that this actually solves an issue that is not already solved by the content of the `.desktop` file.

PR #8641, opened 2020-09-26
Update BUG_REPORT.md

SmartManoj:


john-preston:
This way there will be a lot of 2.3.1-s, and I'll have no way to know if the author checked the version number.

SmartManoj:
Then I will go with the user script.

PR #8657, opened 2020-09-29
Fix showing autohidden taskbar with maximized window

### LINKED ISSUE #3579: Does not respect Auto-hidden Taskbar when the window is maximized
### Steps to reproduce
1. enable Auto hide the taskbar in windows 8.1
2. maximize the main window of telegram
3. hover on taskbar to show it ...

### Expected behaviour
It should show taskbar, but the telegram window is not  allowing it.

### Actual behaviour
Taskbar is hidden when trying to show it by hovering the mouse.

### Configuration
**Operating system:** Windows 8.1

**Version of Telegram Desktop:** 1.1.7

**Used theme**: Default

ilya-fedin:
Fixes #3579

### DIFF
--- a/Telegram/SourceFiles/platform/win/windows_event_filter.cpp
+++ b/Telegram/SourceFiles/platform/win/windows_event_filter.cpp
@@ -35,6 +35,35 @@ bool IsCompositionEnabled() {
 	return success && result;
 }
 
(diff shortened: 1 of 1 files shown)


john-preston:
Thanks!

PR #8667, opened 2020-10-01
Dynamically load packed resources from separate file

mymedia2:
You may be interested in this patch. It reduces amount of main memory required for linking stage.

### DIFF
--- a/Telegram/CMakeLists.txt
+++ b/Telegram/CMakeLists.txt
@@ -34,7 +34,21 @@ include(cmake/td_lang.cmake)
 include(cmake/td_scheme.cmake)
 include(cmake/td_ui.cmake)
 
-set_target_properties(Telegram PROPERTIES AUTOMOC ON AUTORCC ON)
+set_target_properties(Telegram PROPERTIES AUTOMOC ON)
+if (TDESKTOP_RESOURCES STREQUAL "Bundled")
+    set_target_properties(Telegram PROPERTIES AUTORCC ON)
+elseif (TDESKTOP_RESOURCES STREQUAL "Packed")
+    set(rcc_flags --binary "$<TARGET_PROPERTY:Telegram,AUTORCC_OPTIONS>")
+    set(qrc_files "$<FILTER:$<TARGET_PROPERTY:Telegram,SOURCES>,INCLUDE,\\.qrc$>")
+    target_compile_definitions(Telegram PRIVATE TDESKTOP_USE_PACKED_RESOURCES)
+    target_sources(Telegram PRIVATE "${CMAKE_BINARY_DIR}/tresources.rcc")
+    add_custom_command(OUTPUT "${CMAKE_BINARY_DIR}/tresources.rcc"
+        DEPENDS ${qrc_files}  # FIXME: incomplete dependencies list, missing qrc content
+        COMMAND Qt5::rcc ${rcc_flags} -o "${CMAKE_BINARY_DIR}/tresources.rcc" ${qrc_files}
+        COMMAND_EXPAND_LISTS VERBATIM
+    )
+    set_target_properties(Telegram PROPERTIES RESOURCE "${CMAKE_BINARY_DIR}/tresources.rcc")
+endif()
 
 target_link_libraries(Telegram
 PRIVATE
@@ -1285,7 +1299,11 @@ endif()
 if (LINUX AND DESKTOP_APP_USE_PACKAGED)
     include(GNUInstallDirs)
     configure_file("../lib/xdg/telegramdesktop.appdata.xml.in" "${CMAKE_CURRENT_BINARY_DIR}/telegramdesktop.appdata.xml" @ONLY)
-    install(TARGETS Telegram RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}" BUNDLE DESTINATION "${CMAKE_INSTALL_BINDIR}")
+    install(TARGETS Telegram
+        RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
+        BUNDLE DESTINATION "${CMAKE_INSTALL_BINDIR}"
+        RESOURCE DESTINATION "${CMAKE_INSTALL_DATADIR}/TelegramDesktop"
+    )
     install(FILES "Resources/art/icon16.png" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/16x16/apps" RENAME "telegram.png")
     install(FILES "Resources/art/icon32.png" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/32x32/apps" RENAME "telegram.png")
     install(FILES "Resources/art/icon48.png" DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/icons/hicolor/48x48/apps" RENAME "telegram.png")
--- a/Telegram/SourceFiles/core/file_utilities.cpp
+++ b/Telegram/SourceFiles/core/file_utilities.cpp
@@ -402,16 +402,16 @@ bool GetDefault(
 void Resources::LoadAllData() {
 #ifdef TDESKTOP_USE_PACKED_RESOURCES
 	// Load resources packed into a separated file
-	QStringList paths;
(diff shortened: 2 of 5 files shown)


Telegram/CMakeLists.txt
```diff
@@ -68,7 +68,20 @@ set(dependent_style_files
...
+set_target_properties(Telegram PROPERTIES AUTOMOC ON)
+if (TDESKTOP_RESOURCES STREQUAL "Bundled")
+    set_target_properties(Telegram PROPERTIES AUTORCC ON)
+elseif (TDESKTOP_RESOURCES STREQUAL "Packed")
+    set(qrc_files "$<FILTER:$<TARGET_PROPERTY:Telegram,SOURCES>,INCLUDE,\\.qrc$>")
+    target_compile_definitions(Telegram PRIVATE TDESKTOP_USE_PACKED_RESOURCES)
+    target_sources(Telegram PRIVATE "${CMAKE_BINARY_DIR}/tresources.rcc")
+    add_custom_command(OUTPUT "${CMAKE_BINARY_DIR}/tresources.rcc"
+        DEPENDS "${qrc_files}"  # FIXME: incomplete dependencies list, missing qrc content
+        COMMAND rcc --binary -o "${CMAKE_BINARY_DIR}/tresources.rcc" "${qrc_files}"
```
xvitaly, line 80: RCC binary is not in $PATH on the most of distributions.

Telegram/CMakeLists.txt
```diff
@@ -1354,7 +1367,11 @@ endif()
 if (LINUX AND DESKTOP_APP_USE_PACKAGED)
     include(GNUInstallDirs)
     configure_file("../lib/xdg/telegramdesktop.appdata.xml.in" "${CMAKE_CURRENT_BINARY_DIR}/telegramdesktop.appdata.xml" @ONLY)
-    install(TARGETS Telegram RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}" BUNDLE DESTINATION "${CMAKE_INSTALL_BINDIR}")
+    install(TARGETS Telegram
+        RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
+        BUNDLE DESTINATION "${CMAKE_INSTALL_BINDIR}"
+        RESOURCE DESTINATION "${CMAKE_INSTALL_DATADIR}/TelegramDesktop"
```
xvitaly, line 1373: Please use lower case for the resources directory name - `telegram-desktop` (must follow package name).

Telegram/cmake/telegram_options.cmake
```diff
@@ -10,6 +10,9 @@ set(TDESKTOP_API_ID "0" CACHE STRING "Provide 'api_id' for the Telegram API acce
 set(TDESKTOP_API_HASH "" CACHE STRING "Provide 'api_hash' for the Telegram API access.")
 set(TDESKTOP_LAUNCHER_BASENAME "" CACHE STRING "Desktop file base name (Linux only).")
 
+set(TDESKTOP_RESOURCES "Bundled" CACHE STRING "Determines how to import resource files.")
```
xvitaly, line 13: String value for setting is not good for me. Please use boolean with ON/OFF.

Telegram/CMakeLists.txt
ilya-fedin: > must follow package name

No, it must not. It should follow project name, which is `tdesktop` (according to the repo name)

xvitaly: tdesktop is suitable too. Only lowercase, please.

Telegram/CMakeLists.txt
mymedia2: Your suggestions? [Here][1] is example how to invoke the tool. And that doc does not mention custom $PATH.

 [1]: https://doc.qt.io/qt-5/rcc.html

Telegram/cmake/telegram_options.cmake
mymedia2: I am not agree with you. A third option is possible here yet not implemented. For "Copied" resources, which would be placed on file system as is. And this third option may been incompatible with "Bundled" or "Packed" resources.

Telegram/CMakeLists.txt
mymedia2: The directory name is not arbitrary here. From what I understand, it should match application name set in Launcher::init(). This is how QStandardPaths works.

https://github.com/telegramdesktop/tdesktop/blob/ac5079bac2f42e3c3e5282198943ceec66fba707/Telegram/SourceFiles/core/launcher.cpp#L302

Telegram/CMakeLists.txt
mymedia2: On the other hand, there is the [AUTORCC_EXECUTABLE][1] property determined by CMake. I am not 100% sure should I interfere automatic CMake magic here, though.

 [1]: https://cmake.org/cmake/help/v3.18/prop_tgt/AUTORCC_EXECUTABLE.html

mymedia2: I was wrong. CMake seems to not define the property by itself. As I've rewritten the line with get_target_property() and a fallback, can you please check if it works for you?

xvitaly: Still not good. On Fedora, cmake will never find rcc, because it is located in `/usr/lib64/qt5/bin/rcc`. Use `${qt_loc}/bin/rcc` instead.

mymedia2: Well, the ${qt_loc} variable is not defined in packaged build. But I just discovered here the Qt5::rcc imported target that points to /usr/lib/qt5/bin/rcc on my system. There is no need to reinvent the wheel :tada:

xvitaly: `Qt5::rcc` is LGTM now. Thanks.

john-preston:
I'm not going to use that in my builds of tdesktop (because almost all of them are single-binary), but I can merge it if it is really needed to (more than one?) package maintainers. If the demand is not strong then I'd leave it as it is, because it means less code to maintain.

Telegram/SourceFiles/core/file_utilities.cpp
```diff
@@ -397,3 +398,20 @@ bool GetDefault(
...
+#ifdef _DEBUG
+	paths += cExeDir();
+#endif
+	paths += QStandardPaths::standardLocations(QStandardPaths::AppDataLocation);
+	for (QString directory : paths) {
+		if (QResource::registerResource(directory + u"/tresources.rcc"_q)) {
+			return;  // found
+		}
+	}
+	qFatal("Packed resources not found");
```
ilya-fedin, line 415:
```suggestion
	Unexpected("Packed resources not found");
```

Telegram/SourceFiles/core/file_utilities.cpp
```diff
@@ -397,3 +398,20 @@ bool GetDefault(
...
+	QStringList paths;
+#ifdef _DEBUG
+	paths += cExeDir();
+#endif
+	paths += QStandardPaths::standardLocations(QStandardPaths::AppDataLocation);
+	for (QString directory : paths) {
+		if (QResource::registerResource(directory + u"/tresources.rcc"_q)) {
+			return;  // found
+		}
+	}
```
ilya-fedin, line 414:
```suggestion
	const auto paths = QStringList{
#ifdef _DEBUG
		cExeDir(),
#endif
	} + QStandardPaths::standardLocations(QStandardPaths::AppDataLocation);
	for (const auto directory : paths) {
		if (QResource::registerResource(directory + u"/tresources.rcc"_q)) {
			return;  // found
		}
	}
```
or
```suggestion
#ifdef _DEBUG
	if (QResource::registerResource(cExeDir() + u"/tresources.rcc"_q)) {
		return;
	}
#endif

	const auto standardPaths = QStandardPaths::standardLocations(
		QStandardPaths::GenericDataLocation);

	for (const auto directory : standardPaths) {
		if (QResource::registerResource(
			directory + u"/tdesktop/tresources.rcc"_q)) {
			return;
		}
	}
```

john-preston:
Ok, my builds don't support such loading anyway. Can be reopened if it is really needed by more than one downstream maintainers.

PR #8686, opened 2020-10-01
Little cleanup for Linux platform code & build

ilya-fedin:
CheckCXXSourceCompiles is not needed anymore

Material wayland decorations could be checked just with IsQtPluginsBundled

### DIFF
--- a/Telegram/SourceFiles/platform/linux/specific_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/specific_linux.cpp
@@ -1368,15 +1368,11 @@ void start() {
 		"this may lead to font issues.");
 #endif // DESKTOP_APP_USE_PACKAGED_FONTS
 
-	if(IsStaticBinary()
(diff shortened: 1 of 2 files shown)


Telegram/SourceFiles/platform/linux/specific_linux.cpp
```diff
@@ -1368,11 +1368,7 @@ void start() {
 		"this may lead to font issues.");
 #endif // DESKTOP_APP_USE_PACKAGED_FONTS
 
-	if(IsStaticBinary()
-		|| InAppImage()
-		|| InFlatpak()
-		|| InSnap()
-		|| IsQtPluginsBundled()) {
+	if(IsQtPluginsBundled()) {
```
john-preston, line 1371:
```suggestion
	if (IsQtPluginsBundled()) {
```

john-preston (changes requested):

PR #8706, opened 2020-10-02
Automatic Build App

BLUE-DEVIL1134:
**I was facing problem in build,
so i made this change..**

### DIFF
--- a/docs/build.py
+++ b/docs/build.py
@@ -1,0 +1,172 @@
+from os import system as cmd
+
+"""
+Hello, This Is Akash Pattnaik From India.
+I Was Facing A Problem While Building TeleGram
+So I Decided To Make A Script Which Auto Builds The
+App And We Don\'t Need To Write Each Command Again And Again...
+
+
+N.B : I Made It For Those Who Are Facing Problem While
+      Coping Each And Every Code From Github...
+      Its Really Easy...
+
+Github : https://github.com/BLUE-DEVIL1134/
+Telegram : https://t.me/AKASH_AM1
+"""
+
+# Next Part
+cmd(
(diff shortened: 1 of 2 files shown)


docs/build.py
```diff
@@ -0,0 +1,172 @@
...
+    """SET PATH=%cd%\\ThirdParty\\Strawberry\\perl\\bin;%cd%\ThirdParty\\Python27;%cd%\\ThirdParty\\NASM;%cd%\\ThirdParty\\jom;%cd%\\ThirdParty\\cmake\\bin;%cd%\\ThirdParty\\yasm;%PATH%""")
+cmd("git clone --recursive https://github.com/telegramdesktop/tdesktop.git")
+cmd("cls")
+cmd("mkdir Libraries")
+cmd("cd Libraries")
+cmd("SET LibrariesPath=%cd%")
+cmd("git clone https://github.com/desktop-app/patches.git")
+cmd("cd patches")
+cmd("git checkout ddd4084")
+cmd("cd ..")
```
ratijas, line 29: At this point, why would you use python at all? Besides, plain `*.bat` for cmd would benefit from not having to escape every special character in strings, like backslash (\\).

ratijas (commented):
See my comment below.

docs/build.py
BLUE-DEVIL1134: well,
i am a python programmer and dont know batch much..
so i did this...

xvitaly: All cmd() calls must be replaced by plain Python functions.

ratijas: >  i am a python programmer and dont know batch much..
> so i did this...

Well, this is all literally a batch, but wrapped inside python `cmd` calls. It gains nothing from being wrapped like that.

john-preston:
Sorry, I'd say that all you need to do is copy-paste all lines from original instruction into a .bat script - and that's all. No need for Python and new information, including your github and telegram profile links.

PR #8726, opened 2020-10-04
Fallback to D-Bus methods if XCB-based LastUserInputTime failed

ilya-fedin:
Before 6635d03818c90f62cfddb516acadc293255b8905 XCBLastUserInputTime could fail only when there are no X connection, after that it is clear that it can fail even with X connection and fallback is needed.


john-preston:
Thanks!

PR #8741, opened 2020-10-04
Count audio with zero duration as Music

### LINKED ISSUE #8226: Heavy audio file without specified duration is missing controls
### Steps to reproduce
1. Upload mp3 audio via bot api with >10MB size and unspecified duration
2. Try to play this track

### Expected behaviour
As in mobile and swift client: 
Audio file is playable, and you can switch tracks back and forth and track is visible in playlist.
Swift client:
image
Telegram android:
image

### Actual behaviour
In Telegram desktop audio is playable but buttons for switching back and forth is not available.
Also track is not visible in playlist when listening to other tracks.
image

### Configuration
**Operating system:**
Archlinux
2.1.13
**Installation source (Linux Only)** - the official website / GitHub releases / flatpak / snap / distribution package:
community repository
**Used theme**:
default night mode
</details>

vlakam:
Known issues: 
- when chat is loaded first time all music with zero duration is still listed as files

Fixes #8226

### DIFF
--- a/Telegram/SourceFiles/data/data_document.cpp
+++ b/Telegram/SourceFiles/data/data_document.cpp
@@ -1505,7 +1505,7 @@ bool DocumentData::isAudioFile() const {
 
 bool DocumentData::isSharedMediaMusic() const {
 	if (const auto songData = song()) {
-		return (songData->duration > 0);
+		return (songData->duration >= 0);
 	}
 	return false;
 }


john-preston:
This method should behave the same way as server does, because in case of not loaded part of message history we load shared media list from server. As far as I remember, server doesn't add duration=0 files to shared music section. If this is true, then with that change there will be desync between local list and server side list, and some messages will or will not be in the list depending on whether we scrolled till them in the message history, which is not nice.

john-preston:
But I should check once again, maybe the behaviour on the server has changed.

vlakam:
FYI
https://github.com/DrKLO/Telegram/blob/master/TMessagesProj/src/main/java/org/telegram/messenger/MessageObject.java#L4500

Android client considers audio documents without voice flag as music OR if supplied mime looks like music.

https://github.com/TelegramMessenger/Telegram-iOS/blob/7fb0b48ded4b83297453444c9aba89139c4a726a/submodules/SyncCore/Sources/TelegramMediaFile.swift#L479
IOS  client seems to consider all audio documents as music if voice flag is false

Both of them don't look at duration

john-preston:
Yes, this will be changed on the server and in the next version. But I've changed it a little bit differently, like you say, without looking on duration at all.

PR #8829, opened 2020-10-14
building-xcode.md: remove gyp

a-x-:


ilya-fedin:
Strange PR, if you remove gyp from the instruction, the later step with crashpad will fail

a-x-:
Yep, Please help to complete update the manual

I don't know how you build it

a-x-:
`gyp` is not used now as I know

ilya-fedin:
> `gyp` is not used for tdesktop now as I know

Well, that's false

a-x-:
so, we came back to issue.
Gyp installation is broken

john-preston:
I've ran right now:
```
git clone https://chromium.googlesource.com/external/gyp
cd gyp
git checkout 9f2a7bb1
git apply ../patches/gyp.diff
```

on macOS and it ran without any issues 🤔

PR #8866, opened 2020-10-20
typo in LEGAL

philsmd:
This seems to be a small typo in the `LEGAL` file. I suggest changing `of` to `or`.

Thx


john-preston:
I guess this is the phrase that is used everywhere:

https://people.gnome.org/~markmc/openssl-and-the-gpl.html - specific explanation

https://www.gnu.org/software/ccrtp/doc/refman/html/SrtpSymCrypto_8cpp_source.html

https://enterprise.dejacode.com/licenses/public/openssl-exception-gpl-2.0/

So I'll leave it that way.

PR #8898, opened 2020-10-26
Check Qt version in runtime in CanOpenDirectoryWithPortal

ilya-fedin:
It is not guaranteed that app is running with the same Qt version that was compiled


xvitaly:
> It is not guaranteed that app is running with the same Qt version that was compiled

Telegram Desktop will have major issues when running with different Qt version due to Qt private API usage. You can add a Qt version check on startup and terminate the application if they don't match.

ilya-fedin:
>  You can add a Qt version check on startup and terminate the application if they don't match.

This doesn't have any sense and unrelated

xvitaly:
> This doesn't have any sense and unrelated

Just a proposal for the new pull request.

john-preston:
Thanks!

PR #8911, opened 2020-10-27
Use crl::on_main in MainWindow::sniSignalEmitted

ilya-fedin:


Telegram/SourceFiles/platform/linux/main_window_linux.cpp
```diff
@@ -592,9 +592,13 @@ void MainWindow::sniSignalEmitted(
 		gchar *sender_name,
 		gchar *signal_name,
 		GVariant *parameters,
-		MainWindow *window) {
+		gpointer user_data) {
```
23rd, line 595: Why did you throw away the MainWindow pointer?

ilya-fedin: @john-preston told me that it is dangerous to get MainWindow pointer from a non-main thread

23rd: Maybe it would be better to use lambda instead of `static void sniSignalEmitted` to capture the MainWindow pointer?

ilya-fedin: MainWindow::initHook is pretty big, I'm not sure that it is a good idea to make it bigger

john-preston: @23rd The fix is about this method being called from some non-main thread. In that case there is no safe way of using raw MainWindow pointer.

john-preston:
Thanks!

PR #8930, opened 2020-10-29
Ignore additional video streams inside an audio file

### LINKED ISSUE #4327: Telegram client uploads mp3 as file
### Steps to reproduce
1. Try to add mp3 file (I'd rather say the bug relates to files with > 10Mb size, but there's no straight correlation)

### Expected behaviour
The mp3 file is uploaded with music content-type

### Actual behaviour
The mp3 file is uploaded as file - no embedded player available

### Configuration
**Operating system:** Ubuntu 17.10

**Version of Telegram Desktop:** 1.1.23

**Used theme**: default

### LINKED ISSUE #5357: [BUG] Music uploading as files on Linux
### Steps to reproduce
1. Drag and Drop any music file to Telegram / Select a file with the clip button

### Expected behaviour
Content is uploaded as a streamable music file, with the song tags being displayed instead of the file name, as when doing the same on Windows.

### Actual behaviour
Content is uploaded as an "unknown" file. Its title is the same as the file name instead of the song's tags, and is not possible to stream the song on other clients.

### Sugestion
As it happens with pictures (send quickly or as a file), adding an option to send Music as file or as music would be a nice idea.

### Configuration
**Operating system:** ArchLinux x86_64, Kernel 4.18.16

**Version of Telegram Desktop:** 1.4.3

**Used theme**: Default

### LINKED ISSUE #5840: Some ogg/opus files are not recognized as "audio files"

andry-dev:
On Linux, audio files that embed a cover art will not be sent with their metadata but as simple documents.
This happens because the cover art embedded in an audio file is a video stream, which makes `FFMpegAttributesReader::open()` error out. On Windows and Android (and I also guess on macOS?) this doesn't happen and the file is sent properly with its cover art.

Telegram still works, the metadata are properly encoded/decoded, I'm able to send tracks with cover arts and see them with other clients (from other OSes).

This commit fixes #5840, fixes #5357 and fixes #4327 (they're all related).

### DIFF
--- a/Telegram/SourceFiles/media/audio/media_audio.cpp
+++ b/Telegram/SourceFiles/media/audio/media_audio.cpp
@@ -1531,15 +1531,12 @@ public:
 		int res = 0;
 		char err[AV_ERROR_MAX_STRING_SIZE] = { 0 };
 
-		int videoStreamId = av_find_best_stream(fmtContext, AVMEDIA_TYPE_VIDEO, -1, -1, &codec, 0);
-		if (videoStreamId >= 0) {
-			DEBUG_LOG(("Audio Read Error: Found video stream in file '%1', data size '%2', error %3, %4").arg(_file.name()).arg(_data.size()).arg(videoStreamId).arg(av_make_error_string(err, sizeof(err), streamId)));
-			return false;
-		}
-
 		for (int32 i = 0, l = fmtContext->nb_streams; i < l; ++i) {
 			const auto stream = fmtContext->streams[i];
 			if (stream->disposition & AV_DISPOSITION_ATTACHED_PIC) {
+				if (!_cover.isNull()) {
+					continue;
+				}
 				const auto &packet = stream->attached_pic;
 				if (packet.size) {
 					const auto coverBytes = QByteArray(
@@ -1555,9 +1552,15 @@ public:
 					if (!_cover.isNull()) {
 						_coverBytes = coverBytes;
 						_coverFormat = format;
-						break;
 					}
 				}
+			} else if (stream->codecpar->codec_type == AVMEDIA_TYPE_VIDEO) {
+				DEBUG_LOG(("Audio Read Error: Found video stream in file '%1', data size '%2', error %3, %4")
+					.arg(_file.name())
+					.arg(_data.size())
+					.arg(i)
+					.arg(av_make_error_string(err, sizeof(err), streamId)));
+				return false;
 			}
 		}
 


ilya-fedin:
How did you install tdesktop? tdesktop source code is written for ffmpeg 3.x which is included into official static binary. If you installed tdesktop via any other source, it may use ffmpeg 4.x and behave wrong. Please check with the official binary.

ilya-fedin:
Oh, Ubuntu 17.10 contains ffmpeg 3.x, so this couldn't be the cause. Although, tdesktop official binary's ffmpeg is configured with a very little set of codecs and that could be the reason why the difference is there. If you confirm this by checking if the bug is not present on the official binary, then the proper fix is to find which exactly video codec it finds in mp3 files and exclude it frothe check

andry-dev:
I checked with the official binary and the bug is _not_ there, metadata is handled properly.

Up until now I used my distro's package manager (Void Linux) which by default compiles TDesktop with ffmpeg 4.3; other rolling release distros (like Arch) do the same because they don't include older versions of ffmpeg. Also, the same problem happens with the flatpak version from flathub (I tested it thinking it would be the same as the static binary, my bad, I should've checked out the static bin before anything else).

ilya-fedin:
Well, I think you should find which codec should be excluded rather than deleting the whole check

andry-dev:
Welp, I guess you're right.
The most commonly used ones for cover arts are still JPEGs and PNGs, they should be enough for most use-cases.

I'll go back to it, thank you.

ilya-fedin:
I've heard that ffplay is able to open images. Maybe it finds the image in the file and returns it with AVMEDIA_TYPE_VIDEO since there is no AVMEDIA_TYPE_IMAGE exist? Maybe there are some method to get the mimetype of the stream and check if it starts with "image/" :thinking:

ilya-fedin:
Looks like I found something
https://stackoverflow.com/a/24000492

andry-dev:
I ran `ffplay` on a test file and these are the results:

```
Input #0, flac, from 'test.flac':  0KB vq=    0KB sq=    0B f=0/0
  Metadata:
    ALBUM           : 愧人贖悪 ～ Evil People as the True Object of Salvation ～
    album_artist    : Demetori
    ARTIST          : Demetori
    BPM             : 182
    GENRE           : Game
    TITLE           : Introduction ~ Double Dealing
    TRACKTOTAL      : 9
    EVENT           : C88
    DATE            : 2015.08.14
    track           : 1
  Duration: 00:00:55.75, start: 0.000000, bitrate: 500 kb/s
    Stream #0:0: Audio: flac, 44100 Hz, stereo, s16
    Stream #0:1: Video: mjpeg (Baseline), yuvj444p(pc, bt470bg/unknown/unknown), 600x600, 90k tbr, 90k tbn, 90k tbc (attached pic)
    Metadata:
      comment         : Cover (front)
      title           : 290.jpeg
```

As you can see the second stream contains the image as a jpeg and it reports it as a video, so you're right on the `AVMEDIA_TYPE_VIDEO`; I'm not familiar with ffmpeg's API but I think there should be a function to return the `mjpeg` filetype to the program, I'll have to look up ffmpeg's docs.

andry-dev:
> Looks like I found something
Yup, I think that's it.

ilya-fedin:
This should be even more helpful:
https://github.com/PHP-FFMpeg/PHP-FFMpeg/pull/687
https://github.com/FFmpeg/FFmpeg/blob/release/4.2/libavformat/movenc.c#L149-L154

ilya-fedin:
This should work
```diff
diff --git a/Telegram/SourceFiles/media/audio/media_audio.cpp b/Telegram/SourceFiles/media/audio/media_audio.cpp
index 6e4ad2d17..2cf452c96 100644
--- a/Telegram/SourceFiles/media/audio/media_audio.cpp
+++ b/Telegram/SourceFiles/media/audio/media_audio.cpp
@@ -1531,15 +1531,12 @@ public:
 		int res = 0;
 		char err[AV_ERROR_MAX_STRING_SIZE] = { 0 };
 
-		int videoStreamId = av_find_best_stream(fmtContext, AVMEDIA_TYPE_VIDEO, -1, -1, &codec, 0);
-		if (videoStreamId >= 0) {
-			DEBUG_LOG(("Audio Read Error: Found video stream in file '%1', data size '%2', error %3, %4").arg(_file.name()).arg(_data.size()).arg(videoStreamId).arg(av_make_error_string(err, sizeof(err), streamId)));
-			return false;
-		}
-
 		for (int32 i = 0, l = fmtContext->nb_streams; i < l; ++i) {
 			const auto stream = fmtContext->streams[i];
 			if (stream->disposition & AV_DISPOSITION_ATTACHED_PIC) {
+				if (!_cover.isNull()) {
+					continue;
+				}
 				const auto &packet = stream->attached_pic;
 				if (packet.size) {
 					const auto coverBytes = QByteArray(
@@ -1555,9 +1552,15 @@ public:
 					if (!_cover.isNull()) {
 						_coverBytes = coverBytes;
 						_coverFormat = format;
-						break;
 					}
 				}
+			} else if (stream->codec->codec_type == AVMEDIA_TYPE_VIDEO) {
+				DEBUG_LOG(("Audio Read Error: Found video stream in file '%1', data size '%2', error %3, %4")
+					.arg(_file.name())
+					.arg(_data.size())
+					.arg(i)
+					.arg(av_make_error_string(err, sizeof(err), streamId)));
+				return false;
 			}
 		}
 
```

andry-dev:
I was waiting for my code to recompile after a clean build but you sniped me with your solution lol.

Your solution works! But it gives me a warning saying that `->codec_type` is deprecated, should I ignore it or switch to `avcodec_find_encoder`?

ilya-fedin:
> But it gives me a warning saying that `->codec` is deprecated, should I ignore it?

If you know how to do this in a non-deprecated manner, that would better of course :)

andry-dev:
Ok then, I'll use `->codecpar`. I thought that maybe that would cause problems with ffmpeg 3.x.

I'll force-push now, thank your for your help.

ilya-fedin:
> I thought that maybe that would cause problems with ffmpeg 3.x.

I've checked that codecpar is present in ffmpeg 3.4, so should be ok

ilya-fedin:
> This commit fixes #5840, #5357 and #4327 (they're all related).

Can you prepend "fixes" to each issue number (github is too stupid to parse that)?

andry-dev:
>Can you prepend "fixes" to each issue number (github is too stupid to parse that)?

Done, do you want me to do the same in the commit message so it's all uniform?

ilya-fedin:
> Done, do you want me to do the same in the commit message so it's all uniform?

No, this is enough to close all the issues automatically

PR #8964, opened 2020-11-01
Add flatpak and snap update URLs

ilya-fedin:


Telegram/SourceFiles/core/update_checker.cpp
```diff
@@ -1575,9 +1576,16 @@ void UpdateApplication() {
...
 #elif defined OS_MAC_STORE // OS_WIN_STORE
 			return "https://itunes.apple.com/ae/app/telegram-desktop/id946399090";
-#else // OS_WIN_STORE || OS_MAC_STORE
+#elif defined Q_OS_UNIX && !defined Q_OS_MAC // OS_WIN_STORE || OS_MAC_STORE
+			if (Platform::InFlatpak()) {
+				return "https://flathub.org/apps/details/org.telegram.desktop";
+			} else if (Platform::InSnap()) {
+				return "https://snapcraft.io/telegram-desktop";
+			}
+			return "https://desktop.telegram.org";
```
xvitaly, line 1585: Please allow GNU/Linux maintainers to specify their own URL here using Cmake.

ilya-fedin: I'm not sure this is a good idea

Issue #8810, opened 2020-10-12
[Feature Request] Scale Telegram Desktop in Smaller Screens


antony-jr:
### Is your feature request related to a problem?

In Phosh + Postmarket OS telegram scales well but some content is capped, need to scale smaller to about at least 700 px width.

### Describe the solution you'd like

Scale Telegram Desktop properly as per the screen size. i.e Scale Telegram application to at least a minimum of 700 px width and respective height. 

### Describe alternatives you've considered

None.

### Additional context

The right side of Telegram seems to be cut off only by a bit.

![photo_2020-10-13_02-46-06](https://user-images.githubusercontent.com/33930174/95791131-527bbe80-0cfe-11eb-9d7b-24953d5701ea.jpg)

ilya-fedin:
As already been discussed in #8058, tdesktop is not designed to be used on phones nor on any other touch devices and this is intended.

antony-jr:
Seems like I have to fork or patch this on my own and release it in a different channel for phosh users. Okay Thank you for your time.

Aokromes:
you may want to use kotatogram with sub 100% scaling and larger fonts.

allnamesaretakeningithub:
> As already been discussed in #8058, tdesktop is not designed to be used on phones nor on any other touch devices and this is intended.

@ilya-fedin may I ask:

- Why is this intended?
- Should that information not be in a FAQ?

Also, with the Librem 5 phones and pinephones, we are taking about having the telegram desktop application on a GNU/Linux machine that happens to have a 5 inch touchscreen as main interface.

What both closed requests ask in short is to allow a slightly smaller window so that telegram-desktop for linux fits better on a linux device that has network connection (and that "accidentally" also makes phone calls).

One last question is: 

If you do not intend to provide support for touch devices, will you provide any type of support for example for tdesktop on GNU/Linux tablets? Like the pinetablet?

antony-jr:
@allnamesaretakeningithub 6 and 6.5 inch screen fits Telegram desktop nicely.

allnamesaretakeningithub:
>  6 and 6.5 inch screen fits Telegram desktop nicely.

@antony-jr that does not reply do some of my questions. As the statement made was about no support for: touch devices. 

Nor does it answer to the question of what is the reasoning behind it, nor if it sohuld not be in a FAQ that info

Issue #8058, opened 2020-06-12
[Limitation/niche use] If the minimum vertical size were a tad smaller, Telegram would work a lot better on the PinePhone


ell1e:
I am trying to use Telegram on the [PinePhone](https://wiki.pine64.org/index.php/PinePhone) with an ARM64 build provided by [Alpine Linux](https://pkgs.alpinelinux.org/packages?name=telegram-desktop) which I believe runs this code base here.

When using it, there is a problem that even in portrait mode, vertically Telegram Desktop just about won't fit the screen which results in missing elements:

![test](https://user-images.githubusercontent.com/64124388/84539028-3e896700-ace2-11ea-933b-e8481780f046.png)

As you can see, with the virtual keyboard hidden the input bar comes back:

![test1](https://user-images.githubusercontent.com/64124388/84539487-059dc200-ace3-11ea-88a9-d8ce4d1cdfb6.png)

This is a bit sad since otherwise the layout is responsive enough that it works almost perfectly as a phone app. (Please note this may confuse you since obviously it's named "telegramdesktop", but the PinePhone is a Linux phone so it doesn't really distinguish anymore between phone and desktop apps.) So it would help a lot if this could be changed, since it otherwise works fine!

### Steps to reproduce
1. Install Telegram on a PinePhone, or just resize the window manually on your desktop to simulate
2. See that vertical space doesn't really go to very small, which on the PinePhone leads to the keybaord covering the input bar making typing almost impossible

### Expected behaviour
Telegram window resizes sufficiently to fit on screen, hiding elements as necessary.

### Actual behaviour
Telegram window *almost* resizes as needed, but with the virtual keyboard open the input bar is cut off.

### Configuration
**Operating system:** [postmarketOS (ARM64 Linux with Wayland-based desktop)](https://postmarketos.org)

**Version of Telegram Desktop:** 2.1.11-r0

**Installation source (Linux Only)** Alpine Linux Package

**Used theme**: "tinted"

<details><summary><b>Logs</b>:</summary>
(I don't think log is relevant)
</details>

Aokromes:
use sub 100% scaling.

ilya-fedin:
> As you can see, with the virtual keyboard hidden the input bar comes back:

If you know how to fix it, let's PR a fix.
#3274

ell1e:
[...]

PR #9007, opened 2020-11-06
data: List compatible form-factors

Kekun:
This will allow the app to be available in Phosh and PureOS Store.

### DIFF
--- a/lib/xdg/telegramdesktop.appdata.xml.in
+++ b/lib/xdg/telegramdesktop.appdata.xml.in
@@ -66,4 +66,8 @@
     <provides>
         <binary>telegram-desktop</binary>
     </provides>
+    <custom>
+        <value key="Purism::form_factor">workstation</value>
+        <value key="Purism::form_factor">mobile</value>
+    </custom>
 </component>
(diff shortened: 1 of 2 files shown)


ilya-fedin:
I'm not sure this is valid since tdesktop doesn't aim to adapt for smartphones

Nirzak:
> I'm not sure this is valid since tdesktop doesn't aim to adapt for smartphones
> 
> Not to say that tdesktop minimal width and height are too big for these form-factors, like been reported by #8058
> 
> So, this won't correspond to reality. But you still could PR this only with "Workstation" and it will be fine I think

Yeah, I agree with that. It should only for workstations or tablets, not mobile devices. As such small screen is not feasible for tdesktop.

ilya-fedin:
@john-preston maybe close this? :thinking:

kinglyonais (approved):

john-preston:
@Kekun Are you sure it works fine and fits on mobile devices that allow to install apps from those "Phosh and PureOS Store"? If not, please remove mobile compatibility mentioning, if yes, I'll merge without changes.

ilya-fedin:
#8058, #8810 and #10193 definitely show that it's not

ilya-fedin:
why no-response doesn't close that :hmm:

john-preston:
😞

PR #9009, opened 2020-11-06
Updated Qt version in GitHub Actions workflows

zurg3:


ilya-fedin:
Looks like this doesn't work

Scrxtchy:
Since the workflow files have been changed the cache key for them would also be changed.
This means that the entire workflow is rebuilding the dependencies fresh. As the issue is related to an FFMPEG error and not a Qt one, the source of the error doesn't so much pertain to the patch that is supplied.

The error refers to the declaration of `av_muxer_iterate`, which appears in 
https://github.com/FFmpeg/FFmpeg/blob/5c7823ff1cffff55f25169492d263bca934880a2/libavformat/allformats.c#L535
and should be imported properly in 
https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/SourceFiles/ffmpeg/ffmpeg_utility.h#L19

However, I'm unable to replace this bug, as the current project is unable to actually find qt5.
This is due to a change at https://github.com/desktop-app/cmake_helpers/commit/3d2b39935caf747a48a0430e5bbfe160320b3397 that updates the qt version in use, something this patch is meant to fix.

Since I'm not going to sit through an hour of a workflow again since updating the workflow to match this pull request, gaining support for version of qt that is linked to in the repo, I would have to just put down my theory that there is an issue with the ffmpeg portion of the build on all platforms, this could just be a one-off issue or not. The last time the build scripts have were updated was 10 days ago (or 13 for a commit with an action), and both these seem to be in working order.

NB I only checked the windows build of this

ilya-fedin:
@Scrxtchy ffmpeg just need an update as well

ilya-fedin:
Works right now, linux action should be migrated to the docker build and the most convenient way is to use GitHub Packages I guess. This could be done in another PR if @john-preston doesn't mind it.

john-preston:
Thanks!

PR #9022, opened 2020-11-08
Add support for open with on linux

ilya-fedin:


magnus-gross:
Great stuff! Does this work on Qt DEs as well (e.g. KDE) and if so, what do I need to have installed on the system?

ilya-fedin:
> Does this work on Qt DEs as well (e.g. KDE) and if so, what do I need to have installed on the system?

I took the implementation from Engrampa, you could test with it.

mikhailnov:
It's a great feature, thanks a lot!

john-preston:
Thanks, that's cool.

Issue #6721, opened 2019-10-27
[Feature Request] Add option to disable red counter on systray icon


petRUShka:
### Is your feature request related to a problem?
It would be great to have an option to get rid of red counter on telegram systray icon:

![2019-10-27-184910_29x27_scrot](https://user-images.githubusercontent.com/146893/67637345-ce60d900-f8ea-11e9-842d-8afbb2e55983.png)

The idea is to don't distract yourself more than you want to be distracted (especially with minimized notifications options).

### Describe the solution you'd like
Add above-mentioned option and remove red counter.

### Describe alternatives you've considered
There are two workarounds at the moment:
1. Disable systray icon at all. Cons: if you have to minimize Telegram you should close it. And then you have to wait till it starts (could be long if you have to use Proxy in your country).
2. Mute all contacts without exclusion. Isn't easy to maintain that setup.

### Additional context
Why is so important to have option to not have red counter and decrease distraction? Tristan Harris former Product Philosopher at Google [has explained](https://medium.com/thrive-global/how-technology-hijacks-peoples-minds-from-a-magician-and-google-s-design-ethicist-56d62ef5edf3).

petRUShka:
I believe this is still an issue. So It would be great to have it done.

varlesh:
![image](https://user-images.githubusercontent.com/8083855/103255353-74521c80-49a2-11eb-8d24-430e95b46bf1.png)
This red badge overlap Telegram logo. I can't understand what is this? Skype? Telegram? or other app?
Why needed know number of messages? It doesn't change anything, but it spoils the whole style.
1 msg or 1400 msg... who cares?

varlesh:
Please devs added option disable this red badge or revert env `TDESKTOP_DISABLE_TRAY_COUNTER=1`
![image](https://user-images.githubusercontent.com/8083855/103256298-2808db80-49a6-11eb-8099-9cb216e530c1.png)

marcoValkyriweb:
@varlesh your overlap is quite large, mine is very small compared to yours:
![image](https://user-images.githubusercontent.com/34270069/103295522-d50e4300-49fc-11eb-9b5b-d66f88459f60.png)

AndydeCleyre:
Adding the effect of upgrading from `2.4.7` to `2.5.1` in my context. Before:

![image](https://user-images.githubusercontent.com/1787385/103310007-953b6180-49e4-11eb-90ef-6a6a95f305e7.png)

After (currently):

[...]

PR #9056, opened 2020-11-12
Remove the last workaround in tray implementation

ilya-fedin:
Since tdesktop gets icon theme pretty well now, there's no need for any workaround.

### DIFF
--- a/Telegram/SourceFiles/platform/linux/main_window_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/main_window_linux.cpp
@@ -52,12 +52,9 @@ extern "C" {
 } // extern "C"
 #endif // !DESKTOP_APP_DISABLE_DBUS_INTEGRATION
 
-#include <glib.h>
-
 namespace Platform {
 namespace {
 
-constexpr auto kDisableTrayCounter = "TDESKTOP_DISABLE_TRAY_COUNTER"_cs;
 constexpr auto kPanelTrayIconName = "telegram-panel"_cs;
 constexpr auto kMutePanelTrayIconName = "telegram-mute-panel"_cs;
 constexpr auto kAttentionPanelTrayIconName = "telegram-attention-panel"_cs;
@@ -201,13 +198,6 @@ QIcon TrayIconGen(int counter, bool muted) {
 
 	const auto iconName = GetTrayIconName(counter, muted);
 
-	if (qEnvironmentVariableIsSet(kDisableTrayCounter.utf8())
-		&& !iconName.isEmpty()) {
-		const auto result = QIcon::fromTheme(iconName);
-		UpdateIconRegenerationNeeded(result, counter, muted, iconThemeName);
-		return result;
-	}
-
 	QIcon result;
 	QIcon systemIcon;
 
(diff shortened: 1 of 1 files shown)


zeparu:
just got the update today, and the telegram icon still wont have the correct color, always defaulting to the black icon on all icon themes (except on the ones that dont have monochrome tray icons)
image

ilya-fedin:
@zeparu i can only sugest to switch to dark icon theme version, unfortunately.

efeciftci:
I'm afraid by removing the `setIconByName()` function calls, Telegram has lost the ability to be properly themed by KDE Plasma; which has been explained here: https://techbase.kde.org/Development/Tutorials/Plasma5/ThemeDetails#.22icons.22_folder

And besides, it seems this commit has also broken disabling badge on tray count because no matter what I do in Settings -> Notifications, Telegram always displays unread message count in system tray now.

ilya-fedin:
> I'm afraid by removing the `setIconByName()` function calls, Telegram has lost the ability to be properly themed by KDE Plasma; which has been explained here: https://techbase.kde.org/Development/Tutorials/Plasma5/ThemeDetails#Theming_Application_Icons_in_the_System_Tray

Yeah. This ability was temporary until we decide what to do with it. @john-preston told me that he doesn't want to introduce the ability to disable tray counter, so the temporary env var was removed.

> And besides, it seems this commit has also broken disabling badge on tray count because no matter what I do in Settings -> Notifications, Telegram always displays unread message count in system tray now.

There's no ability to disable the badge (#6721), the option you mentioned just has confusing name and counts messages instead ol chats in enabled state

ilya-fedin:
Just proposed another translation of the option.
https://translations.telegram.org/en/tdesktop/settings/lng_settings_count_unread

renyhp:
> @john-preston doesn't want to introduce the ability to disable tray counter

I am really curious as to why. More options means more flexibility and more users getting what they want. If anyone wants their tray counter, it's enabled by default. This way, people who don't want it are stuck with a bad UX.

> the telegram icon still wont have the correct color

I am getting issues too, my icon is blurry. Both these experiences prove that Telegram is not currently able to "get icon theme pretty well". It looks like the workaround removed by this PR is very much needed, or, alternatively, a fix for this bug is needed.

ilya-fedin:
> I am really curious as to why. More options means more flexibility and more users getting what they want.

preston wants to have as less as possible options to keeps things simpler for users :(
Sadly, but flexibility is not about tdesktop.

> It looks like the workaround removed by this PR is very much needed, or, alternatively, a fix for this bug is needed.

No one reported any bugs without the workaround, so it been removed.

renyhp:
> No one reported any bugs without the workaround, so it been removed.

As soon as the workaround was removed, the issue appeared and I reported #9982, which you closed.

ilya-fedin:
> As soon as the workaround was removed, the issue appeared and I reported #9982, which you closed.

You reported that the variable doesn't work, what is not a bug.

renyhp:
My issue was not clear enough, I guess, I apologize. If I open another one to report issues with the panel icon, will it stay open?

ilya-fedin:
Yes. But one problem = one issue.

PR #9087, opened 2020-11-17
Add -Werror to actions

ilya-fedin:


john-preston:
I've cherry-picked it, thanks.

PR #9769, opened 2020-11-19
Restore 16px tray icon size

ilya-fedin:
Looks like there are support for this size since b703f4e5557053d4cc320e42861db197c0ce7790


john-preston:
👍

PR #9792, opened 2020-11-23
Detect more dark themes on Linux

ilya-fedin:
QGnomePlatform changed its detection method as well https://github.com/FedoraQt/QGnomePlatform/commit/23ee752d31069c222ab2e05ce4c37482fc22ddf4

### DIFF
--- a/Telegram/SourceFiles/platform/linux/specific_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/specific_linux.cpp
@@ -779,7 +779,7 @@ std::optional<bool> IsDarkMode() {
 
 		const auto themeName = Libs::GtkSetting("gtk-theme-name").toLower();
 
-		if (themeName.endsWith(qsl("-dark"))) {
+		if (themeName.contains(qsl("-dark"))) {
 			return true;
 		}
 


vimpostor:
BTW Qt already contains some dark theme checking in its private API. Although it is a bit hacky (it basically checks if the text color of the platform theme is closer to white or black), it works without relying on the naming scheme of the theme: https://code.woboq.org/qt5/qtquickcontrols2/src/quickcontrols2/qquickstyle.cpp.html#_ZL23qt_is_dark_system_themev

```c++
static bool qt_is_dark_system_theme()
{
  if (const QPlatformTheme *theme = QGuiApplicationPrivate::platformTheme()) {
    if (const QPalette *systemPalette = theme->palette(QPlatformTheme::SystemPalette)) {
      const QColor textColor = systemPalette->color(QPalette::WindowText);
      return textColor.red() > 128 && textColor.blue() > 128 && textColor.green() > 128;
    }
  }
  return false;
}
```

That being said, Qt 6 will contain non-private API for dark system theme support, so maybe it will be better to use it when Telegram Desktop transitions to Qt 6.

ilya-fedin:
> BTW Qt already contains some dark theme checking in its private API. Although it is a bit hacky (it basically checks if the text color of the platform theme is closer to white or black), it works without relying on the naming scheme of the theme:

The API you provided checks if QML theme is dark, tdesktop doesn't use QML.

> That being said, Qt 6 will contain non-private API for dark system theme support, so maybe it will be better to use it when Telegram Desktop transitions to Qt 6.

It's very unlikely that tdesktop will be ported to Qt 6.

vimpostor:
> > BTW Qt already contains some dark theme checking in its private API. Although it is a bit hacky (it basically checks if the text color of the platform theme is closer to white or black), it works without relying on the naming scheme of the theme:
> 
> The API you provided checks if QML theme is dark, tdesktop doesn't use QML.
> 

It does not. Yes, the API is inside the QML module, but it checks if the default platform theme, which is the default system theme of the platform which is independent of QML, is dark.

ilya-fedin:
> but it checks if the default platform theme, which is the default system theme of the platform which is independent of QML, is dark.

tdesktop can't use platform themes, statically linked Qt is unable to load plugins, so this API is useless.

vimpostor:
OK that's unfortunate :(

PR #9802, opened 2020-11-24
Media Play key is expected to behave like Toggle Media Play/Pause

### LINKED ISSUE #3342: Play/pause button on keyboard doesn't work as expected
### Steps to reproduce
1. Open music with Telegram internal player
2. Click the play/pause button on the keyboard

### Expected behaviour
Music should stop, or at least Telegram should ignore the button at all.

### Actual behaviour
Music volume goes down for a second, then it rise up at previous level. Music doesn't stop. With music paused, the key works as expected (it plays the paused music).

### Configuration
**Operating system:** Windows 10 1703, I had the same issue with all previous versions.

**Version of Telegram Desktop:** 1.0.34 alpha and all of previous versions.

**Used theme**: default.

**Logs**: https://pastebin.com/hcSNQPrM

ilya-fedin:
Looks like Qt maps "Media Play" to the common play/pause button on keyboards:
https://github.com/qt/qtbase/blob/5.15/src/platformsupport/input/xkbcommon/qxkbcommon.cpp#L293
https://github.com/qt/qtbase/blob/5.15/src/plugins/platforms/windows/qwindowskeymapper.cpp#L392
https://github.com/qt/qtbase/blob/5.15/src/plugins/platforms/android/androidjniinput.cpp#L519
https://github.com/qt/qtbase/blob/5.15/src/plugins/platforms/android/androidjniinput.cpp#L607

While "Toggle Media Play/Pause" is some special case:
https://github.com/qt/qtbase/blob/5.15/src/plugins/platforms/windows/qwindowskeymapper.cpp#L488
https://github.com/qt/qtbase/blob/5.15/src/plugins/platforms/ios/quiview.mm#L584

It is worth to behave on "Media Play" like on "Toggle Media Play/Pause".

Fixes #3342

### DIFF
--- a/Telegram/SourceFiles/media/player/media_player_instance.cpp
+++ b/Telegram/SourceFiles/media/player/media_player_instance.cpp
@@ -733,7 +733,7 @@ void Instance::setupShortcuts() {
 	) | rpl::start_with_next([=](not_null<Shortcuts::Request*> request) {
 		using Command = Shortcuts::Command;
 		request->check(Command::MediaPlay) && request->handle([=] {
-			play();
+			playPause();
 			return true;
 		});
 		request->check(Command::MediaPause) && request->handle([=] {


john-preston:
Thanks!

PR #9915, opened 2020-12-14
Update VS2019 build instructions

Himself65:
there is a submodule that need to clone, see https://github.com/desktop-app/tg_owt/tree/master/src/third_party


john-preston:
Can you change it to `git clone --recursive`? Because we use this form in all other places.

john-preston:
Thanks.

PR #10062, opened 2021-01-04
Fix kSNIWatcherService/kSNIWatcherInterface misusage

ilya-fedin:
Even though they're the same, there should be interface specified

### DIFF
--- a/Telegram/SourceFiles/platform/linux/main_window_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/main_window_linux.cpp
@@ -402,10 +402,8 @@ bool UseUnityCounter() {
 
 	return Result;
 }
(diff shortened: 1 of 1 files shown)


Telegram/SourceFiles/platform/linux/main_window_linux.cpp
```diff
@@ -842,25 +838,28 @@ void MainWindow::updateIconCounters() {
 #ifndef DESKTOP_APP_DISABLE_DBUS_INTEGRATION
 	if (UseUnityCounter()) {
 		const auto launcherUrl = "application://" + GetLauncherFilename();
+		// Gnome requires that count is a 64bit integer
+		qint64 counterSlice = (counter > 9999) ? 9999 : counter;
```
23rd, line 842:
```suggestion
		const qint64 counterSlice = std::min(counter, 9999);
```

Issue #7891, opened 2020-05-14
[Feature Request] Consistency - New Telegram Logo


hartkotet:
Since the new logo replaced pretty much everywhere the old one (even on the website now), it would be cool if TDesktop could "update" too. ..for the sake of consistency..

Cheers.

maxwxyz:
I have changed the icons to the new logo and created a pull request #10098

Eric-01:
The icon has finally been updated in commit 551732738bee2f8fc6b588dfc48c02bfbe22ea36 :tada:

Eric-01:
The icon has finally been updated in commit 551732738bee2f8fc6b588dfc48c02bfbe22ea36 🎉

Unfortunately, this feature request is not mentioned there, so it hasn't been automatically closed by GitHub.

PR #10098, opened 2021-01-10
changed icon to new logo #7891

maxwxyz:
Updated the outdated icon/logo of the desktop client to the current one.


Aokromes:
i think you need  to add all res png files and not only .ico file.

23rd:
I think it will not be merged anyway.

maxwxyz:
I added also all res png files.
Why would it not be merged? @23rd

23rd:
Such a fundamental thing as the application icon is handled by the Telegram team itself and their designers, no one else.

john-preston:
The icon for macOS was not updated. But indeed, when the designer that works with my app says I need to update it and sends the exact artworks - I will. Sorry.

PR #10142, opened 2021-01-15
Don't add counter when icon theme has 'panel' icon

ilya-fedin:
These icons should have a dot indicating unread messages, counter is redudant for them


varlesh:
@ilya-fedin Big thx man +100500

PR #10231, opened 2021-01-26
Fix crash with fcitx and custom titlebar

### LINKED ISSUE #10229: Invalid static_cast of QWidget
https://github.com/telegramdesktop/tdesktop/blob/b1b25b0df9d285400cee786fb22639e2ff7fc4d5/Telegram/SourceFiles/window/window_title_qt.cpp#L251

It is possible that this object is not a QWidget but a QWindow if any external plugin using non-QWidget thing for GUI. If you only want to check QWidget, use qobject_cast<QWidget *> instead. It will return null if the underlying type can not be cast to QWidget

ilya-fedin:
Fixes #10229


wengxt:
LGTM
Thanks for your fast fix!

PR #10240, opened 2021-01-28
Another attempt to implement shadows on Wayland

ilya-fedin:
Works only with patched Qt, depends on https://github.com/desktop-app/patches/pull/26


john-preston:
Thanks!

Issue #10237, opened 2021-01-27
Client won't preview uncompressed image from clipboard when enabled 'Ask download path for each file'


mooons:
### Steps to reproduce
1. Settings - Advanced - **Check** 'Ask download path for each file' (By default this is unchecked, files will be auto downloaded to user's `Downloads\Telegram Desktop` directory)
2. Restart Telegram Desktop
3. Copy an image to the clipboard (for repro, simply Win+Shift+S take a random area screenshot), then paste the image to a chat, say, "Saved Messages".
4. Uncheck 'Compress images', send.
5. Click on the upload file, it won't show the in-app preview.

### Expected behaviour
In Step 4 it should show an in-app preview of the image.

In fact, if you go back to Settings - Advanced and **uncheck** 'Ask download path for each file' back. Then click that same image file in the chat, it will show the image. I suspect this has something to do with the desktop client having to download the file first before showing preview, but when 'Ask download path for each file' is checked, it won't download the file (should expect a dialogue box to save the file, or just save it to %temp% and preview it IMO).

~~This bug may apply to videos as well.~~

### Actual behaviour
As described in Step 5.

### Configuration
**Operating system:** Windows 10 20H2 19042.746

**Version of Telegram Desktop:** ~~2.5.1~~ 2.5.8

**Used theme**: Night mode

mooons:
The issue is likely caused by the failure of `data->saveFromDataSilent()` in `data_document.cpp`:

https://github.com/telegramdesktop/tdesktop/blob/3251b8bf6e740e55777e85d63065a61e43df54fe/Telegram/SourceFiles/data/data_document.cpp#L355

This bug can be triggered with these condition combined:
1. 'Ask download path for each file' is checked
2. and the file is an image
3. and the original file on the disk does not exist at its location any more

john-preston:
@mooons I can't reproduce that.

1. I've changed the settings.
2. I've taken an image 123.jpg and sent it to a chat.
3. I've removed the file from disk.
4. I've clicked on the sent file.

At click the downloaded file turned to a non-downloaded (with an arrow down on the thumbnail) and a "Save file" dialog box appeared. When I chose a filename it saved the file there. When I clicked on the file once again, it opened the file in the in-app viewer.

mooons:
@john-preston

Try taking a random screenshot with Win+R (so the image is saved to the clipboard), and paste it in a chat, uncheck compress images, Send. Then attempt to open the image.

[...]

PR #10280, opened 2021-01-31
Fix the bug where client won't preview uncompressed images when enabled 'Ask download path for each file' and the source file is removed

mooons:
Please check out https://github.com/telegramdesktop/tdesktop/issues/10237 for the problem description and repro:

Before the fix: image won't be saved to disk and preview won't pop up.

After the fix: image is prompted to be saved to disk. If saved, the preview window will show immediately after the image is 
downloaded.

Please kindly review. :)

### DIFF
--- a/Telegram/SourceFiles/data/data_document.cpp
+++ b/Telegram/SourceFiles/data/data_document.cpp
@@ -352,7 +352,7 @@ void DocumentOpenClickHandler::Open(
 		} else {
 			Core::App().showDocument(data, context);
 		}
-	} else if (data->saveFromDataSilent()) {
+	} else if (data->saveFromData()) {
 		openFile();
 	} else if (data->status == FileReady
 		|| data->status == FileDownloadFailed) {


ilya-fedin:
But your change is equally to replacing saveFromDataSilent with saveFromData
https://github.com/telegramdesktop/tdesktop/blob/3251b8bf6e740e55777e85d63065a61e43df54fe/Telegram/SourceFiles/data/data_document.cpp#L1188-L1196

I guess silent variant is used for a reason...

john-preston:
@mooons I was hoping to remove this setting some day, it really makes everything very messy. Are you really using this?

mooons:
> @mooons I was hoping to remove this setting some day, it really makes everything very messy. Are you really using this?

Yup I might be one of the few that are actually using this feature. The main reason I enable 'Ask download path for each file' is that by default it is downloading those image/videos to user downloads and they just stays there, not encrypted or anything, which is not great from a security/privacy point of view.

A cool idea would be just cache those images in memory when previewing them, not even saving them to a temp folder IMO.

mooons:
> But your change is equally to replacing saveFromDataSilent with saveFromData
> https://github.com/telegramdesktop/tdesktop/blob/3251b8bf6e740e55777e85d63065a61e43df54fe/Telegram/SourceFiles/data/data_document.cpp#L1188-L1196

Indeed.

> I guess silent variant is used for a reason...

Yes I wonder the same. @john-preston Any insights into this?

john-preston:
@mooons There is a problem with caching files in memory / encrypted storage, because the files are meant to be opened by external applications ("Open With" menu, default launching of them etc), and external applications are awaiting a non-encrypted real file.

mooons:
> @mooons There is a problem with caching files in memory / encrypted storage, because the files are meant to be opened by external applications ("Open With" menu, default launching of them etc), and external applications are awaiting a non-encrypted real file.

Got it. I understand.

What about the built-in preview feature for **un**compressed images? Looks like we are already either saving the "compressed" image to `tdata_user_data` or memory (haven't checked, I guess the former) before previewing them with the internal preview panel? Since compressed images doesn't seem to be affected by this config checkbox.

john-preston:
@mooons Yes, they're saved to encrypted cache.

We can't say for sure if we will show a file in the in-app viewer before we save it (we can know for sure for photos, but not for files), so we need first to download it and then to decide if we show the file in-app or open it in an external app.

ilya-fedin:
Why with two commits? :thinking:

mooons:
Thanks @john-preston for the answers.

So.. back to the point. Anything that prevents me from changing `` to `` for the fix?

Looks like `DocumentOpenClickHandler::Open` is mostly used for doc click handler when I quickly checked for references.

mooons:
@ilya-fedin My bad. I would do a squash and force-push in a moment.

mooons:
Fixed by https://github.com/telegramdesktop/tdesktop/commit/5feb381cb26fe4dda0e3438f4635e70ccf05fafe

Issue #8780, opened 2020-10-07
Gboard stickers shows as .webp files


spectraldani:
### Steps to reproduce
1. Open Gboard on an Android smartphone
2. Send a Gboard sticker to a Telegram chat
3. Notice that it displays as a Telegram sticker in Telegram for Mac and Android but not in Telegram Desktop

### Expected behaviour
Gboard sticker appears as a Telegram sticker in chat, just like in Telegram for Mac and phone apps.
![image](https://user-images.githubusercontent.com/6686762/95383855-27d6d380-08c2-11eb-9aab-3582f57ccaf8.png)

### Actual behaviour
The message appears as a .webp file with filename that matches the following regular expression: `-[0-9]+_-[0-9]+\.webp`
![image](https://user-images.githubusercontent.com/6686762/95383770-04ac2400-08c2-11eb-91ae-5e1b09da6efd.png)

### Configuration
**Operating system:** Windows 10

**Version of Telegram Desktop:** 2.2.4

**Used theme**: Night

Aokromes:
https://github.com/telegramdesktop/tdesktop/issues/8284

PR #10302, opened 2021-02-03
Add .webp to ImageExtensions list

mooons:
This would allow .webp images to be sent as an **image** rather than a file without any trickery.

I found that .webp images can be natually sent with the client. In fact if one renames the extension of an .webp image to .png (yes this becomes an extension mismatch), it can still be sent and received correctly. This should indicate that the .webp decoding capability comes for free (not the case for .avif images though, unfortunately). Tested under Windows 10 20H2.

e.g. https://www.gstatic.com/webp/gallery/1.sm.webp

Not sure if I should add this to `ExtensionsForCompression` list as well. Please advise.

Thanks.

### DIFF
--- a/Telegram/SourceFiles/ui/chat/attach/attach_extensions.cpp
+++ b/Telegram/SourceFiles/ui/chat/attach/attach_extensions.cpp
@@ -16,6 +16,7 @@ const QStringList &ImageExtensions() {
 		u".jpeg"_q,
 		u".png"_q,
 		u".gif"_q,
+		u".webp"_q,
 	};
 	return result;
 }


soredake:
@john-preston

john-preston:
I tested this and it doesn't change anything.

mooons:
@soredake @john-preston Thanks for checking this out.

Using https://www.gstatic.com/webp/gallery/1.webp ([source](https://developers.google.com/speed/webp/gallery1)) as a test file:

1. With the latest release 2.6.1, drag the .webp file to a chat. The only option is "send as document". But when actually sent, it is displayed as a **small** preview.



2. With the change, drag the same file to a chat. Both "send in a quick way" (large preview) and "send without compression" (as a file) option show up, just what we would expect, like a jpeg/png.



So the point is to really make drag&sending a .webp file to a chat behave the **same** as jpeg/png (and others listed in `ImageExtensions`).

mooons:
Update: It looks like the .webp file upload behavior is changed when I rebase my change on the latest `dev` branch.

It works fine ( "send in a quick way" gives a **large** preview) and "send without compression" gives a file) on my `dev-webp` branch. On latest `dev` (1fc24398a03818a8fa228e9c4dba0966b30055bd) both options still shows up, but either results in a **small** preview. Some other changes have broken the expected behavior it seems.

Nirzak (approved):

john-preston:
@mooons But it is intended to show .webp as stickers (if it is sent and received as a file). In case of "Compressed" checkbox on it should recompress it as opaque jpeg and send that way. Perhaps some of the changes broke that recompression if it happened before. But it didn't work at the time I had time to test that :(

If it is rebased in a working way I can merge that, but not by reverting the commit that changes the way .webp files (sent as files, not as jpeg photos) are treated by the app.

mooons:
I have rebased the patch. Also changed the `kLargetsStickerSide` from 2560 to 512 (Or are we actually going to support huge 2560x2560 stickers? I thought we only support [512x512 webp stickers](https://github.com/telegramdesktop/tdesktop/blob/0ee63950220649cace5c68f8be59ac0665d652f5/Telegram/SourceFiles/data/data_types.h#L238)?)

Tested that webp images with more than 512x512 pixels can now be shown as a file, or displayed as an image, just like jpg, png and others would behave.

Using https://www.gstatic.com/webp/gallery/4.webp for testing:

The one at the top is sent with "Drop photos here to send them in a quick way".
The one at the bottom is sent with "Drop images here to send them without compression".

test

Pls take a look. Thx.

john-preston:
@mooons No, it is intended to show even larger .webp files as stickers, not as files. A lot of bots use that feature to send sticker-esque images with sides larger than 512 and both official mobile apps show them as stickers.

Additionally TDesktop allows clicking on those ".webp files shown as stickers with at least one side larger than 512 px" and on click they're opened in the media viewer, allowing to see the full, not scaled down, image with transparency and everything.

So please don't change the display way of the uncompressed .webp files. Thanks!

Tonkittipongg (approved):

john-preston:
@mooons Current patchset (that adds only one string to the attach_extensions.cpp) doesn't change anything for me in regards of sending .webp

Is it supposed to work?

mooons:
> @mooons Current patchset (that adds only one string to the attach_extensions.cpp) doesn't change anything for me in regards of sending .webp
> 
> Is it supposed to work?

Yes it is working, but only on large .webp images (over 2560x2560 pixels) because we are not intending to lower the `kLargetsStickerSide` value:




Image used for testing in this case: https://www.learningcontainer.com/wp-content/uploads/2020/08/Webp-Sample-Image-file-Download.webp

john-preston:
@mooons But I asked only not to change the way uncompressed .webp files are shown.

I thought that a "Compress" checkbox will still appear for them (for all .webp files being sent) and it will work (it will compress them to JPG photos, if checked, and leave as uncompressed files, displayed as stickers, if it isn't checked).

Wasn't that the idea?

homeworkprod:
> @mooons No, it is intended to show even larger .webp files as stickers, not as files.

AFAIU, this whole "WebP implies displaying as a sticker" thing seems to have serious issues for users intending to use WebP images as actual images.

> A lot of bots use that feature to send sticker-esque images with sides larger than 512 and both official mobile apps show them as stickers.

Since this behavior *seems* to be only relevant for bots, how about introducing an option to explicitly show it as a sticker that bots can use? Is that metadata that can be attached to the image object so it is available at rendering time?

ilya-fedin:
> Since this behavior _seems_ to be only relevant for bots

it's relevant not only for bots, #8780

> how about introducing an option to explicitly show it as a sticker that bots can use?

You mean changes to bot API? Well, no one who can do such a change reads github, so discussing API changes is pointless here.

homeworkprod:
OK, so Telegram (Desktop) is basically stuck in a corner where usage of WebP as normal images is not realistic for the foreseeable future?

ilya-fedin:
¯\_(ツ)_/¯

ilya-fedin:
@mooons there were no updates for a long time, could the PR be closed?

john-preston:
Should work correctly now.

PR #10311, opened 2021-02-04
Add XDG Desktop Portal based file dialog implementation from Qt

ilya-fedin:
This allows to use portal dialogs more flexibly (e.g. fallback mechanism)
This also allows to have any changes we want for portal dialogs without patchig Qt

No more need to override QT_QPA_PLATFORM to use portal dialogs


john-preston:
Thanks!

PR #10377, opened 2021-02-14
Fix "view thread" not showing up on replies

9ary:
`repliesCount` is always 0 on messages that are not the top of the thread (that field is never set on them), but the presence of a reply ID should be a reliable indicator.

Side note: it seems like the Android client has a similar issue when no channel is attached to the group, I'm not equipped to attempt a fix myself and I have no idea how to report the bug otherwise, any pointers in that regard?

### DIFF
--- a/Telegram/SourceFiles/history/history_inner_widget.cpp
+++ b/Telegram/SourceFiles/history/history_inner_widget.cpp
@@ -1540,9 +1540,9 @@ void HistoryInner::showContextMenu(QContextMenuEvent *e, bool showFromTouch) {
 			});
 		}
 		const auto repliesCount = item->repliesCount();
-		const auto withReplies = IsServerMsgId(item->id)
-			&& (repliesCount > 0);
-		if (withReplies && item->history()->peer->isMegagroup()) {
+		const auto isThread = IsServerMsgId(item->id)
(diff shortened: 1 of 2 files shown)


john-preston:
Right now it works the same way as on mobile versions :(

9ary:
Is there a particular reason why it is this way at the moment? This behavior feels like a bug to me, it doesn't even work anymore for groups where a channel is linked. My patch makes it work for all supergroups.

john-preston:
@9ary There is a problem with threads that start in a deleted message.

9ary:
Oh that makes sense, I didn't think of that.

Louis6or (approved):

Nirzak (approved):

john-preston:
Unfortunately right now there is no way to know by a message if "view thread" will work or if it starts in a deleted message and won't work. So for now the behaviour will stay the same as mobile apps.

9ary:
Wouldn't attempting to fetch the top of the thread and caching the result work?

john-preston:
@9ary right now all context menu showing is synchronous. If it stays that way you won't be able to show correct button the first time you open the menu. Or you'll be forced to wait for the menu while the network request is completed (which could be really long if you've lost the connection).

PR #10443, opened 2021-02-26
Add caps active icon

snxx-lppxx:
Added caps lock activity icon


john-preston:
This looks like a mistake PR, just adding one empty file.

PR #10446, opened 2021-02-27
Expand moderating mode

MrSnowball-dev:
Add additional bot commands binds for moderators.

Clicking on them with mouse is tedious, but they still work just as regular commands.

I edited Key_9 to Key_2 because there is just 2 buttons in the bot and I suggest none will be added in the near future. Also binding existing commands to Key_3 and Key_4 helps to stay one-handed either on keyboard itself or on a numpad.


john-preston:
👍

PR #10457, opened 2021-03-03
Move from lock-threads app to lock-threads action

ilya-fedin:
lock-threads app seems to be deprecated


mirh:
Not that I'm saying this is *necessarily* stupid, but did you **really** need it?
On top of everything, I got like ***40*** notifications in less than a week.

ilya-fedin:
> On top of everything, I got like _**40**_ notifications in less than a week.

Probably you want to unsubscribe from issues you're not really interested in

> Not that I'm saying this is _necessarily_ stupid, but did you **really** need it?

People think that if an issue is closed, it's still looked by someone (it's not). This should force these people to open new issues instead of commenting on old closed issues.

mirh:
I get that noobs are gonna noob, but there are also legit reasons to still argue about older issues, if even. 
The moron complaining about osx not switching audio in the thread for the problem on linux is bad, but there were also people that were genuinely missing they had to bump/link/whatever tg against a newer version of openal for example. 

> Probably you want to unsubscribe from issues you're not really interested in

I *was* interested to them.. it's just that they had been closed over the span of 7 years. 
And to be sure \*nobody\* had been bumping them until now.

Aokromes:
> Not that I'm saying this is _necessarily_ stupid, but did you **really** need it?
> On top of everything, I got like _**40**_ notifications in less than a week.

xd if you think 40 are a lot.... i get 400... / day xd

PR #10480, opened 2021-03-05
Restore mouse button release hack in title widget event filter

ilya-fedin:


Telegram/SourceFiles/window/window_title_qt.cpp
```diff
@@ -255,6 +255,10 @@ bool TitleWidgetQt::eventFilter(QObject *obj, QEvent *e) {
 
 			if (e->type() == QEvent::MouseMove
 				&& mouseEvent->buttons() == Qt::NoButton) {
+				if (_mousePressed) {
```
mymedia2, line 258: Redundant `if`. You could simply assign the desired value to the boolean variable without a pre-check.

ilya-fedin: it is already merged

mymedia2: That doesn't matter. This `if` remains pointless.

ilya-fedin: image

PR #10492, opened 2021-03-06
Add portal-based open with dialog implementation

ilya-fedin:


Telegram/SourceFiles/platform/linux/linux_xdp_open_with_dialog.cpp
```diff
@@ -0,0 +1,180 @@
...
+			return false;
+		}
+
+		const auto fdGuard = gsl::finally([&] { ::close(fd); });
+		auto outFdList = Glib::RefPtr<Gio::UnixFDList>();
+
+		const auto parentWindowId = [&]() -> Glib::ustring {
+			std::stringstream result;
+			if (const auto activeWindow = Core::App().activeWindow()) {
+				if (IsX11()) {
```
mymedia2, line 84: In C++17 you can even join this `if` condition and the one above.
```suggestion
			if (const auto activeWindow = Core::App().activeWindow(); IsX11()) {
```

Telegram/SourceFiles/platform/linux/linux_xdp_open_with_dialog.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+		if (signalId != 0) {
+			QGuiApplicationPrivate::showModalWindow(this);
+			loop.exec();
+			QGuiApplicationPrivate::hideModalWindow(this);
+		}
+
+		return true;
+	} catch (...) {
+	}
```
mymedia2, line 167: Is it worth adding debug logging here?

ilya-fedin: I usually add logging only in cases when "Disconnected"/"ServiceUnknown" errors couldn't be thrown or if the method doesn't have any visual impact (e.g. LastUserInputTime)

TawfikYasser (approved):

Issue #8037, opened 2020-06-10
Channel and Group announcements use incorrect quotation marks in English localization


brendanfattig:
### Steps to reproduce
1. Make a channel or group chat
2. Send a message
3. Pin that message in the chat

### Expected behaviour
Using the English localization, the announcement that a message was pinned should read:
> [name] pinned "[message]"

### Actual behaviour
The announcement reads:
> [name] pinned <<[message]>>

### Configuration
**Operating system:**
Fedora Linux 32

**Version of Telegram Desktop:**
2.1.11

**Installation source (Linux Only)** - the official website / GitHub releases / flatpak / snap / distribution package:
flatpak

**Used theme**:
Day

<details><summary><b>Logs</b>:</summary>
Insert log.txt here (if necessary)
</details>

PR #10509, opened 2021-03-09
American quotes in UI strings

mymedia2:
I have converted all quotation marks in the lang.strings file to double (some to single) quotes `“”` for consistency within the whole file.

Fixes: #8037


john-preston:
I can merge this, but it won't change what people see in the app. The lang.strings is just a fallback for cloud translations loaded from api.

mymedia2:
Okay, I see. There is special [langpack.getStrings](https://core.telegram.org/method/langpack.getStrings) method in core API.

john-preston:
@mymedia2 The strings are managed on the translations platform at https://translations.telegram.org

PR #10522, opened 2021-03-13
Apply autofixes from clazy

ilya-fedin:


23rd:
I think it would be better to remove updating of `lib_ui` submodule from here.

ilya-fedin:
> I think it would be better to remove updating of `lib_ui` submodule from here.

I planned to push without it initially, but preston said to push with it.

23rd:
¯\\\_(ツ)_/¯

PR #10600, opened 2021-03-23
CAPS active 

snxx-lppxx:
Help wanted. Idk QT5.


john-preston:
It would be nice to show the caps lock symbol, but you need to use some platform APIs to detect if it is now pressed (you can't just watch for the press events, because it could be pressed not in tdesktop or even before tdesktop was launched).

This looks more like a feature request, not a pull request.

snxx-lppxx:
Okay, sorry.

PR #13827, opened 2021-03-24
Add a way to get dark mode state on KDE

### LINKED ISSUE #10595: Night mode doesn't sync with linux
### Steps to reproduce
1. Select a light color scheme.
2. Start telegram.
3. Activate the option to sync night mode with system colors.
4. Change the system colors to a dark color scheme.

### Expected behaviour
The theme in telegram should now use dark colors.

### Actual behaviour
Nothing happens. This also happens the other way around (starting with dark color scheme, then changing to a light scheme). The theme only changes when telegram is restarted.

### Configuration
**Operating system:** Linux (Manjaro with KDE Plasma)

**Version of Telegram Desktop:** 2.6.1

**Installation source (Linux Only)** - the official website / GitHub releases / flatpak / snap / distribution package: Pacman

**Used theme**: Default with custom colors

ilya-fedin:
Depends on https://github.com/desktop-app/lib_base/pull/56
Fixes #10595

### DIFF
--- a/Telegram/SourceFiles/platform/linux/linux_mpris_support.cpp
+++ b/Telegram/SourceFiles/platform/linux/linux_mpris_support.cpp
@@ -367,7 +367,7 @@ void MPRISSupport::Private::updateTrackState(
 		const auto &item1,
 		const auto &item2) {
 		return item1.first == item2.first
-			&& item1.second.equal(item2.second);	
+			&& item1.second.equal(item2.second);
 	})) {
 		metadata = currentMetadata;
 		PlayerPropertyChanged(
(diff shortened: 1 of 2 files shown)


Pointedstick:
@gikari perhaps using KConfigWatcher would be a better way to watch for changes?

gikari:
> @gikari perhaps using KConfigWatcher would be a better way to watch for changes?

Yes, but if I understand correctly, they cannot depend on KDE frameworks, so they cannot use it.

ilya-fedin:
There's no problem in using small libraries like kconfig, when looking for its dependencies I found though that it requires Qt5Xml what is not really good... Not to say that only using kconfig won't work in confined environments (flatpak/snap) and portals are preinstalled in most distros, so there are no real intent to support both portals and kconfig.

PR #16095, opened 2021-04-05
Watch for network aviability changes

ilya-fedin:
Allows to not to click "Try now" after connecting to Wi-Fi network


john-preston:
Thanks!

Nirzak (commented):
Thanks for this, man!It will save time ;)

PR #16105, opened 2021-04-07
Remember that media view was switched to PIP and open it as PIP again.

sashacmc:
Very useful, especially on large monitors.
Since the permanent opening of the video in full screen interferes and hide other applications.
This functionality allows remembering (in the current session) that the window has been switched to a PIP mode and will open a new video directly in the PIP mode.


john-preston:
Thanks!

Neurotoxin001:
Please make option to disable it, it's annoying.

sashacmc:
@Neurotoxin001
Thanks for your feedback, I can't add the option because of telegram policies, but I agree that volume bar on the PIP window is needed.

sashacmc:
Dear @Neurotoxin001,
I added a pull request with volume bar on the pip window.
https://github.com/telegramdesktop/tdesktop/pull/16274
Feel free to vote it.

PR #16117, opened 2021-04-11
Remove dangling timer in MTP Session

loicmolinari:
Commit bdc7f4114faa472407e5afbc27eb826e4cd38bf2 got rid of `_timeouter`'s callback without removing the timer that still fires every second.


john-preston:
Thanks!

Issue #8825, opened 2020-10-14
[Feature Request] Deep Link for Opening Profile, Like Android


PhardAlef:
Like Android and iOS
tg://openmessage?user_id=142364654
t.me/@id142364654
opens the profile of person, having group in chat/common
(or going directly to a private message, by including message id too)

MasterGroosha:
t.me/@id1234567 links doesn't work on Android too

M299792458:
It's now available in TD x64:
https://t.me/tg_x64/166
Make a pr.

ArmeF97:
Following @M299792458 advice, I made a PR based on https://github.com/TDesktop-x64, cherrypicking the commits 
#16124

MKRhere:
@john-preston from the PR

> A link to a specific message is even a more strange feature. You can't know any valid message id in someone else inbox, and you have no way of getting a message id or user id in any of the official apps

There is one usecase here; bots can know message IDs from their private chats and the user's chat ID. Bots can create links to messages in their private chat with the user.

> it will or won't work depending on did you or did you not see this user (in a group or in chats list) in this exact session.

If that's the limitation, we just have to live with that limitation then.

john-preston:
@MKRhere well, sent inline result's message ids are not supposed to be relied on as real inbox message ids and nobody promises anywhere in docs that those ids are going to match, so this is not a supported use case anyway. They are supposed to be used only from the bots side in his internal process.

MKRhere:
@john-preston I'm not talking about inline query. I'm talking about the bot's private chat.

john-preston:
@MKRhere Hmm. How do you get message id in users inbox in bot's private chats on bot's side?

MKRhere:
For messages from the user, we receive the message update, and we're able to access: `update.chat.id` and `update.message.message_id`.

For messages sent by the bot, as a response to `sendMessage`, similar to:

```TS
const message = telegram.sendMessage({ chatId, text });
message.message_id;
```

john-preston:
@MKRhere This is message_id in the bots messages box, not the message_id in users messages box. As with chats between users the message_ids for one private chat party is unrelated to message_ids for the second party.

john-preston:
@MKRhere Each user has his messages box ids start with 1 (I think, or smth like that), so if you take several different users, each of whom just joined Telegram, opened chat with the bot and sent him a "/start" message, for every user this message will have ID = 1, while for the bot the message_ids of those messages would be different.

[...]

PR #16124, opened 2021-04-12
Deep Link for Opening Profile, Like Android

ArmeF97:
This PR aims to follow @M299792458 advice to make a PR based on https://github.com/TDesktop-x64 fork, cherry-picking the commits.

See #8825

### DIFF
--- a/Telegram/SourceFiles/chat_helpers/message_field.cpp
+++ b/Telegram/SourceFiles/chat_helpers/message_field.cpp
@@ -173,7 +173,7 @@ void EditLinkBox::prepare() {
 
 	const auto submit = [=] {
 		const auto linkText = text->getLastText();
-		const auto linkUrl = qthelp::validate_url(url->getLastText());
+		auto linkUrl = qthelp::validate_url(url->getLastText());
 		if (linkText.isEmpty()) {
 			text->showError();
 			return;
@@ -182,6 +182,18 @@ void EditLinkBox::prepare() {
 			return;
 		}
 		const auto weak = Ui::MakeWeak(this);
+		if (linkUrl.contains("tg://user?id=")) {
+			const auto uid = linkUrl.split("tg://user?id=")[1].toInt();
+			if (uid > 0) {
+				const auto user = session->data().userLoaded(uid);
+				if (user != nullptr) {
+					linkUrl = "mention://user." + TextUtilities::MentionNameDataFromFields({uid, user->accessHash()});
+				} else {
+					url->showError();
+					return;
+				}
+			}
+		}
 		_callback(linkText, linkUrl);
 		if (weak) {
 			closeBox();
--- a/Telegram/SourceFiles/core/local_url_handlers.cpp
+++ b/Telegram/SourceFiles/core/local_url_handlers.cpp
@@ -413,8 +413,8 @@ bool HandleOpenMessage(
 		return false;
 	}
 	const auto params = url_parse_params(
-			match->captured(1),
-			qthelp::UrlParamNameTransform::ToLower);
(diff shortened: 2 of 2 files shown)


john-preston:
I don't really like the idea of "openmessage" handler. How can it be used? How can you know any valid id of a message in users inbox? Only user knows ids of his messages, this is not a channel, where everyone has the same messages.

john-preston:
Also, those `userId = stringValue("user_id").toInt()` + `static_cast<PeerId>(userId)` won't work anymore. You need something like `userId = UserId(stringValue("user_id").toULongLong())` + `PeerId(userId)`.

john-preston:
This "user" link is a strange thing as well. This link will work or won't work depending on the fact that user saw such user in current launch session or not. How can it be used reliably?

ArmelliniFederico:
I don't have answers for your question because I simply cherrypicked the commits from a working fork (telegramx64). 
Maybe those commits will help you to implement this feature in your way and close #8825

ArmeF97:
@john-preston 
I don't have answers for your question because I simply cherrypicked the commits from a working fork (telegramx64). 
Maybe those commits will help you to implement this feature in your way and close #8825

john-preston:
Well, I don't see a use case for such links, that sometimes work and sometimes do not. In the official client there is no way to get a user id, so why would you like to support links by user id that sometimes work and sometimes do not - I don't know.

There is no problem in implementing something like that. There is a problem with explanation of why this will bring any real value to the users.

ArmeF97:
@john-preston 
The reason why is explained in #8825 and it's simply the following: it's true that telegram desktop client won't show user id, but other clients can. On android links like tg://user?id=xxxx works so I see no reason why not make it work also on tdesktop. Official android client can do it so the main reason here is simply consistency between mobile and desktop

john-preston:
Yes, they work there, but there is no explanation why would you like them.

Also, they won't work well on tdesktop, because when you launch tdesktop you have no information about any users at all, so if you click on such link it will or won't work depending on did you or did you not see this user (in a group or in chats list) in this exact session.

This is unreliable so I was wondering why would anyone want such unreliable feature before implementing it. A link to a specific message is even a more strange feature. You can't know any valid message id in someone else inbox, and you have no way of getting a message id or user id in any of the official apps (if I remember correctly).

ArmeF97:
@john-preston 
Well, now that I understood what you are referring to, I think you are right. It would have been simply something "more" than something less, but if you see no value in it, it's your call. Close also #8825 then

bb010g:
@john-preston

> Also, they won't work well on tdesktop, because when you launch tdesktop you have no information about any users at all, so if you click on such link it will or won't work depending on did you or did you not see this user (in a group or in chats list) in this exact session.

tdesktop could query the UID if necessary when resolving a `tg:user?id=${UID}` URL? Linking directly off of UIDs is useful for when you want to refer to a specific account robustly over username changes, or when you want to store a link to a Telegram user that doesn't have a username outside of Telegram, for use in Telegram at a later date.

> You can't know any valid message id in someone else inbox, and you have no way of getting a message id or user id in any of the official apps (if I remember correctly).

Your own user ID at least can be trivially retrieved through the official Bot API. (I think Telegram clients should allow for opting-in to viewing IDs, like Discord clients when Developer Mode is enabled, but that's another matter.)

john-preston:
@bb010g No, tdesktop (or any other Telegram app) can't query users by ID only, Telegram API simply doesn't give you user by his ID, only by his ID and a 64 bit access hash, that is different for different users. You get this access hash when server gives you the user somewhere (in your chats list, or contacts list, or some group members list).

PR #16126, opened 2021-04-13
Add "Archive Sticker" into Sticker Box 3dots menu

### LINKED ISSUE #2357: [Feature request] Provide option to archive sticker sets

c0re100:
This PR allow user archive sticker pack to Archive Stickers via Telegram Desktop.

{8D9D0501-5B41-4C6A-9D4B-D32D134A3900}

Closes #2357

### DIFF
--- a/Telegram/SourceFiles/boxes/sticker_set_box.cpp
+++ b/Telegram/SourceFiles/boxes/sticker_set_box.cpp
@@ -198,6 +198,38 @@ void StickerSetBox::copyStickersLink() {
 	QGuiApplication::clipboard()->setText(url);
 }
 
+void StickerSetBox::archiveStickers() {
+	const auto weak = base::make_weak(_controller.get());
+	const auto setId = _set.c_inputStickerSetID().vid().v;
+	_controller->session().api().request(MTPmessages_InstallStickerSet(
+		_set,
+		MTP_boolTrue()
+	)).done([=](const MTPmessages_StickerSetInstallResult &result) {
+		const auto controller = weak.get();
+		if (!controller) {
+			return;
+		}
+		if (result.type() == mtpc_messages_stickerSetInstallResultSuccess) {
+			Ui::Toast::Show(tr::lng_stickers_has_been_archived(tr::now));
+            
+			const auto &session = controller->session();
+			auto &order = session.data().stickers().setsOrderRef();
+			const auto index = order.indexOf(setId);
+			if (index == -1) {
+				return;
+			}
+			order.removeAt(index);
+            
+			session.local().writeInstalledStickers();
+			session.local().writeArchivedStickers();
+            
+			session.data().stickers().notifyUpdated();
+		}
+	}).fail([](const MTP::Error &error) {
+		Ui::Toast::Show(Lang::Hard::ServerError());
+	}).send();
+}
+
 void StickerSetBox::updateTitleAndButtons() {
 	setTitle(_inner->title());
 	updateButtons();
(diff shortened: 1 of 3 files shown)


Telegram/SourceFiles/boxes/sticker_set_box.cpp
```diff
@@ -198,6 +198,17 @@ void StickerSetBox::copyStickersLink() {
 	QGuiApplication::clipboard()->setText(url);
 }
 
+void StickerSetBox::archiveStickers() {
+	_controller->session().api().request(MTPmessages_InstallStickerSet(
+		_set,
+		MTP_boolTrue()
+	)).done([=](const MTPmessages_StickerSetInstallResult &result) {
+		Ui::Toast::Show(tr::lng_stickers_has_been_archived(tr::now));
+	}).fail([=](const MTP::Error &error) {
```
23rd, line 207: I'm afraid it's not so simple.
You have to apply the current freshly archived sticker set locally and remove it from the list of installed sticker sets.
I think that would be enough for now.

```suggestion
	const auto weak = base::make_weak(_controller.get());
	const auto setId = _set.c_inputStickerSetID().vid().v;
	_controller->session().api().request(MTPmessages_InstallStickerSet(
		_set,
		MTP_boolTrue()
	)).done([=](const MTPmessages_StickerSetInstallResult &result) {
		const auto controller = weak.get();
		if (!controller) {
			return;
		}
		if (result.type() == mtpc_messages_stickerSetInstallResultSuccess) {
			Ui::Toast::Show(tr::lng_stickers_has_been_archived(tr::now));

			const auto &session = controller->session();
			auto &order = session.data().stickers().setsOrderRef();
			const auto index = order.indexOf(setId);
			if (index == -1) {
				return;
			}
			order.removeAt(index);

			session.local().writeInstalledStickers();
			session.local().writeArchivedStickers();

			session.data().stickers().notifyUpdated();
		}
	}).fail([](const MTP::Error &error) {
```

Telegram/SourceFiles/boxes/sticker_set_box.cpp
```diff
@@ -198,6 +198,38 @@ void StickerSetBox::copyStickersLink() {
...
+				return;
+			}
+			order.removeAt(index);
+            
+			session.local().writeInstalledStickers();
+			session.local().writeArchivedStickers();
+            
+			session.data().stickers().notifyUpdated();
+		}
+	}).fail([=](const MTP::Error &error) {
```
23rd, line 228:
```suggestion
	}).fail([](const MTP::Error &error) {
```

john-preston:
Is there a way to reproduce the failure to archive the sticker set? Because if happens only when there are some problems on the server side I wouldn't add this phrase to the lang strings. Usually we try to add for translation only strings you can see and check (as a translator) in the app.

c0re100:
> Is there a way to reproduce the failure to archive the sticker set? Because if happens only when there are some problems on the server side I wouldn't add this phrase to the lang strings. Usually we try to add for translation only strings you can see and check (as a translator) in the app.

I think there is no way to reproduce in normal case.
As I know, if sticker set was deleted, it will popup `Sticker pack not found` message, so that sticker set can't archive.

Only `MTP_inputStickerSetEmpty()` can reproducing a failure in my test.
It will return error code 400 `STICKERSET_INVALID`
But this is a invalid case...?

john-preston:
@c0re100 Then please remove lng_stickers_failed_to_archive (no reliable way to see this message) and replace the usage with just `Lang::Hard::ServerError()`.

Telegram/SourceFiles/boxes/sticker_set_box.cpp
VChernomYA: 81ac646ceb90e1aaffa30a9716dc6a5e

PR #16135, opened 2021-04-13
Added CLion-generated files to .gitignore

humanzer0:
CLion IDE generates cmake-build-debug and .idea files, which is not ignored by tdesktop .gitignore.


ilya-fedin:
If you need to ignore any files for your own needs, you can add them to `.git/info/exclude` or

john-preston:
Well, why not.

PR #16140, opened 2021-04-14
Use fine grained session timers

loicmolinari:
The check of sent requests and containers is done unconditionally every second even though the request timeout is 10 seconds and the container timeout is 600 seconds. This commit uses fine grained timers instead in order to avoid useless system wake-up events.

The check of sent requests is now scheduled on demand when a new request is queued. Then the callback, while parsing queued requests, computes the delta to the closest expiring request and automatically schedules the next check if necessary.

Given the high value of the container timeout, its callback is called repeatedly every 600 seconds, unless it computes a lower delta for an expiring container using the same logic as for the requests.

### DIFF
--- a/Telegram/SourceFiles/mtproto/session_private.cpp
+++ b/Telegram/SourceFiles/mtproto/session_private.cpp
@@ -39,7 +39,6 @@ constexpr auto kPingSendAfterForce = 45 * crl::time(1000);
 constexpr auto kTemporaryExpiresIn = TimeId(86400);
 constexpr auto kBindKeyAdditionalExpiresTimeout = TimeId(30);
 constexpr auto kTestModeDcIdShift = 10000;
-constexpr auto kCheckSentRequestsEach = 1 * crl::time(1000);
 constexpr auto kKeyOldEnoughForDestroy = 60 * crl::time(1000);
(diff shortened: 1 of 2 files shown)


Telegram/SourceFiles/mtproto/session_private.cpp
```diff
@@ -296,12 +302,20 @@ void SessionPrivate::clearOldContainers() {
 				resend(innerMsgId, -1, true);
 			}
 		} else {
+			if (i->second.sent - checkTime < nextTimeout) {
+				nextTimeout = i->second.sent - checkTime;
+			}
```
23rd, line 307:
```suggestion
			nextTimeout = std::min(i->second.sent - checkTime, nextTimeout);
```

loicmolinari: Fine. Then I guess it should be the same for the requests at line 270?
```
-			} else if (request->lastSentTime - checkTime < nextTimeout) {
-				nextTimeout = request->lastSentTime - checkTime;
-			}
+			else {
+				nextTimeout = std::min(request->lastSentTime - checkTime, nextTimeout);
+			}
```

23rd: I think so, yes.
But this is just my suggestion.

PR #16145, opened 2021-04-14
Fix voice record button infinite animation

loicmolinari:
Enabling animations triggers an animation of the voice record button that keeps firing uselessly at ~120 calls per second until the button is manually toggled. The animation callback cannot stop itself since it expects the button to be visible.

This commit fixes the issue by preventing spawning the animation if the widget is hidden.


23rd:
Yeah, I can confirm the problem, thanks a lot.

john-preston:
Thanks!

PR #16171, opened 2021-04-21
Update README.md to split VS2019 32 and 64 bits

LuisAlfredo92:
Since there are different methods to build for VS for 32 and 64 bits, I think it'd be good to split the label


ilya-fedin:
@john-preston i thought 64-bit instructions is not ready yet for public use, e.g. no one can switch build mode to win64 without DesktopPrivate folder

PR #16185, opened 2021-04-26
Expand moderating mode some more

MrSnowball-dev:
Add missing /undo, /next, /stats options
Add 5 and 6 keys as translating option for NumPad

Also, revert previous key check for bot keyboard, as it now does not recognize 3-9 keys as inputtable numbers for some reason.


john-preston:
@MrSnowball-dev Did you test that? It looks strange:

image

If you read this code, you see, that Qt::Key_3,4,5,6 won't ever get into the second part "} else if (user) {", because they'll always get into the first "if (key >= Qt::Key_1 && key <= Qt::Key_9) {".

MrSnowball-dev:
Oh yeah, that makes sense why they didn't work previously. I'll revert that.

PR #16199, opened 2021-04-27
Fixed misuse of comma

DuckMasterAl:
Instead of the comma, I added a semicolon, but a period can also be used.
This is necessary because the sentence has two independent clauses. This means it can either be compound sentence or two separate sentences.


SpaceXLiquid (approved):

john-preston:
cmon

PR #16273, opened 2021-05-03
Fix video player volume level changing

sashacmc:
Bug:
Sometimes changing of the volume level or mute/unmute has no effect.
It happens because Fader::onTimer get a current volume level from the Mixer,
but it gets an event about the volume modification from the settings.
Bug appear when the method onTimer calling between updating
of the settings and the Mixer volume.

Solution:
Updating the Mixer volume before the settings.
(maybe will be better to get the volume level
from the settings in place of the Mixer,
but I am not sure about other side effects of this)


john-preston:
Thanks!

PR #16303, opened 2021-05-10
Enhancement: Hide photo viewer overlay after zooming an image

### LINKED ISSUE #7931: Enhancement: Hide photo viewer overlay after zooming an image

Often photos have a long textual description, or sent as album. In this case, after clicking on it, the internal image viewer will show both description (shortened) and two of other photos from this album.

This interactive information will stay on screen until some timeout (about 3 sec) and then slowly fade. But only if no user mouse input was detected! So, for example, if you want to pan zoomed-in image, you will trigger displaying of this interface, which might overlap some important parts of the picture.

Waiting for it to disappear every time is not convenient.

Disable _any_ onscreen controls in image viewer if this picture was zoomed-in. Especially right after pressing the middle mouse button.

When a photo zoom is restored to default (second middle click) or just made smaller than screen-fit size (Ctrl + Wheel Down) the behavior of on-screen controls should remain as it is currently (instantly visible with fade timeout).

There is another possible way to hide all interactive controls: by implementing a dedicated button that will toggle them.

For example, it could be a square labeled "i" ('information' icon), located in top-left corner of this overlay, just like 'X' button is located at tor-right corner currently.

By default after restart of Telegram Desktop client, the state of overlay should be globally visible. This 'i'-button will be visible too (and will fade-out just like other buttons).

When clicked, all of other buttons, links and texts on overlay should disappear, but only this button ought to stay on screen (semi-transparent like 'X'-close) and continue to conform the 3-sec timeout (so eventually it will fade-out, but mouse move will restore it).

Second click on 'i'-button will toggle display of overlay back. The state of overlay visibility should be kept global (for all images and chars) until Telegram Desktop restarts.

When main buttons on overlay are hidden, the image viewer should still accept any keyboard shortcuts to navigate: for example, Left/Right arrow to open next photo, Ctrl+S to save, and Escape to exit picture view mode.

Another possibility would be a context menu entry for image overlay, titled "Hide controls" initially, but "Show controls" when this function was already used to hide the overlay.

### Configuration

**Operating system:** Windows 7 SP1
**Version of Telegram Desktop:** 2.1.6
**Used theme**: Day

### Additional context

Here are some screenshots of the issue: (with 125% interface scale)

**photo with description:**

Initial:
description - initial

Default:
description - default
_(Note that semi-transparent box with description text and links could still severely interfere with image content)_

Timeout:
description - timeout
_(And you should hold your mouse very still, to be able to read it like this)_

**photo albums:**

Initial:
album - initial

Default:
album - default
_(Note that almost opaque album previews also interfere with content)_

Zoomed to the bottom:
album - zoomed
_(The main issue: even after zooming in picture manually, this unnecessary overlay still visible)_

So it would be very logically to hide overlays on zoom, since if user had zoomed – then he want to see a part of image, and not the text/album previews, which should become visible again after zooming out.

aleksusklim:
Fixes #7931.

In OverlayWidget, I've added new ControlsState called ControlsDisabled, which locks visibility of all overlay controls and prevents mouse interaction for anything other than previous/next navigation and close button.

It is enabled when user zooms in the image or presses reset-zoom button (middle mouse). Then all overlay controls are instantly disappearing, while mouse cursor will be permanently visible (to allow the user to click on navigation buttons, though invisible).

After zooming out the image or after switching to another photo, the overlay controls will appear gradually.

To make this feature more accessible, I've also added extra hotkeys to change zoom level: NumPad Plus and NumPad Minus to zoom in and out (which currently required to hold Ctrl key) and NumPad Multiply to reset zoom level, just as middle mouse button (currently this is only possible by Ctrl+Zero, which is not working reliably for some reason).

Also for simple photos (not videos and not documents) the Enter and Space keys will reset zoom level, effectively enlarging and hiding controls. This is convenient, because after switching to another photo (left/right keys) the controls will always appear again.

### DIFF
--- a/Telegram/SourceFiles/media/view/media_view_overlay_widget.cpp
+++ b/Telegram/SourceFiles/media/view/media_view_overlay_widget.cpp
@@ -964,7 +964,7 @@ void OverlayWidget::waitingAnimationCallback() {
 }
 
 void OverlayWidget::updateCursor() {
-	setCursor(_controlsState == ControlsHidden
+	setCursor(_controlsState == ControlsHidden && !isZoomedIn()
 		? Qt::BlankCursor
 		: (_over == OverNone ? style::cur_default : style::cur_pointer));
 }
@@ -3291,7 +3291,7 @@ void OverlayWidget::paintEvent(QPaintEvent *e) {
 		}
 
 		// more area
-		if (_moreNavIcon.intersects(r)) {
+		if (_moreNavIcon.intersects(r) && !isZoomedIn()) {
 			auto o = overLevel(OverMore);
 			p.setOpacity((o * st::mediaviewIconOverOpacity + (1 - o) * st::mediaviewIconOpacity) * co);
 			st::mediaviewMore.paintInCenter(p, _moreNavIcon);
@@ -4131,7 +4131,7 @@ void OverlayWidget::updateOver(QPoint pos) {
 		updateOverState(OverRotate);
 	} else if (_document && documentBubbleShown() && _docIconRect.contains(pos)) {
 		updateOverState(OverIcon);
-	} else if (_moreNav.contains(pos)) {
+	} else if (_moreNav.contains(pos) && !isZoomedIn()) {
 		updateOverState(OverMore);
 	} else if (_closeNav.contains(pos)) {
 		updateOverState(OverClose);
--- a/Telegram/SourceFiles/media/view/media_view_overlay_widget.cpp
+++ b/Telegram/SourceFiles/media/view/media_view_overlay_widget.cpp
@@ -1035,6 +1035,12 @@ void OverlayWidget::resizeContentByScreenSize() {
 	}
 	_x = (width() - _w) / 2;
 	_y = (height() - _h) / 2;
+	_initialZoom = _zoom;
+	if (_controlsState == ControlsDisabled) {
+		_controlsState = ControlsHidden;
+		activateControls();
+		updateCursor();
+	}
 }
 
 float64 OverlayWidget::radialProgress() const {
@@ -1155,7 +1161,7 @@ void OverlayWidget::zoomReset() {
 	auto newZoom = _zoom;
 	const auto full = _fullScreenVideo ? _zoomToScreen : _zoomToDefault;
 	if (_zoom == 0) {
-		if (qFloor(full) == qCeil(full) && qRound(full) >= -kMaxZoomLevel && qRound(full) <= kMaxZoomLevel) {
+		if (qFloor(full) == qCeil(full) && qRound(full) >= -kMaxZoomLevel && qRound(full) <= kMaxZoomLevel && qRound(full) != 0) {
 			newZoom = qRound(full);
 		} else {
 			newZoom = kZoomToScreenLevel;
@@ -1271,6 +1277,9 @@ void OverlayWidget::close() {
 }
 
 void OverlayWidget::activateControls() {
+	if (_controlsState == ControlsDisabled) {
+		return;
+	}
 	if (!_menu && !_mousePressed) {
 		_controlsHideTimer.callOnce(st::mediaviewWaitHide);
 	}
@@ -1304,7 +1313,11 @@ void OverlayWidget::onHideControls(bool force) {
 	if (_fullScreenVideo) {
 		_streamed->controls.hideAnimated();
 	}
-	if (_controlsState == ControlsHiding || _controlsState == ControlsHidden) return;
+	if (_controlsState == ControlsHiding
+		|| _controlsState == ControlsHidden
+		|| _controlsState == ControlsDisabled) {
+		return;
+	}
 
 	_lastMouseMovePos = mapFromGlobal(QCursor::pos());
 	_controlsState = ControlsHiding;
@@ -3681,6 +3694,22 @@ void OverlayWidget::setZoomLevel(int newZoom, bool force) {
 	if (!force && _zoom == newZoom) {
 		return;
 	}
+	if (!_fullScreenVideo) {
+		if ((_initialZoom == 0 && newZoom > 0)
+			|| (_initialZoom != 0 && newZoom >= 0 && newZoom != _initialZoom)) {
+			if (_controlsState != ControlsDisabled) {
+				activateControls();
+				onHideControls(true);
+				updateControlsAnimation(_controlsAnimStarted + st::mediaviewHideDuration);
+				_controlsState = ControlsDisabled;
+				updateCursor();
+			}
+		} else if (_controlsState == ControlsDisabled) {
+			_controlsState = ControlsHidden;
+			activateControls();
+			updateCursor();
+		}
+	}
 
 	const auto full = _fullScreenVideo ? _zoomToScreen : _zoomToDefault;
 	float64 nx, ny, z = (_zoom == kZoomToScreenLevel) ? full : _zoom;
@@ -4034,6 +4063,15 @@ void OverlayWidget::updateOverRect(OverState state) {
 
 bool OverlayWidget::updateOverState(OverState newState) {
 	bool result = true;
+	if (_controlsState == ControlsDisabled
+		&& newState != OverLeftNav
+		&& newState != OverRightNav
+		&& newState != OverClose
+		&& newState != OverVideo) {
+		_over = OverNone;
+		updateCursor();
+		return false;
+	}
 	if (_over != newState) {
 		if (newState == OverMore && !_ignoringDropdown) {
 			_dropdownShowTimer.callOnce(0);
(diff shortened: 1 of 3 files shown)


Telegram/SourceFiles/media/view/media_view_overlay_widget.h
```diff
@@ -498,6 +499,7 @@ private Q_SLOTS:
 		ControlsShown,
 		ControlsHiding,
 		ControlsHidden,
+		ControlsDisabled
```
23rd, line 502:
```suggestion
		ControlsDisabled,
```

aleksusklim:
@23rd, It is not confusing as the user is knowing what he is doing. The mouse cursor exactly shows that the "button" is under.

Often when a bright image is zoomed in – you can't clearly see those arrows anyway!

Hm-m, do you think that the button should appear when hovering over it? I'm not sure it is easy to make as is, but it's certainly possible in theory,

23rd:
> It is not confusing as the user is knowing what he is doing. The mouse cursor exactly shows that the "button" is under.

This is just my opinion. I think only the bottom info should be hidden, the navigation buttons and the cross button should keep their behavior unchanged.

> It is enabled when user zooms in the image or presses reset-zoom button (middle mouse).

Perhaps the bottom info should only be hidden when it overlaps content.

aleksusklim:
But that's the point of «fullscreen view» – to view something is a full screen, without obscuring it with interface elements. (Also note that this media viewer currently doesn't allow user to drag something out of screen borders: for example, I cannot move zoomed-in picture down-left to make its top-right corner NOT snapped to screen).

Even if the user becomes confused, a single click into the picture will be enough to close the viewer instantly!

Wait, I have an idea: scratch this approach and make it differently: let the «disappear timeout for controls» much shorter when the image was zoomed. I will test that to see how it fits…

aleksusklim:
@23rd, I've made a second attempt. Now I don't block the controls, just shorten their «hide time» if the image was zoomed in.

Also, I prevent controls from disappearing if one of them is under the mouse cursor.

Finally, I've removed all proposed keyboard binding, instead enabling Plus, Minus and Zero keys to zoom simple photos without held Ctrl key.

Telegram/SourceFiles/media/view/media_view_overlay_widget.cpp
```diff
@@ -3709,6 +3711,12 @@ void OverlayWidget::setZoomLevel(int newZoom, bool force) {
 		_x = qRound(nx / (-z + 1) + width() / 2.);
 		_y = qRound(ny / (-z + 1) + height() / 2.);
 	}
+	if(isZoomedIn()){
+		if(_controlsState == ControlsHiding){
```
23rd, line 3715:
```suggestion
	if (isZoomedIn()) {
		if (_controlsState == ControlsHiding) {
```

23rd:
I'm afraid this spoils the user experience even more.
<details>

https://user-images.githubusercontent.com/4051126/118115985-58e82b00-b3f2-11eb-8f22-ffdc9e3e8850.mp4
</details>

I see two options here.

- A hard way: leave navigation + cross buttons and hide the bottom information only when it overlaps the content.

 - An easy way: add a fullscreen mode (`Ctrl + F`) for photos in the same way as for video. In this mode, for advanced users, the whole interface would be hidden.
 
But remember, the final decision is always up to Preston. I'm just giving my opinion.

aleksusklim:
And what's the problem with the mouse cursor? Visible when moving, hidden right after that.

Welp, we can make it visible the whole time while isZoomedIn() returns true. Would that suffice?

One thing I don't like is that right-down menu will still pop-up, preventing the user to put the mouse there. Since this is the same menu that brings up after right click – maybe remove that button from the screen in zoomed state?

This way the user could put the cursor to right-down corner effectively hiding it, and not overlapping with any controls. The menu is still accessible with right click.

I'll try to add this behavior to my commits...

23rd:
Please consider your changes for all possible displayed content, including documents and themes.

aleksusklim:
I tested my code with albums, videos, gifs and files.

For albums: previews of other images in the album (just as the post caption text) are disappearing with all other controls, but they do not prevent interface hiding even when the mouse cursor is over them. I think it is acceptable, because they can fill a large part of the screen, so it is good that they are transparent for hover detection.

For videos, the main playback controls are not disappearing in zoomed state. Also the ResetZoom button sometimes does not enlarge the video to full screen. But the video has its own fullscreen mode, where new behavior has no effect (so the controls will hide after a default delay). Zooming video not in full screen still triggers quicker hiding of anything other than playback controls.

For GIFs, the behavior is identical to photos, except that new simplified key-bindings for zoom are requiring holding the Ctrl key. Other than that, a zoomed-in GIF can be dragged and viewed just as photo (hiding controls), but also can be paused by Space (current TDesktop behavior unchanged).

For files (documents that failed to display as media content, for example videos with unknown codec, some large WEBP and several PDF files) the zoom does not making sense, because it will do nothing: TDesktop will just show a preview (or generic icon) with SaveAs link. Anyway, after a forced zoom-in event, the other controls will hide quicker (because the proposed code does not handle this case separately), but I believe this brings no harm to user experience. 

Some invalid media may be displayed as 1x1 transparent pixel, which can be zoomed/maximized, hiding overlay controls. But again, this is the expected behavior.

Thus I think the proposed code is production-ready. The only tweak that could be done further – is the adjustment of new quicker delays. If you think that the delays are better be different – you can change them as you like. For now, I left them all equal to 150 ms.

23rd:
> For files (documents that failed to display as media content, for example videos with unknown codec, some large WEBP and several PDF files) the zoom does not making sense, because it will do nothing: TDesktop will just show a preview (or generic icon) with SaveAs link. Anyway, after a forced zoom-in event, the other controls will hide quicker (because the proposed code does not handle this case separately), but I believe this brings no harm to user experience.

If zooming in doesn't change anything for files, then it shouldn't speed up controls hiding, should it?

aleksusklim:
> If zooming in doesn't change anything for files, then it shouldn't speed up controls hiding, should it?

It does not making any difference. The user has no reason to zoom, and even if he'll do it – nothing breaks. What's the problem?

Displaying of invalid documents – is an exceptional situation. Like, there is a button to «Rotate» a picture. It does nothing for documents. I could say, «Hey, then this button should be hidden if it is not working!»

But it's such a minor detail that hurts noone.

23rd:
> Displaying of invalid documents – is an exceptional situation.

But displaying themes — is not.

aleksusklim:
How to test a "theme" on test cloud?

23rd:
> How to test a "theme" on test cloud?

You can upload a file of theme to Saved Messages.

aleksusklim:
OK, I've tested it. (I notice than pop-up menu in right-down corner is overlapped by «APPLY THIS THEME» but it is irrelevant here).

The theme preview does not zooming either. So, «zoom does nothing». Oh, wait! It shortens the delay of hiding overlay controls.

So, when the users become aware that middle mouse click effectively «goes fullscreen» and hides the overlay – they will use this behavior for themes preview too.

For example, now I saw the whole theme preview thing for the first time – and the overlay still obscured some of visible elements. Especially when the interface scale is increased like mine. Yet, the bottom panel (cancel/apply) at large scale is interfering way too severely! Example at 125% on 1024*768 screen, theme MossAmbrosia: 

<details> <summary> MossAmbrosia_1.png </summary> 


</details>

Though for default size it looks pretty neat:

<details> <summary> MossAmbrosia_2.png </summary> 


</details>

And I don't feel anything wrong with triggering quicker hiding of controls (at least some of them) after a middle-click:

<details> <summary> MossAmbrosia_3.png </summary> 


</details>

This is how the bug with an overlapped menu looks like, in different scales:

<details> <summary> MossAmbrosia_4.png </summary> 


</details>

I mean, if you want to have less small visual glitches, then it's good to get rid of them. But it should not delay the implementation of new independent features and enhancements.

lokpozzta:
m: Telegram Support <login+canned.response@stel.com>
Date: 20 June 2021 at 3:04:40 AM GMT+7
To: sreyleak2703@icloud.com
Subject: Re: Invalid phone number: +855 1122562

Aokromes:
> m: Telegram Support [login+canned.response@stel.com](mailto:login+canned.response@stel.com)
> Date: 20 June 2021 at 3:04:40 AM GMT+7
> To: [sreyleak2703@icloud.com](mailto:sreyleak2703@icloud.com)
> Subject: Re: Invalid phone number: +855 1122562

you leaked your phone number.

utkarsh914 (approved):

utkarsh914 (approved):

Sirflyzoner (approved):
- [ ] thank u

ilya-fedin:
Conflicts

Issue #18, opened 2014-06-11
[feature request] minimize to tray by pressing Esc key


yunusga:
Please add minimize to tray by pressing Esc key

splintor:
Esc key is excellent for this (see also #1367). If this is problematic, you can make it optional.

PR #16321, opened 2021-05-14
Update Dockerfile

psychosys:
Delta RPMs disabled because /usr/bin/applydeltarpm not installed

### DIFF
--- a/Telegram/build/docker/centos_env/Dockerfile
+++ b/Telegram/build/docker/centos_env/Dockerfile
@@ -12,6 +12,7 @@ ENV Qt5_DIR ${QT_PREFIX}
 
 RUN yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
 RUN yum -y install https://packages.endpoint.com/rhel/7/os/x86_64/endpoint-repo-1.7-1.x86_64.rpm
+RUN yum -y install deltarpm
 RUN yum -y install centos-release-scl
 
 RUN yum -y install git cmake3 meson ninja-build autoconf automake libtool \


ilya-fedin:
What's delta RPMs?

psychosys:
> 
> 
> What's delta RPMs?

Some of the The RPM packages that is going to be installed on Line #18 are DELTA RPMs ( Delta RPM packages contain the difference between an old and a new version of an RPM package ) so because Delta RPM doesn't exist on Docker Container you will see this error during the installation : "**Delta RPMs disabled because /usr/bin/applydeltarpm not installed**" ( After Execution of yum on Line #18 ) and it breaks the Installation on a **fresh** installed Centos 

So, Installing the Delta RPM on docker container **before** installation of the main packages prevents the error and installation continues smoothly to the END.

ilya-fedin:
> and it breaks the Installation on a **fresh** installed Centos

I don't know what do you mean by this, but Dockerfile works in the current state and I don't think that redudant package is needed

psychosys:
> 
> 
> > and it breaks the Installation on a **fresh** installed Centos
> 
> I don't know what do you mean by this, but Dockerfile works in the current state and I don't think that redudant package is needed

Deltrarpm is an essential package, it is already installed on your system but it was missing on Docker Container

ilya-fedin:
I don't think it's installed on my system given that I use NixOS

psychosys:
> 
> 
> I don't think it's installed on my system given that I use NixOS

I had this issue with **Centos 7** and adding **DeltaRPM** to the docker resolved the problem

ilya-fedin:
I don't believe in that since docker container shouldn't change its behavior depending on host system

PR #16324, opened 2021-05-15
Fix BIO_new_mem_buf leaks

ilya-fedin:


Telegram/SourceFiles/_other/packer.cpp
```diff
@@ -430,7 +442,15 @@ int main(int argc, char *argv[])
 	uint32 siglen = 0;
 
 	cout << "Signing..\n";
-	RSA *prKey = PEM_read_bio_RSAPrivateKey(BIO_new_mem_buf(const_cast<char*>((BetaChannel || AlphaVersion) ? PrivateBetaKey : PrivateKey), -1), 0, 0, 0);
+	RSA *prKey = [&] {
```
23rd, line 445: Is the capture by reference needed here?

ilya-fedin: Not sure

PR #16329, opened 2021-05-16
Use mallocng on Linux

### LINKED ISSUE #16084: The app consumes an unexpectedly large amount of RAM
### Steps to reproduce
1) Apparently, leaving the application running in the background for quite a long time

### Actual behavior
I don't know exactly how much the app should consume normally. I left my computer running overnight with the app just running in the background. In the morning I opened the task manager and saw that the app was currently consuming about 720 MB of RAM. Only Plasma was consuming more than that.
After restarting, the app was consuming about 150 megabytes from the start, after about an hour and as of this writing, the app is consuming 340 MB of memory, a figure that is slowly growing.
The behavior looks like a memory leak.

### Configuration
OS: Arch Linux, 5.11.11-arch1-1
TG ver.: 2.7.1
Source: installed from community repository of my distribution.
Theme: Color, with enabled Night Mode.

I'm sorry, I don't know if a log is needed here or how to get one.

ilya-fedin:
Fixes #16084

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -91,3 +91,6 @@
 [submodule "Telegram/lib_webview"]
 	path = Telegram/lib_webview
 	url = https://github.com/desktop-app/lib_webview.git
+[submodule "Telegram/ThirdParty/mallocng"]
+	path = Telegram/ThirdParty/mallocng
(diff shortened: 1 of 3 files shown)


john-preston:
Ok, let's try this one.

PR #16364, opened 2021-05-24
Actual link to Vazir Fonts license in README

anthonyaxenov:
The previous link led to the license text which was relevant up to [v26.0.2](https://github.com/rastikerdar/vazir-font/tree/v26.0.2), but removed in [v27.0.0](https://github.com/rastikerdar/vazir-font/blob/v27.0.0/LICENSE)


john-preston:
Thanks!

PR #16420, opened 2021-06-07
[app.cpp] Sorted header files

EndlessArch:


23rd:
I'm afraid in this case the list of headers should be sorted in all files.

john-preston:
Why? 🙈

PR #16561, opened 2021-07-04
Update App Icon on Desktop

gauthamkrishna9991:
# Updating App Icons to newer design.

I've added all the new icons, modified to match the present icons (in both **shadows** and **offsets**) and have updated it to match the ones here, and have replaced those icons with the newer icons, which are presently found in telegram app, whose assets were found [here](https://telegram.org/tour/screenshots).

[New Icon Work Files (`TelegramDesktop_ImageAssets.zip`, GIMP XCF Files)](https://github.com/telegramdesktop/tdesktop/files/6759862/TelegramDesktop_ImageAssets.zip)

## Old Icon (in repository):
telegram_icon_old

## Newer Icon (replacing)
telegram_icon_new

I've also modified/rendered icons in a way that also strips down the unnecessary icon metadata, which cut down on final icon sizes massively (**`304.28KiB`** -> **`137.02KiB`**, 55% size reduction).


23rd:
https://github.com/telegramdesktop/tdesktop/pull/10098#issuecomment-765383293.

gauthamkrishna9991:
There are some errors with CLA? It says I've signed the CLA, but I'm not able to see that here.

gauthamkrishna9991:
@23rd is there any hope for them to update the application icons in this case??

The newer icons are in the exact specifications as was from the previous version.

23rd:
> @23rd is there any hope for them to update the application icons in this case??

I'm afraid there is no chance.

john-preston:
I've asked once again and I was told to leave the current icon alone, at least for now. Only to update it for Big Sur guidelines (without changing the icon itself), only on macOS.

Eric-01:
If I may ask, what was the reasoning in short? Usually brand is kept consistent everywhere for a reason.

john-preston:
@Eric-01 Maybe to keep different icons with macOS native app.

gauthamkrishna9991:
It seems as if both Android and iOS both have the newer icons too. That also is the case for the MacOS Version also.

The only platforms that does seem to hold out on this are Windows and Linux (correct me if I'm wrong), and the icons have been worked out to the exact specs as the previous icons. In any case, implementing this would only make this app look more consistent with the other platforms.

I'd also want to know why.

john-preston:
@gauthamkrishna9991 Only native macOS app has different icon from Windows app, Telegram Desktop on macOS has the same icon as Telegram Desktop on Windows and Linux.

I guess I was asked to leave the current icon so it won't be mistaken with macOS native app icon and they could be easily told apart from each other.

Eric-01:
> I guess I was asked to leave the current icon so it won't be mistaken with macOS native app icon and they could be easily told apart from each other.

That's a very good point. My suggestion for the Telegram design team would be to update the current (outdated) Telegram Desktop icons to the new ones in order to have consistent branding across all webpages, social media, and applications, and to distinguish and stand out a bit from the native macOS app in Swift, intentionally make Telegram Desktop icon on macOS **not** follow the new Apple's icon guidelines (i.e. icons in rounded squares).

gauthamkrishna9991:
Do we have an ETA for when the Big Sur App Icons would be updated?

If so, can we keep this PR open until that happens? I don't want to forget about doing this, and I don't want to also launch another PR for this, when (and IF) this update does happen.

Aokromes:
> Do we have an ETA for when the Big Sur App Icons would be updated?
> 
> If so, can we keep this PR open until that happens? I don't want to forget about doing this, and I don't want to also launch another PR for this, when (and IF) this update does happen.

https://github.com/telegramdesktop/tdesktop/pull/16561#issuecomment-874213061

ilya-fedin:
> If so, can we keep this PR open until that happens? I don't want to forget about doing this, and I don't want to also launch another PR for this, when (and IF) this update does happen.

For what? As @john-preston pointed, only Telegram designers have the right to change these icons, so any PRs changing the icons have no chances to be merged. In other words, there's no reason to have this PR open.

PR #16565, opened 2021-07-05
Update setup.iss

bovirus:
@ilya-fedin 
@john-preston 

Update Inno Setup compiler script with some additional info
Add in main window title the info about 32bit or 64bit version
Update Copyright info in File Properties
Add in Program list Add/remove list the info about 32bit/64bit

Please check and merge.

### DIFF
--- a/Telegram/build/setup.iss
+++ b/Telegram/build/setup.iss
@@ -35,9 +35,20 @@ DisableProgramGroupPage=no
 ArchitecturesAllowed=x64
 ArchitecturesInstallIn64BitMode=x64
 OutputBaseFilename=tsetup-x64.{#MyAppVersionFull}
+AppVerName={#MyAppName} {#MyAppVersion} (64bit)
+VersionInfoDescription={#MyAppName} 64bit installer
+VersionInfoProductName={#MyAppName} {#MyAppVersion} (64bit)
+UninstallDisplayName={#MyAppName} (64bit)
+
 #else
+AppVerName={#MyAppName} {#MyAppVersion} (32bit)
 OutputBaseFilename=tsetup.{#MyAppVersionFull}
+VersionInfoDescription={#MyAppName} 32bit installer
+VersionInfoProductName={#MyAppName} {#MyAppVersion} (32bit)
+UninstallDisplayName={#MyAppName} (32bit)
 #endif
+
+AppCopyright={#MyAppPublisher}
 
 [Languages]
 Name: "english"; MessagesFile: "compiler:Default.isl"


john-preston:
I don't see benefits here. I was going to autoupdate 32 bit versions to 64 bit versions on 64 bit hardware at some point, and with those changes I suppose I'll need to update some strings in the registry as well. I'm not sure this is important.

bovirus:
@john-preston

I beleieve that here this post is not to discuss the benfit of 32bit/64bit program version.

Currently there two version (32bit and 64bit) and my scope is to modify installer script to see in the installer the info about installing version and platform.

Eric-01:
If I can put in my two cents, I'm okay with most of the changes, but I don't like this:
> Add in Program list Add/remove list the info about 32bit/64bit

Why do you think this is a positive change? For a few years now, 64-bit programs are "the default" ones (mostly thanks to macOS and some Linux distros), so why should they be specially marked? If you really want to mark something, IMO you should do so with 32-bit versions only to draw the users' attention that they may have accidentally installed "the wrong" version on their system (similarly to how browsers now show HTTP and HTTPS connections, for example).

Eric-01:
If I can put in my two cents, I'm okay with most of the changes, but I don't like this:
> Add in Program list Add/remove list the info about 32bit/64bit

Why do you think this is a positive change? For a few years now, 64-bit programs are "the default" ones (mostly thanks to macOS and some Linux distros), so why should they be specially marked? If you really want to mark something, IMO you should do so with 32-bit versions only to draw the users' attention that they may have accidentally installed "the wrong" version on their system (similarly to how web browsers now show HTTP and HTTPS connections, for example).

bovirus:
@Eric-01 

Pelase think about current situation.
You have two type of instalelr 32bit and 64bit (and 32bit version can be instaleld on 64bit system).
But currently it register both in program list as program name without specifcation about platform then in program list you cannot check if you install 32bit or 64bit versionif you installed

Eric-01:
I'm sorry, but I don't understand. Why is it important to have such info THERE.

The vast majority of users don't even know that something like bitness even exists, so such info will crate an unnecessary visual clutter for them. For those who know about bitness, it's also not necessary as they can just launch Telegram Desktop and either:
* click the hamburger menu in the top left corner and then look at the bottom left corner
* open Task Manager and see if there is "(32 bit)" next to Telegram Desktop entry or not

They can even determine this (in most cases) based on whether Telegram Desktop was installed in Program Files or Program Files (x86).

Eric-01:
I'm sorry, but I don't understand. Why is it important to have such info THERE.

The vast majority of users don't even know that something like bitness even exists, so such info will crate an unnecessary visual clutter for them. For those who know about bitness, it's also not necessary as they can just launch Telegram Desktop and either:
* click the hamburger menu in the top left corner and then look at the bottom left corner
* open Task Manager and see if there is "(32 bit)" next to Telegram Desktop entry or not

They can even determine this (in most cases) based on whether Telegram Desktop was installed in Program Files (x86) or Program Files.

tcNmjeD6SpVrpLC:
> They can even determine this (in most cases) based on whether Telegram Desktop was installed in Program Files (x86) or Program Files.
First of all then you need to ask everyone (all companies) else who uses 64-bit installer, 32-bit installer distinction to remote them, because it is not important for you.
No, 64 bit is not really MAIN one right now. Again, not the place for discussing 32 vs 64.


> I don't see benefits here. I was going to autoupdate 32 bit versions to 64 bit versions on 64 bit hardware at some point, and with those changes I suppose I'll need to update some strings in the registry as well. I'm not sure this is important.

As he explained we need to have differences between x32 and x64 installations, now we don't have them.

tcNmjeD6SpVrpLC:
> Add in Program list Add/remove list the info about 32bit/64bit

Can you elaborate this? At least with screenshots, how we should read it? "Add info about architecture into programs list"? Because "Add/remove list the info about" making no sense, hard to understand.

Aokromes:
> > Add in Program list Add/remove list the info about 32bit/64bit
> 
> Can you elaborate this? At least with screenshots, how we should read it? "Add info about architecture into programs list"? Because "Add/remove list the info about" making no sense, hard to understand.

he means 

XLt7W2c 1

tcNmjeD6SpVrpLC:
Where i can find that menu?

bovirus:
My reuest is simple and clear.
It's available Teelgram application for 32bit and 64bit.
You can install 32bit version on 64bit OS.
All program that has similar situation to avoid misunderstanding add in main installer window and in program list info about program name wand platform. Ex.
- Main window installer
Telegram v. x.x.x.x (32bit)
Telegram v. x.x.x.x (64bit)
- Program list
Telegram (32bit)
Telegram (64bit)

Eric-01:
@tcNmjeD6SpVrpLC, I'm not talking about installer windows at all, and like I already said, I'm okay with all the changes made by bovirus to them. Whereas I don't think that info about bitness is important in the Windows list of installed programs as it can already be seen in 3 different places, so IMO 4th one is redundant and will create an unnecessary visual clutter.

john-preston:
I don't think this is needed, 32 bit version is becoming increasingly obsolete and will almost never appear on the new installations. At some point even the old 32 bit versions will be autoupdated to 64 bit versions on 64 bit systems.

PR #16663, opened 2021-07-19
Fix applying of patches in Dockerfile.

mahdi-jfri:
The original command failed for me. Tried it out and the original given path was wrong. This commit fixes those patches.


ilya-fedin:
This fix is wrong

ilya-fedin:
The right fix is to add `WORKDIR ..` after `git checkout` in patches section

mahdi-jfri:
you're right. fixed it

ilya-fedin:
'of libproxy and glibmm' part is no longer valid in commit message, you're fixing this for all the patches

mahdi-jfri:
Nice catch. Fixed that too.

john-preston:
Thanks!

PR #16665, opened 2021-07-19
window title for call window

### LINKED ISSUE #10243: Empty call window title. Difficult to find/select the window on linux desktop.
### Steps to reproduce
1.     Initiate a voice call

### Expected behaviour
The voice call window have a tittle with contact name.

### Actual behaviour
Voice call window appear with an empty title.
The voice call window appear with no title in tasks bar or windows list.

### Configuration
**Operating system:**
Gentoo Linux

Telegram Desktop 2.5.7, PC 64bit, Linux X11 glibc 2.32

**Installation source (Linux Only)**
binary release 2.5.7 from Github

**Used theme**:
Dark

eworm-de:
Having an empty window title brings different issues, for example:

* invisible on task bar
* invisible in task switcher
* apply properties by match
* ...

So let's use the contact user name for window title.

Fixes #10243

### DIFF
--- a/Telegram/SourceFiles/calls/calls_panel.cpp
+++ b/Telegram/SourceFiles/calls/calls_panel.cpp
@@ -118,7 +118,7 @@ void Panel::initWindow() {
 	window()->setAttribute(Qt::WA_NoSystemBackground);
 	window()->setWindowIcon(
 		QIcon(QPixmap::fromImage(Image::Empty()->original(), Qt::ColorOnly)));
-	window()->setTitle(u" "_q);
+	window()->setTitle(_user->name);
 	window()->setTitleStyle(st::callTitle);
 
 	window()->events(


eworm-de:
What was the previous argument `u" "_q` supposed to do?

john-preston:
@eworm-de Initially it was to hide the text on all platforms inside the window itself, this one needs to be tested (on macOS / Windows / Linux), but perhaps the in-window title can be hidden somehow else, if it is such a problem not having a window title.

eworm-de:
The current situation is really annoying at best.

Running Telegram on Linux here. The window is opened without decorations, so where is the place it needs to be hidden?

john-preston:
Ok, macOS that bothers me, I'll try to merge and test on macOS then.

john-preston:
Thanks.

PR #16668, opened 2021-07-19
set (group) call window icon

eworm-de:


ilya-fedin:
This shouldn't be needed since Qt fallbacks to application-wide icon

john-preston:
@eworm-de You should replace the setWindowIcon call with empty icon two lines above each of your added lines instead of adding a second call.

I can merge this, but there is a possibility that the reason those empty icons were set (I don't remember, why, maybe because of an old Ui::Window implementation showing those icons where they shouldn't be) is still valid, in that case I'll have to revert the PR.

ilya-fedin:
maybe just remove setWindowIcon calls at all? Qt fallbacks to QApplication::windowIcon anyway that is set via Platform::SetApplicationIcon in tdesktop

eworm-de:
Replaced the call in 06e2e9ffa009ff430e7a0e141705bcf825456b10, finally dropped it completely in 054a82c30357b1f401940fbe75ae1975981ae5c2.

eworm-de:
BTW, is there a reason that the desktop application does not use the latest icon from mobile app and website?

Aokromes:
> BTW, is there a reason that the desktop application does not use the latest icon from mobile app and website?

marketing team wants telegram desktop icon to be diferent from telegram for macos.

john-preston:
@Aokromes btw this will resolve funny, it'll be updated, but to a round icon, like right now, not BigSur style, in a white rectangle.

PR #4068, opened 2017-11-16
Optimize png images losslessly via Google's zopflipng

PeterDaveHello:
418 png images re-compressed losslessly as below, which can help deliver smaller package, save storage usage on user devices:

```
 Telegram/Resources/art/icon128.png                                                   | Bin 9318 -> 8686 bytes
 Telegram/Resources/art/icon128@2x.png                                                | Bin 35629 -> 18778 bytes
 Telegram/Resources/art/icon16.png                                                    | Bin 826 -> 774 bytes
 Telegram/Resources/art/icon16@2x.png                                                 | Bin 17183 -> 1846 bytes
 Telegram/Resources/art/icon256.png                                                   | Bin 21094 -> 18778 bytes
 Telegram/Resources/art/icon256@2x.png                                                | Bin 61383 -> 41273 bytes
 Telegram/Resources/art/icon32.png                                                    | Bin 1913 -> 1848 bytes
 Telegram/Resources/art/icon32@2x.png                                                 | Bin 19402 -> 4071 bytes
 Telegram/Resources/art/icon48.png                                                    | Bin 3065 -> 2958 bytes
 Telegram/Resources/art/icon48@2x.png                                                 | Bin 21610 -> 6257 bytes
 Telegram/Resources/art/icon512.png                                                   | Bin 47106 -> 41275 bytes
 Telegram/Resources/art/icon512@2x.png                                                | Bin 97401 -> 68944 bytes
 Telegram/Resources/art/icon64.png                                                    | Bin 4238 -> 4076 bytes
 Telegram/Resources/art/icon64@2x.png                                                 | Bin 9318 -> 8686 bytes
 Telegram/Resources/art/icon_green.png                                                | Bin 86149 -> 68198 bytes
 Telegram/Resources/art/iconbig_green.png                                             | Bin 24316 -> 20117 bytes
 Telegram/Resources/art/logo_256.png                                                  | Bin 21094 -> 18778 bytes
 Telegram/Resources/art/logo_256_no_margin.png                                        | Bin 24982 -> 22515 bytes
 Telegram/Resources/icons/add_contact_phone.png                                       | Bin 380 -> 202 bytes
 Telegram/Resources/icons/add_contact_phone@2x.png                                    | Bin 708 -> 344 bytes
 Telegram/Resources/icons/add_contact_user.png                                        | Bin 316 -> 166 bytes
 Telegram/Resources/icons/add_contact_user@2x.png                                     | Bin 579 -> 263 bytes
[...]

PR #16669, opened 2021-07-19
re-compress preview image

eworm-de:
The preview image takes some time to load, just re-compressing it can
save about 13% of data without changing the result.


eworm-de:
This is done with [oxipng](https://github.com/shssoichiro/oxipng).

There are a lot of png files in the source tree, some of these can be compressed even better - with saving up to 70%.
Are you interested in re-compressing all of them?

ilya-fedin:
see #4068

eworm-de:
Ok, looks like we do not want the resources to be re-compressed. The preview image is loaded by the browser when visiting the GitHub page, though.

PR #16793, opened 2021-08-10
Mention some missing includes in payments/ subdirectory

mymedia2:
not sure where it comes from but compiler refuses to proceed :cry:


ilya-fedin:
These includes are in PCH

john-preston (changes requested):

PR #16794, opened 2021-08-10
Disable crash reports entirely through macro

mymedia2:
This saves compile time and binary size. Guarding code by preprocessor is the only effective way to guarantee its elimination, even in most modern C++. And more importantly, one do not need to care of it for porting to old Qt.


ilya-fedin:
Why are you doing that?

mymedia2:
@ilya-fedin I didn't ask you meaning of life

john-preston (changes requested):
@mymedia2 I don't like the idea of doing this using preprocessor. I think there should be core/not_started_window module, that contains code without preprocessor that is always built and core/crash_report_window module, that uses this core/not_started_window module. core/crash_report_window module should be excluded from the build in CMake using remove_target_sources in case of crash reports not being used.

ilya-fedin:
Closing due to no changes in 2 years

PR #16795, opened 2021-08-10
simplified the if condition

DavidRomanovizc:
I think this will make this condition more beautiful


john-preston:
Say 'Hi' to the Dash guy! #16788

mymedia2:
Well, you'd increased `if` nesting. Huge nesting, in general, isn't a sign of good code.

PR #16867, opened 2021-08-30
Use `error` function

sammiee5311:
Replaced 2 lines with `error` function since `error` function is defined.

### DIFF
--- a/Telegram/configure.py
+++ b/Telegram/configure.py
@@ -47,8 +47,7 @@ elif officialTarget in ['win64', 'uwp64']:
 if officialTarget != '':
     officialApiIdFile = scriptPath + '/../../DesktopPrivate/custom_api_id.h'
     if not os.path.isfile(officialApiIdFile):
-        print("[ERROR] DesktopPrivate/custom_api_id.h not found.")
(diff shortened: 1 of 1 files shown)


john-preston:
Thanks for you efforts. Sorry, I've decided not to merge PRs that won't affect the app end-users in some meaningful way.

sammiee5311:
The reason why I created this PR is that there are duplicated `finish` functions so i thought it is better to define only function once and use it to other python files. ([configure.py](https://github.com/telegramdesktop/tdesktop/blob/c1a7332a5ec46b7137dc4752e5366c7c4ae8c7c9/Telegram/configure.py#L16), [set_version.py](https://github.com/telegramdesktop/tdesktop/blob/3343880ed0e5a86accc7334af54b3470e29ee686/Telegram/build/set_version.py#L10), [updates.py](https://github.com/telegramdesktop/tdesktop/blob/3343880ed0e5a86accc7334af54b3470e29ee686/Telegram/build/updates.py#L36))

But I guess it wont affect the app end-users in some meaningful way..

Anyway, thanks for comment !

Issue #3845, opened 2017-08-29
Proposal: Better RTL detection


MadaraUchiha:
Howdy!

A recurring theme I've noticed in most chat programs today (Telegram included 😢) is that text direction of a given message is determined by the first (relevant) character, which is a shame because:

1. If I start my message with a link, the message is instantly LTR.
2. If I use mixed language in a message (insert English words inside of Hebrew/Arabic/Whatever) and the first word is of the wrong direction, the message looks very bad.

So I propose a better algorithm which isn't much more complicated to detect the desired direction of a message, count the number of character in each language/direction (excluding links), and the one with more characters in it wins. More formally:

1. Strip the message of links
2. Let x be the number of RTL characters in the message. 
3. Let y be the number of LTR characters in the message.
4. If x > y, set direction to RTL, finish.
5. If x < y, set direction to LTR, finish.
6. If x equals y, use the current algorithm (based on the first character)

Examples of this algorithm can be seen with Google Hangouts (which is the only chat I can tell that actually has a smarter algorithm than "look at the first character"). 

Of course, this is a proposal and the concrete algorithm is open to change, but I think that it's a very good compromise between code complexity and correctness.

stek29:
This would be slow tho, and not easy to implement.
Also, what about emoji, or numbers, or other symbols used in both layouts?

MadaraUchiha:
Once per message? (or even once per keystroke?) doesn't sound too slow. It's not something you continually need to do per frame, and if you choose the more UX-y per keystroke, you can just "remember" the current counters for the message on the side, and just increase the relevant one.

As for numbers/emoji/media/symbols, they do not count for either RTL or LTR. (just like they don't today, if the first character is a number, it would look at the second character, etc) 

![image](https://user-images.githubusercontent.com/1188951/29860226-03c7ca12-8d6e-11e7-9a9b-9ef2615fe11a.png)

Can you point me towards the relevant piece of code where the direction is selected? I'm willing to PR this.

john-preston:
I can try to alter the way it works for messages in bubbles, but not in the message input field. Is the problem there as well?

MadaraUchiha:
The problem is there as well, yes. And I do think that it's good if we can change it there as well, but the bubbles are more prominent (you write once, read many times and by many people). I can prepare an example of how that can be done in a fiddle if you'd like, I don't think it would be more than 10 LoC

Socialdarwinist:
[...]

PR #16893, opened 2021-09-03
Export CSS: per-line automatic bidi direction

### LINKED ISSUE #16120: When you export chats with RTL messages containing LTR words, the words will have bad order.
### Steps to reproduce

1. In a group type messages in RTL direction like
    1. <code><span dir="rtl">سلام hello درود</span></code>
    2. <code><span dir="rtl">سلام👋</span></code>
    3. <code><span dir="rtl">با پاپیتا میتوانی به جمله <span dir="ltr">What is خرمالو in English?</span> جواب بدهی.</span></code> that contains:
        1. <code>با پاپیتا میتوانی به جمله</code>
        2. Start of left-to-right embedding
        3. <code>What is خرمالو in English?</code>
        4. Pop directional formatting
        5. <code dir="rtl">جواب بدهی.</code>
2. Export the chat history in HTML.

### Actual/Unexpected behavior

It shows the messages incorrectly as shown in the photo:
current order of the words
HTML texts are:

<pre><code>       &ltdiv class="text"&gt
سلام hello درود
       &lt/div&gt</pre></code>

<pre><code>       &ltdiv class="text"&gt
سلام👋
       &lt/div&gt</pre></code>

<pre><code>       &ltdiv class="text"&gt
با پاپیتا میتوانی به جمله  What is خرمالو in English?  جواب بدهی.
       &lt/div&gt</pre></code>

I don't know HTML too much but 1 solution that I know is using span.

<pre><code>       &ltdiv class="text"&gt
&ltspan dir="rtl"&gtسلام hello درود&lt/span&gt
       &lt/div&gt</code></pre>

<pre><code>       &ltdiv class="text"&gt
&ltspan dir="rtl"&gtسلام👋&lt/span&gt
       &lt/div&gt</code></pre>

<pre><code>       &ltdiv class="text"&gt
&ltspan dir="rtl"&gtبا پاپیتا میتوانی به جمله &ltspan dir="ltr"&gtWhat is خرمالو in English?&lt/span&gt جواب بدهی.&lt/span&gt
       &lt/div&gt</code></pre>
and also I think solving this issue is related to the [bidi algorythm](https://www.unicode.org/reports/tr9/).

### Configuration

**Operating system:** PC 32bit, Windows 7

**The version of Telegram Desktop:** 2.7.1

**Used theme:** Classic

cben:
Fixes #16120.
`unicode-bidi: plaintext` makes browser auto-detect base bidi direction according to bidi algorithm heuristics, on each line separately.
https://drafts.csswg.org/css-writing-modes/#bidi-algo
AFAICT this matches both Android and Desktop behaviors, and is the friendly behavior for mixed-direction users.
Purely LTR users should not be affected — the heuristics return LTR even on "neutral" lines that contain no letters, only punctuation/emojis/digits.
(http://www.unicode.org/reports/tr9/#P2 — rule P3 defaults to level 0, which is LTR.)
Whereas all RTL language users in practice need to deal with mixed-direction texts, and the ability to split lines to prevent LTR/RTL mixups is extremely handy...

The `text-align: start` ensures LTR-detected lines are displayed left-aligned, and RTL-detected lines right-aligned.
Which again matches both Android and Desktop behaviors.
This is redundant — it seems already the default with current CSS — but is important for mixed-language users, so I wanted to ensure it in case somebody sets alignment on bigger parts of the document in the future.

## before
Ekrankopio de 2021-09-03 13-25-15

## after
Ekrankopio de 2021-09-03 13-34-18

Tested by editing exported CSS from Telegram 2.7.1 on Linux, viewed with Chrome 92.

### DIFF
--- a/Telegram/Resources/export_html/css/style.css
+++ b/Telegram/Resources/export_html/css/style.css
@@ -291,6 +291,8 @@ div.selected {
 .default .text {
     word-wrap: break-word;
     line-height: 150%;
+    unicode-bidi: plaintext;
+    text-align: start;
 }
 .default .reply_to,
 .default .media_wrap {


cben:
Note that if #3845 gets implemented — custom direction heuristics — there will be no simple CSS way to match it with flat text + `<br>` tags, Telegram would need to emit separate element per line with explicit `dir` attributes.
But for now I think this is fine.

One thing I haven't confirmed is whether per-line auto direction indeed works on all platforms.  I only have Android & Linux.
- [x] Android
- [ ] iOS?
- [x] Linux
- [ ] Windows?
- [ ] Mac?

john-preston:
Thanks!

Issue #16905, opened 2021-09-05
Changing volume with media controls doesn't work


udf:
### Steps to reproduce

1. Play an audio file
2. Change volume with media controls (e.g. `playerctl -p tdesktop volume 0.1`)

### Expected behaviour

Playback volume should change.

### Actual behaviour

Slider updates with the correct value, but the playback volume does not change.

### Operating system

Arch Linux

### Version of Telegram Desktop

3.0.1

### Installation source

Static binary from official website

udf:
Related to #16276

PR #16906, opened 2021-09-05
Fix song volume controls, closes #16276 #16905

### LINKED ISSUE #16276: Volume Control works only when I slide it
### Steps to reproduce
1. Play a music
2. Slide (click and drag) volume down
3. Click on higher volume

### Expected behaviour
Volume should rise

### Actual behaviour
It doesn't rise

### Configuration
**Operating system:**
Ubuntu 18.04
Telegram Desktop 2.7.4
**Installation source (Linux Only)** - the official website / GitHub releases / flatpak / snap / distribution package:
the official website
**Used theme**:
Tinted
</details>

udf:
The audio mixer tracks changes to the volume using the settings:
https://github.com/telegramdesktop/tdesktop/blob/5000902d61d556960c4afe20f7624933fd914499/Telegram/SourceFiles/media/audio/media_audio.cpp#L588-L596
However it uses the internal value `_volumeSong` to set the volume, so if the setting is updated before the mixer value, the previous value is used.

When the slider value changes, the value in settings gets updated before the mixer value, causing the previous value to be used in the mixer, which causes #16276

Additionally, setting the volume with media controls only updates the setting, causing the playback volume to not change, which causes #16905

### DIFF
--- a/Telegram/SourceFiles/media/player/media_player_volume_controller.cpp
+++ b/Telegram/SourceFiles/media/player/media_player_volume_controller.cpp
@@ -72,8 +72,8 @@ void VolumeController::setVolume(float64 volume) {
 
 void VolumeController::applyVolumeChange(float64 volume) {
 	if (volume != Core::App().settings().songVolume()) {
+		mixer()->setSongVolume(volume);
 		Core::App().settings().setSongVolume(volume);
-		mixer()->setSongVolume(Core::App().settings().songVolume());
 	}
 }
 
--- a/Telegram/SourceFiles/media/system_media_controls_manager.cpp
+++ b/Telegram/SourceFiles/media/system_media_controls_manager.cpp
@@ -248,6 +248,9 @@ SystemMediaControlsManager::SystemMediaControlsManager(
 		_controls->volumeChangeRequests(
 		) | rpl::start_with_next([](float64 volume) {
 			Player::mixer()->setSongVolume(volume);
+			if (volume > 0) {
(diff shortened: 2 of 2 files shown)


Telegram/SourceFiles/media/player/media_player_volume_controller.cpp
```diff
@@ -72,8 +72,8 @@ void VolumeController::setVolume(float64 volume) {
 
 void VolumeController::applyVolumeChange(float64 volume) {
 	if (volume != Core::App().settings().songVolume()) {
+		mixer()->setSongVolume(volume);
 		Core::App().settings().setSongVolume(volume);
-		mixer()->setSongVolume(Core::App().settings().songVolume());
```
23rd, line 76: What is the point of this change?

Telegram/SourceFiles/media/system_media_controls_manager.cpp
```diff
@@ -246,6 +247,7 @@ SystemMediaControlsManager::SystemMediaControlsManager(
 
 		_controls->volumeChangeRequests(
 		) | rpl::start_with_next([](float64 volume) {
+			Player::mixer()->setSongVolume(volume);
```
23rd, line 250:
```suggestion
			Player::mixer()->setSongVolume(volume);
			if (volume > 0) {
				Core::App().settings().setRememberedSongVolume(volume);
			}
```

john-preston:
Thanks!

PR #16910, opened 2021-09-06
Migrate to no-response action since github app is archived

ilya-fedin:


Eric-01:
Perhaps GitHub's official Stale bot (https://github.com/marketplace/stale) could be used instead of going for workflows?

ilya-fedin:
> Perhaps GitHub's official Stale bot (https://github.com/marketplace/stale) could be used instead of going for workflows?

stale bot has nothing to do with no-response

Issue #16922, opened 2021-09-08
thread unsafe calls to setlocale() between glib and Qt may crash at startup


glebius:
### Steps to reproduce

Telegram 2.8.11_2 from FreeBSD ports collection. FreeBSD CURRENT operating system. The problem should also exist on Linux as well. Chances to crash are pretty low. Usually the crash happens on a freshly booted machine, when the file cache is cold and setlocale() actually needs to go to disk to get some data.

### Expected behaviour

Do not crash.

### Actual behaviour

The crash happens in thread g_main_loop_run(), which already receives a message from network, but Qt haven't yet finished with its init. Message processing calls into glib, which calls into system libc, which finds locale not a valid structure, yet:

> Core was generated by `telegram-desktop'.                                                                                                                                                     
> Program terminated with signal SIGSEGV, Segmentation fault.                                                                                                                                   
> #0  0x0000000809e0c26e in localeconv_l (loc=loc@entry=0x809f26588 <__xlocale_global_locale>) at /usr/src/git/lib/libc/locale/localeconv.c:91                                                                                                                                         
> 91              M_ASSIGN_CHAR(int_p_cs_precedes);                                                                                                                                             
> [Current thread is 1 (LWP 101403)]                                                                                                                                                            
> (gdb) bt                                                                                                        
> #0  0x0000000809e0c26e in localeconv_l(loc=loc@entry=0x809f26588 <__xlocale_global_locale>) at /usr/src/git/lib/libc/locale/localeconv.c:91                                                                                                                                         
> #1  0x0000000809e63477 in __vfprintf(fp=0x7fffdfbf99f8, locale=0x809f26588<__xlocale_global_locale>, fmt0=0x809d86646 "%s.db", ap=0x7fffdfbf9b30) at /usr/src/git/lib/libc/stdio/vfprintf.c:480                                                                                                                                           
> #2  0x0000000809e61d00 in snprintf(str=0x7fffdfbf9ba0 " \240\277\337\377\177", n=1023, fmt=0x809d86646 "%s.db") at /usr/src/git/lib/libc/stdio/snprintf.c:74                                                                                                                                              
> #3  0x0000000809e001d7 in seq_open_db(cl=0x80de3c4f0, name=0x7fffdfbfa080 "/usr/share/i18n/esdb/esdb.alias") at /usr/src/git/lib/libc/iconv/citrus_lookup.c:147                                                                                                                                      
[...]

PR #16925, opened 2021-09-09
Close a race between Gio threads and Qt application init.

glebius:
QApplication has a historical feature[1] of calling setlocale(3) at init,
which today cause more harm than good.  Such call isn't thread safe neither
per C standard, not per any known libc implementation.  The Gio threads
would call into locale reading functions all the time, so early access to
locale by Gio may race with setlocale() by Qt leading to undefined behavior.

Platform specific start before the Sandbox (aka QApplication) is at core
design of tdesktop and that order can't be changed. There is no way to pause
Gio until QApplication instantiates.

Fortunately, Qt library itself has a static global flag that prevents it
from calling setlocale() twice. We don't even need to instantiate a
temporary QCoreApplication, we can just call into the method.  So call
it right before launching Gio.

[1] https://chat.stackoverflow.com/rooms/63812/discussion-between-phil-armstrong-and-matteo-italia


ilya-fedin:
I don't like having such a call in IsIBusPortalPresent, it feels like a hack. IsIBusPortalPresent can be moved to other place and that call will become useless.

john-preston:
@glebius Thank you very much for this research.

@ilya-fedin We have calls to `Glib::init` and `Gio::init` in `Platform::Launcher::exec`. As the docs state "Gio::init() calls Glib::init(), which sets the global locale as specified by Glib::set_init_to_users_preferred_locale()". (maybe we should leave only `Gio::init` call then?)

I suggest placing the `QCoreApplicationPrivate::initLocale();` call there with a link to #16922. I'm not sure if it should go before `Gio::init` or after (this will result in one or another locale call being made last). I think it can go before, that way it surely won't race with gio code even if it decides to spawn threads and do some other strange things right in `init` call.

ilya-fedin:
> maybe we should leave only `Gio::init` call then?

https://github.com/GNOME/glibmm/blob/master/glib/glibmm/init.cc#L41-L46

ilya-fedin:
> I suggest placing the `QCoreApplicationPrivate::initLocale();` call there with a link to #16922. I'm not sure if it should go before `Gio::init` or after (this will result in one or another locale call being made last). I think it can go before, that way it surely won't race with gio code even if it decides to spawn threads and do some other strange things right in `init` call.

`init` call is a glibmm's (C++ wrapper's) thing, I don't think it will start threads. IMHO, the right place for `QCoreApplicationPrivate::initLocale();`  is in Platform::start, right after mallctl call

john-preston:
@ilya-fedin Ok, let it be `Platform::start` then.

john-preston (changes requested):
> IMHO, the right place for QCoreApplicationPrivate::initLocale(); is in Platform::start, right after mallctl call.

ilya-fedin:
https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/gdbusprivate.c#L317-331
изображение

ilya-fedin:
Since dbus connection object is being destroyed when leaving IsIBusPortalPresent, I would expect the thread to be destoryed as well. But they don't destroy it.

glebius:
Thanks! Will update request in a few minutes.

P.S.
> Since dbus connection object is being destroyed when leaving IsIBusPortalPresent, I would expect the thread to be destoryed as well. But they don't destroy it.

Even if they destroyed them, the race would still be there.

ilya-fedin:
> Even if they destroyed them, the race would still be there.

How it can be there if there would be no thread that doing that?

ilya-fedin:
Also, you should target dev branch, not master branch

glebius:
> How can it be there if there would be no thread doing that?

You are right, that if that code wasn't commented out, the race would not exist. The code actually doesn't destroy threads, the threads cease themself. The commented code waits for them to complete their operation, basically removing any parallelism after gio object is destroyed. Note that g_thread_join() is commented out twice :)

glebius:
> Also, you should target dev branch, not master branch

Thanks, will do!

glebius:
I failed to change target branch for the GitHub pull request, so looks like I need to create a new one. Sorry for that.

ilya-fedin:
You can change target branch, but you can't change source branch. One of the reasons it's a good idea to create a custom branch for PR.

PR #16946, opened 2021-09-12
Implement close to taskbar option

### LINKED ISSUE #1161: "Close to taskbar" option on Windows
Please add this option. Most messengers on Windows has a similar option, some of us don't use the systray anymore as a primary method of notification.

Example implementations:
On Skype: "Keep Skype on taskbar while I'm signed in"
On Pidgin: Automatically minimize to taskbar on close if "Show system tray icon" is set to "Never".

ilya-fedin:
Closes #1161

### DIFF
--- a/Telegram/SourceFiles/core/core_settings.cpp
+++ b/Telegram/SourceFiles/core/core_settings.cpp
@@ -220,7 +220,8 @@ QByteArray Settings::serialize() const {
 			<< qint32(_disableOpenGL ? 1 : 0)
 			<< _photoEditorBrush
 			<< qint32(_groupCallNoiseSuppression ? 1 : 0)
-			<< qint32(_voicePlaybackSpeed * 100);
+			<< qint32(_voicePlaybackSpeed * 100)
+			<< qint32(_closeToTaskbar.current() ? 1 : 0);
 	}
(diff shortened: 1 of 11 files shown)


Telegram/SourceFiles/core/core_settings.cpp
```diff
@@ -221,6 +221,8 @@ QByteArray Settings::serialize() const {
 			<< _photoEditorBrush
 			<< qint32(_groupCallNoiseSuppression ? 1 : 0)
 			<< qint32(_voicePlaybackSpeed * 100);
+		stream
```
23rd, line 224:
```suggestion
```

Telegram/SourceFiles/core/core_settings.cpp
```diff
@@ -221,6 +221,8 @@ QByteArray Settings::serialize() const {
 			<< _photoEditorBrush
 			<< qint32(_groupCallNoiseSuppression ? 1 : 0)
 			<< qint32(_voicePlaybackSpeed * 100);
```
23rd, line 223:
```suggestion
			<< qint32(_voicePlaybackSpeed * 100)
```

PR #16956, opened 2021-09-14
Couple small fixes to compile on FreeBSD

glebius:
Together with pull requests to other components, tdesktop/dev compiles and works perfectly on FreeBSD.

https://github.com/telegramdesktop/libtgvoip/pull/35
https://github.com/desktop-app/tg_owt/pull/69
https://github.com/desktop-app/cmake_helpers/pull/123

Note: the zip.h problem, I believe, can show up on a certain Linux distributions as well. And of course won't show up on FreeBSD if zlib package isn't installed. In other words, it isn't actually related to the topic.

### DIFF
--- a/Telegram/SourceFiles/platform/linux/specific_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/specific_linux.cpp
@@ -46,7 +46,12 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 #include <QtGui/QWindow>
 
 #include <private/qguiapplication_p.h>
+
+#ifdef Q_OS_FREEBSD
+#include <malloc_np.h>
+#else // Q_OS_FREEBSD
 #include <jemalloc/jemalloc.h>
+#endif // Q_OS_FREEBSD
 
 #ifndef DESKTOP_APP_DISABLE_DBUS_INTEGRATION
 #include <glibmm.h>
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -454,7 +454,7 @@ parts:
   webrtc:
     source: https://github.com/desktop-app/tg_owt.git
     source-depth: 1
-    source-commit: 91d836dc84a16584c6ac52b36c04c0de504d9c34
+    source-commit: 575fb17d2853c43329e45f6693370f5e41668055
     plugin: cmake
     build-packages:
       - yasm


Telegram/CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ if (WIN32)
     generate_midl(Telegram ${src_loc}/platform/win/windows_quiethours.idl)
 endif()
 
-set_target_properties(Telegram PROPERTIES AUTOMOC ON AUTORCC ON)
+set_target_properties(Telegram PROPERTIES AUTOMOC ON AUTORCC ON NO_SYSTEM_FROM_IMPORTED ON)
```
ilya-fedin, line 48: I believe this should create a lot of problems in official builds. The use -Werror and tdesktop uses INTERFACE targets a lot. If includes of third-party libraries won't use -isystem, that would lead to various errors.

ilya-fedin:
> Note: the zip.h problem, I believe, can show up on a certain Linux distributions as well. And of course won't show up on FreeBSD if zlib package isn't installed. In other words, it isn't actually related to the topic.

What zip.h problem are you talking about?

ilya-fedin:
Oh, I see in the commit description. I believe we should find another way to fix this problem.

Telegram/CMakeLists.txt
ilya-fedin: E.g. I remember I had a problem with -Wparentheses, clang and gtk headers

ilya-fedin:
> This is a very questionable default of cmake. Most libraries we (or any other normal project) depend on are not system libraries, they don't belong to the OS, they are third party.

For me it's clear: third party libraries should use -isystem. -isystem indicates the code you can't influence on and makes the checks less strict.

Telegram/SourceFiles/platform/linux/specific_linux.cpp
```diff
@@ -46,7 +46,11 @@ For license and copyright information please follow this link:
 #include <QtGui/QWindow>
 
 #include <private/qguiapplication_p.h>
+#ifdef __FreeBSD__
```
ilya-fedin, line 49: Use Q_OS_FREEBSD, please

ilya-fedin: And I think that block should have an empty line after the previous include with qguiapplication_p.h

Telegram/SourceFiles/platform/linux/specific_linux.cpp
```diff
@@ -46,7 +46,12 @@ For license and copyright information please follow this link:
 #include <QtGui/QWindow>
 
 #include <private/qguiapplication_p.h>
+
+#ifdef Q_OS_FREEBSD
+#include <malloc_np.h>
+#else
```
ilya-fedin, line 52:
```suggestion
#else // Q_OS_FREEBSD
```

Telegram/SourceFiles/platform/linux/specific_linux.cpp
```diff
@@ -46,7 +46,12 @@ For license and copyright information please follow this link:
...
 #include <jemalloc/jemalloc.h>
+#endif
```
ilya-fedin, line 54:
```suggestion
#endif // !Q_OS_FREEBSD
```

ilya-fedin:
About -isystem: try to get compiler commandline of the failing file and check the order of -isystem's. Maybe we can fix the order and the problem will be fixed.

glebius:
> About -isystem: try to get compiler commandline of the failing file and check the order of -isystem's. Maybe we can fix the order and the problem will be fixed.

The order starts with /usr/local/include and has /usr/local/include/minizip closer to its end. I spent a lot of time on that trying to figure our if there any way to control that. Will look more.

glebius:
Now this pull request has only malloc_np.h change left.

ilya-fedin:
You can update cmake_helpers submodule and update tg_owt commit in snapcraft.yaml

.gitmodules
```diff
@@ -49,6 +49,7 @@
 [submodule "cmake"]
 	path = cmake
 	url = https://github.com/desktop-app/cmake_helpers.git
+	branch = 3c8e9a0c759a136ac8c5999b7e5c2c0c43319887
```
ilya-fedin, line 52: WTF? All was needed is:
```
cd cmake
git fetch --all
git checkout origin/master
cd ..
git add cmake
```

glebius: I'm sorry. First time using submodules. Will fix ASAP.

john-preston:
Thanks!

PR #16962, opened 2021-09-16
Fix compilation in presence of libzip installed.

glebius:
When both minizip [1] and libzip [2] are present in the build
environment we have two <zip.h> includes. We are interested in
the minizip. Unfortunately, libzip usually installs its zip.h
straight into ${PREFIX}/include where lots of other headers reside.
We pick up ${PREFIX}/include into the include path with the
desktop-app::lib_base target, which is also dependency of
of tdesktop::td_* targets. To fix compilation in presence of
conflicting zip.h we need to put minizip's include directory
before ${PREFIX}/include, thus record its dependency before
all other dependencies that can bring desktop-app::lib_base.

[1] http://www.winimage.com/zLibDll/minizip.html
[2] https://libzip.org

### DIFF
--- a/Telegram/CMakeLists.txt
+++ b/Telegram/CMakeLists.txt
@@ -52,6 +52,14 @@ PRIVATE
     tdesktop::lib_tgcalls_legacy
     tdesktop::lib_tgcalls
     tdesktop::lib_tgvoip
+
+    # Order in this list defines the order of include paths in command line.
+    # We need to place desktop-app::external_minizip this early to have its
+    # include paths (usually ${PREFIX}/include/minizip) before any depend that
+    # would add ${PREFIX}/include. This path may have a different <zip.h>,
+    # for example installed by libzip (https://libzip.org).
+    desktop-app::external_minizip
+
     tdesktop::td_export
     tdesktop::td_mtproto
     tdesktop::td_lang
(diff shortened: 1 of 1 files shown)


john-preston:
Let's put a short version of this explanation in a comment right above the new desktop-app::external_ line, with additional empty lines before the comment and after the line.

ilya-fedin:
Won't that cause problems when static linking? :thinking: 
If gcc strips unused symbols from statically linked llibraries and other libraries using minizip link after minizip (lib_base?), that would lead to linking errors.

ilya-fedin:
CI should show if there's such a problem, though

ilya-fedin:
@glebius I wanted to use sdbus-cpp for dbus communication rather than glibmm initially. Its API is a way more modern, supports various modern C++ features, and needs less code for the same actions. The only reason glib was chosen is that its more portable. I know there's basu now, how easy for FreeBSD would be to port sdbus-cpp if tdesktop switches to it?

glebius:
@ilya-fedin "sd" stands for systemd, is that correct? I'm afraid porting it to anything that is not a modern Linux is difficult.

ilya-fedin:
> "sd" stands for systemd, is that correct?

Yeah, FreeBSD already has the needed extracted part
https://www.freebsd.org/cgi/ports.cgi?query=basu&stype=all

sdbus-cpp is a C++ wrapper, so replacing includes should be enough for porting (I'm not sure, though, that's why I'm asking you).

ilya-fedin:
@glebius If you know Russian, tell your nickname at Telegram and @john-preston will add you to tdesktop contributions group (don't forget to allow adding you to groups in your privacy settings).

glebius:
@ilya-fedin facepalm_tar_gz

glebius:
Это не попытка вставить эмодзи, это ник в Телеграме.

PR #17068, opened 2021-10-07
Display dates with the system date format

### LINKED ISSUE #618: Client should use the date format defined by the system
Right now the date is shown in DD.MM.YY format on the list of conversations. This is not how it's shown everywhere else on my system.

It should ideally take the format from the system.

CoderTimZ:
Fixes #618

### DIFF
--- a/Telegram/SourceFiles/api/api_authorizations.cpp
+++ b/Telegram/SourceFiles/api/api_authorizations.cpp
@@ -82,7 +82,7 @@ Authorizations::Entry ParseEntry(const MTPDauthorization &data) {
 			&& lastDate.weekNumber() == nowDate.weekNumber()) {
 			result.active = langDayOfWeek(lastDate);
 		} else {
-			result.active = lastDate.toString(qsl("d.MM.yy"));
+			result.active = lastDate.toString(cDateFormat());
 		}
 	}
 
--- a/Telegram/SourceFiles/core/application.cpp
+++ b/Telegram/SourceFiles/core/application.cpp
@@ -260,6 +260,7 @@ void Application::run() {
 
 	DEBUG_LOG(("Application Info: inited..."));
 
+	cChangeDateFormat(QLocale::system().dateFormat(QLocale::ShortFormat));
 	cChangeTimeFormat(QLocale::system().timeFormat(QLocale::ShortFormat));
 
 	DEBUG_LOG(("Application Info: starting app..."));
--- a/Telegram/SourceFiles/data/data_peer.cpp
+++ b/Telegram/SourceFiles/data/data_peer.cpp
@@ -1090,7 +1090,7 @@ std::optional<QString> RestrictionError(
 			auto restrictedUntil = channel->restrictedUntil();
 			if (restrictedUntil > 0 && !ChannelData::IsRestrictedForever(restrictedUntil)) {
 				auto restrictedUntilDateTime = base::unixtime::parse(channel->restrictedUntil());
-				auto date = restrictedUntilDateTime.toString(qsl("dd.MM.yy"));
+				auto date = restrictedUntilDateTime.toString(cDateFormat());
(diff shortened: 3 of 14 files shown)


23rd:
The style changes can be squashed with the first commit.

landry314:
I don't see the changes in Telegram 8.2 in regards to this commit. Still says, for example:
Screen Shot 2021-11-03 at 10 56 59 PM

Aokromes:
> I don't see the changes in Telegram 8.2 in regards to this commit.
> 
> Still says, for Oct 8, 2021: Screen Shot 2021-11-03 at 10 56 59 PM
> 
> My system settings are this: Screen Shot 2021-11-03 at 10 59 13 PM

this is telegram desktop not telegram android / ios / macos native.

landry314:
You really shouldn't say this then:

"Telegram Desktop – Official Messenger

This is the complete source code and the build instructions for the alpha version of the official desktop client for the Telegram messenger, based on the Telegram API and the MTProto secure protocol."

Aokromes:
> You really shouldn't say this then:
> 
> "Telegram Desktop – Official Messenger
> 
> This is the complete source code and the build instructions for the alpha version of the official desktop client for the Telegram messenger, based on the Telegram API and the MTProto secure protocol."

there are 2 telegram macos clients, the native and this, earch have a diferent place to report bugs, and this have version 3.2 not 8.2.

FichteFoll:
Unfortunately, it appears that the change didn't work as expected for me. I'm on 3.1.11 (Arch Linux) and I do not get `dd.mm.yyyy` anymore in my chat history, but instead I now get `dd/mm/yyyy` where I would expect `yyyy-mm-dd` as per my `C_TIME=en_DK.UTF-8` locale config.

I'm not used to C++ or Qt for that matter, but I wasn't able to find any definition of `cDateFormat` or `cSetDateFormat` anywhere over the internet, including the entirety of Github, so I was unable to confirm whether `C_TIME` is respected at all. For the record, this is what `date` prints out:

```
~ $ date +"%x %X"
2021-11-05 20:26:13
```

Aokromes:
> Unfortunately, it appears that the change didn't work as expected for me. I'm on 3.1.11 (Arch Linux) and I do not get `dd.mm.yyyy` anymore in my chat history, but instead I now get `dd/mm/yyyy` where I would expect `yyyy-mm-dd` as per my `C_TIME=en_DK.UTF-8` locale config.
> 
> I'm not used to C++ or Qt for that matter, but I wasn't able to find any definition of `cDateFormat` or `cSetDateFormat` anywhere over the internet, including the entirety of Github, so I was unable to confirm whether `C_TIME` is respected at all. For the record, this is what `date` prints out:
> 
> ```
> ~ $ date +"%x %X"
> 2021-11-05 20:26:13
> ```

you very likely need 3.2.x

anibyl:
I have a similar issue to @FichteFoll: after the last update date format in recent chats switched from `dd.mm.yy` to `dd mmm yyyy`, whereas `date` has a different output:
```
~$ date +"%x %X"
2021-11-05 17:47:05
```
Version 3.2.2

john-preston:
@anibyl Now the app uses "QLocale::system().dateFormat(QLocale::ShortFormat)" for those dates, so I guess it is a question about Qt locale support 🙄 or is it used improperly?

teohhanhui:
It's actually not a bug, unfortunately.

https://bugreports.qt.io/browse/QTBUG-72451

anibyl:
Well, at least I'm glad that the issue is not in Telegram anymore, and these new date formats that we see are somewhat more familiar/preferable to us.

FichteFoll:
Thank you for investigating. I can confirm that using `LC_TIME=lt_LT.UTF-8`, which I am already using for Thunderbird, functions as expected, since fortunately Telegram Desktop only uses the short format.

Issue #3334, opened 2017-04-27
Have an option to configure default Delete action on app level


iklimchuk:
![screenshot - 4_27_2017 6_05_59 pm](https://cloud.githubusercontent.com/assets/10230808/25493389/4631a436-2b76-11e7-8006-d8c26e54758d.png)
### Steps to reproduce
1. Open any chat
2. Right-click on your message and select "Delete"

### Actual behavior
I see confirmation dialog with unchecked "**Delete for ....**" checkbox by default

### Expected behavior
I'd like to have a way to configure default value of this check box. As for me (and I believe for other users as well) 99.(9)% of cases when I delete a message - my intention is to have it deleted for all chat members, deleting of message just for me is quite useless (IMHO, at least I cannot imagine a use case when I really need it). But currently I should manually select this option (Delete for all) every time, which is quite annoying. Moreover, if for some reason you missed this checkbox and deleted a message from your chat only (what happens pretty often when you're in rush and trying to delete a message as fast as it is possible) - you basically have no ways to delete it from other members chats anymore...and you're in trouble :)
It will be really great to have an option in settings which allows to configure default action for delete (for me/for all) and stop asking user about it every time he/she deletes a message.

### Configuration
**Operating system:**
Win10

**Version of Telegram Desktop:**
1.0.29

**Used theme**:
default

matheus208:
It's really frustrating UX because when I click delete my instinct tells me to instantly click "OK", which leads to the message being deleted only for me and I can't find it again to delete it for everyone else.

stek29:
@matheus208 even if you found it, API won't allow to revoke messages you've already deleted for yourself, even if you do it instantly after deletion without revoke.

dogancelik:
Native Mac client has this checked by default which I like, but ability to change the default state is really good for user experience and less frustrating for users who prefer it checked.

stek29:
@john-preston What do you think about remembering last checked/unchecked state?

KimTheFirst:
I have just encountered this issue directly after sending a very embarrassing message to the wrong person, deleting it in a panic, and then realizing that I had now made the message undeletable. 

I can now definitively say that in my opinion, having this option off by default is completely retarded to the utmost extreme in all use cases that I can possibly think of. 

Is there any compelling scenario whatsoever in which it would be desirable to delete a message for yourself and not for the other person?

RememberTheAir:
[...]

PR #17268, opened 2021-11-18
Option "also delete for contact" being checked by default

AmirHosseinCV:
I think the checkbox should be checked by default. Because when you send something wrong, you want to delete it as fast as possible. In this situation, you may forget to check the checkbox, and if you delete it only for yourself, there is no way to delete the message for the other side. Also, we usually want to delete messages for both sides.

### DIFF
--- a/Telegram/SourceFiles/boxes/delete_messages_box.cpp
+++ b/Telegram/SourceFiles/boxes/delete_messages_box.cpp
@@ -158,7 +158,7 @@ void DeleteMessagesBox::prepare() {
 				_revoke.create(
 					this,
 					revoke->checkbox,
-					false,
+					true,
 					st::defaultBoxCheckbox);
 				appendDetails(std::move(revoke->description));
 			} else if (peer->isChannel()) {


soredake:
Related: https://github.com/telegramdesktop/tdesktop/issues/3334

RememberTheAir:
This would be inconsistent with all the other Telegram apps work

AmirHosseinCV:
@RememberTheAir
> This would be inconsistent with all the other Telegram apps work

All the other Telegram apps = Only the android version, right?
Also, I think they can talk about it with the android developers, or we can create another pull request on their GitHub repository, too.
When something is wrong, we should change it.

RememberTheAir:
> All the other Telegram apps = Only the android version, right?

Android app, iOS app, and one of the two web clients (the other one doesn't offer a toggle, but two different dialog options "_delete for me_" and "_delete for me and [name]_")

> we can create another pull request on their GitHub repository

The Android app maintainer will not accept pull requests for such changes. Anyway, a suggestion to check that toggle by default has been posted on the suggestions platform: https://bugs.telegram.org/c/2089

> When something is wrong, we should change it.

Telegram is not a community-driven project, even if you open a pull request to change this kind of behavior on all their apps, I highly doubt it will ever be merged. These kind of changes usually require a lot of people asking for such a change, and I guess this one in particular is not going to be accepted

AmirHosseinCV:
Sorry, I had deleted my comment a minute before you send this. Looks like you're right.

AmirHosseinCV:
@mymedia2 Good idea, but I can't do it. I would appreciate it if someone would do it.

stanleyguevara:
@mymedia2 that seems like a workaround for something that shouldn't need workaround in first place. I'd say upvoting and commenting on the generic suggestion to change the default is the right way: https://bugs.telegram.org/c/2089

john-preston:
The state of the checkbox can be saved now.

PR #17319, opened 2021-12-01
Add support to build on M1 Mac

cnSchwarzer:
On M1 mac, some dependencies build two `arm64` builds and makes `lipo` fail.


john-preston:
Thanks!

PR #17324, opened 2021-12-01
Revert "Use kernel accelerated sendfile to copy files on Linux"

### LINKED ISSUE #16884: Update fails on Fedora Linux
### Steps to reproduce
1. Click the blue "Update Telegram" button in the main app window.

### Expected behaviour

The application restarts using the binary executable of the new version.

### Actual behaviour

The application closes but cannot start because the Telegram binary executable file is corrupted.  I once compared the corrupted file that I had locally with the solid one that I downloaded from the website, and it turned out that the updater failed to write the whole file: the contents in the bad and in the good file were identical up to the EOF that was too early in the bad file

### Configuration
**Operating system:** Fedora Linux, version 34 at the moment, but I had this issue on earlier versions.

**Version of Telegram Desktop:** I experience this issue for quite a long time, perhaps a few months.  It worked before, but stopped working a few releases ago.

**Installation source (Linux Only)** - the official website / GitHub releases / flatpak / snap / distribution package: The official website.

**Used theme**: The standard one (I use the dark mode sometimes).

ilya-fedin:
Fixes #16884

### DIFF
--- a/Telegram/SourceFiles/_other/updater_linux.cpp
+++ b/Telegram/SourceFiles/_other/updater_linux.cpp
@@ -8,7 +8,6 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 #include <cstdio>
 #include <sys/stat.h>
 #include <sys/types.h>
-#include <sys/sendfile.h>
 #include <cstdlib>
 #include <unistd.h>
 #include <dirent.h>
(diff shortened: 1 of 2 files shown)


gbcox:
This was fixed, but with today's (January 14, 2022) update, it appears there has been a regression to this fix.  Problem  #16884 is back -  where the new version is automatically download, but has been corrupted.

PR #17361, opened 2021-12-08
Fix drag and drop on Linux

vimpostor:
If the `QDrag` object is destroyed too early, drag and drop will not work on Linux, because communication will still have to happen later between source and target, see: https://freedesktop.org/wiki/Specifications/XDND/

This patch increases the lifetime of `drag`, by changing the type to a raw pointer. This does not create a memory leak, as parent is specified in the `QDrag` constructor, which will free it later.

If you know a better way to increase the lifetime of the `drag` and don't like this "Qt way" of memory management, let me know. :)

In any case this finally fixes Drag and drop on Linux.

### DIFF
--- a/Telegram/SourceFiles/window/main_window.cpp
+++ b/Telegram/SourceFiles/window/main_window.cpp
@@ -964,7 +964,9 @@ int MainWindow::tryToExtendWidthBy(int addToWidth) {
 void MainWindow::launchDrag(
 		std::unique_ptr<QMimeData> data,
 		Fn<void()> &&callback) {
-	auto drag = std::make_unique<QDrag>(this);
+	// Qt destroys this QDrag automatically after the drag is finished
+	// We must not delete this at the end of this function, as this breaks DnD on Linux
+	auto drag = new QDrag(this);
 	drag->setMimeData(data.release());
 	drag->exec(Qt::CopyAction);
 


ilya-fedin:
> This does not create a memory leak, as parent is specified in the `QDrag` constructor, which will free it later.

MainWindow has the lifetime of entire tdesktop run time, so you will create, and create, and create QDrags that will be actually destroyed only when quitting the app.

john-preston:
@vimpostor yes, that's not cool, they'll be leaking till the program termination. What is the moment when it is safe to destroy QDrag? When is it supposed to be destroyed?

ilya-fedin:
It seem QDrag is owned by QDragManager or by QPA either
изображение

Weird this is not documented

vimpostor:
Yes Qt will take care of deleting the `QDrag` once the drag is finished. Unfortunately this is not the case, if `QPlatformDrag.ownsDragObject()` returns `true` for a platform and I don't know of any public API to test this.

Also see https://forum.qt.io/topic/77894/where-to-delete-qdrag-and-qmimedata-from-startdrag-event-memory-leak/5

vimpostor:
> @vimpostor What is the moment when it is safe to destroy QDrag? When is it supposed to be destroyed?

The exact moment cannot be detected reliably. It is usually a few milliseconds after `QDrag::exec()` returns. Unfortunately even  `deleteLater()` is too early for this.

ilya-fedin:
If ownsDragObject returns true then Qt still owns the object, just its QPA part

vimpostor:
> If ownsDragObject returns true then Qt still owns the object, just its QPA part

What about the early returns in `QDragManager::drag()`. I think we still need to delete it ourself in case `Qt::IgnoreAction` is returned, like also mentioned in the above Qt Forum post.

ilya-fedin:
If ownsDragObject returns true it's logical that delete happens in the QPA code, if it returns false, QDragManager does delete on its own to avoid such kind of problems when it's not known who should delete the object, Qt always delete it.

vimpostor:
> If ownsDragObject returns true it's logical that delete happens in the QPA code, if it returns false, QDragManager does delete on its own to avoid such kind of problems when it's not known who should delete the object, Qt always delete it.

So change needed for my patch, right?

ilya-fedin:
yeah, that's right

vimpostor:
I have amended a little comment, would you mind taking a look again? @john-preston

PR #17441, opened 2021-12-26
Mark as single-instance using X-GNOME-SingleWindow key

Pointedstick:
Telegram is a single-instance app; attempting to open a second instance
fails and simply focuses the existing instance. However the app's
.desktop file does not indicate this to Linux desktop shells like GNOME
and KDE Plasma, causing them to show a "Start new instance/Open new
window" action for the app that will not work when used.

This commit fixes that by setting `X-GNOME-SingleWindow=true` in the
desktop file. This is a GNOME-specific key, but it is honored in KDE
Plasma too.

### DIFF
--- a/lib/xdg/telegramdesktop.desktop
+++ b/lib/xdg/telegramdesktop.desktop
@@ -13,6 +13,7 @@ MimeType=x-scheme-handler/tg;
 Keywords=tg;chat;im;messaging;messenger;sms;tdesktop;
 Actions=Quit;
 X-GNOME-UsesNotifications=true
+X-GNOME-SingleWindow=true
 
 [Desktop Action Quit]
 Exec=telegram-desktop -quit


ilya-fedin:
Well, that's not really true, you can launch multiple instances by specifying other data directory with `-workdir` CLI argument

ilya-fedin:
That's why tdesktop has md5 hash of data directory in the name of its single instance socket

Pointedstick:
> Well, that's not really true, you can launch multiple instances by specifying another data directory with -workdir CLI argument

If that's the case, that's cool, but such functionality is not of much use to desktop shells like GNOME and KDE Plasma, which parse this key to determine whether or not they should show a "Start new instance" or "open new window" menu item for Telegram. This desktop file key is only for them and would not interfere with your ability to start a new instance using that command-line option.

Pointedstick:
Thanks for merging!

PR #17451, opened 2021-12-30
 Create SECURITY.md 

Foge9627:
[](url)


Foge9627:
مساعد

john-preston:
The versioning is linear for tdesktop, so I don't really understand what does "supported" mean. Other information is available on https://telegram.org, like here: https://telegram.org/faq#q-why-should-i-trust-you

PR #17465, opened 2021-12-30
add space before EDITED in bare - #fix 17464

Hessam-Koucheki:


john-preston:
What? It is a JSON key in chats export, why would you add a space there.

PR #23869, opened 2022-01-09
Change launcher name

aequabit:
Not sure if this is something that is in-scope, if not I would maintain it in my own fork.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+    "files.associations": {
+        "*.php": "php",
+        "numbers": "cpp"
+    }
+}
--- a/Telegram/SourceFiles/core/update_checker.cpp
+++ b/Telegram/SourceFiles/core/update_checker.cpp
@@ -39,6 +39,8 @@ extern "C" {
 #include <openssl/err.h>
 } // extern "C"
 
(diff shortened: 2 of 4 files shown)


john-preston:
I'm not sure what did you want to do, but I hope it was something different.

aequabit:
Yes, the second commit shouldn't have been in the PR, my bad.

john-preston:
@aequabit but why would you change launcher name of Telegram Desktop (!) from "telegramdesktop" to "teamgramdesktop"? 🙈

aequabit:
> @aequabit but why would you change launcher name of Telegram Desktop (!) from "telegramdesktop" to "teamgramdesktop"? see_no_evil

Since I could not find a document explaining the intention and scope of this project (if there is one, please link to it), I assumed it was to to build a separate messenger downstream from the official Telegram one. I have since switched from building on my main machine to a VM, so running both applications at the same time is not an issue anymore.

To clarify: Do you intend your forks of the clients to connect to official Telegram servers and is the fork of NebulaChat just an alternative to that, or is the project for hosting everything yourself? If so, I think it would be helpful to update the build instructions so they don't suggest that you require Telegram API tokens.

I should've probably clarified that before submitting a pull request, though.

Issue #23877, opened 2022-01-10
Telegram doesn't respect QT_QPA_PLATFORM=wayland anymore


eyalzek:
### Steps to reproduce

Run telegram with native wayland, e.g:

```bash
env QT_QPA_PLATFORM=wayland telegram-desktop -- %u
```
either from cli or by updating the desktop file.

This worked up until 3.4.3.

### Expected behaviour

Telegram should run as a wayland application.

### Actual behaviour

Telegram runs as X

### Operating system

Debian 11

### Version of Telegram Desktop

3.4.3

### Installation source

Apt installation from Debian sources.

### Logs

```shell
$ env QT_QPA_PLATFORM=wayland telegram-desktop -- %u

(telegram-desktop:10145): Telegram-WARNING **: 12:27:53.254: Application has been built with foreign rlottie, animated emojis won't be colored to the selected pack.

(telegram-desktop:10145): Telegram-WARNING **: 12:27:53.254: Application was built without embedded fonts, this may lead to font issues. On Debian-based systems, make sure you have the fonts-open-sans package installed
Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
error: : cannot open
error: : cannot open
error: : cannot open
qt.svg: Error while inflating gzip file: SVG format check failed
Corrupt JPEG data: premature end of data segment
Corrupt JPEG data: premature end of data segment
```

ilya-fedin:
> Static binary from official website

> ```shell
> (telegram-desktop:10145): Telegram-WARNING **: 12:27:53.254: Application has been built with foreign rlottie, animated emojis won't be colored to the selected pack.
> 
> (telegram-desktop:10145): Telegram-WARNING **: 12:27:53.254: Application was built without embedded fonts, this may lead to font issues. On Debian-based systems, make sure you have the fonts-open-sans package installed
> ```

It seems there's some confusion since static binary from official website can't print these lines.

eyalzek:
Hey, sorry. It's installed via apt (debian sources):
```bash
$ apt policy telegram-desktop 
telegram-desktop:
  Installed: 3.4.3+ds-1
  Candidate: 3.4.3+ds-1
  Version table:
 *** 3.4.3+ds-1 500
        500 http://deb.debian.org/debian testing/main amd64 Packages
        100 /var/lib/dpkg/status
```

ilya-fedin:
[...]

PR #23897, opened 2022-01-14
Use QT_QPA_PLATFORM if provided.

hgaiser:
This PR fixes https://github.com/telegramdesktop/tdesktop/issues/23877 .

After this change it should still default to xorg, like with Qt5, while respecting `QT_QPA_PLATFORM` if it is set.


hgaiser:
Should I add a comment in the code why that check is there?

PR #23944, opened 2022-01-20
Fix incorrect admin log

c0re100:
Cause [GenerateParticipantChangeText](https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/SourceFiles/history/admin_log/history_admin_log_item.cpp#L470-L492) haven't handled `Restricted` to `Member` log, so admin or bot delete(remove) a user's exception.

You will see you're invited a user:
{8A95E9D6-3547-4537-A248-9C24A5F90265}

After new commit, admin log should be shown correctly.
{203B3E02-8B4F-4C2A-A826-D1D1D6771D3E}

I think `GenerateBannedChangeText` renamed to `GeneratePermissionsChangeText`, it would be more suitable, cause `Banned` and `Restricted >> Member` now sharing same method.

### DIFF
--- a/Telegram/SourceFiles/history/admin_log/history_admin_log_item.cpp
+++ b/Telegram/SourceFiles/history/admin_log/history_admin_log_item.cpp
@@ -219,7 +219,7 @@ TextWithEntities GenerateAdminChangeText(
 	return result;
 };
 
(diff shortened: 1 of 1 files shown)


john-preston:
Thanks!

PR #23976, opened 2022-01-26
Remove unused variables

klemensn:
.

### DIFF
--- a/Telegram/SourceFiles/chat_helpers/gifs_list_widget.cpp
+++ b/Telegram/SourceFiles/chat_helpers/gifs_list_widget.cpp
@@ -44,7 +44,6 @@ namespace ChatHelpers {
 namespace {
 
 constexpr auto kSearchRequestDelay = 400;
-constexpr auto kInlineItemsMaxPerRow = 5;
 constexpr auto kSearchBotUsername = "gif"_cs;
 constexpr auto kMinRepaintDelay = crl::time(33);
 constexpr auto kMinAfterScrollDelay = crl::time(33);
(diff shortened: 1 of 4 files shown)


john-preston:
Thanks! The last one being not used is a bug :) Also there is one more kInlineItemsMaxPerRow in inline_results_inner.h that's not used as well.

klemensn:
I can rebase once you committed a fix and also remove the one I missed.

john-preston:
Here we go.

john-preston:
Please squash the changes into a single commit.

Telegram/SourceFiles/main/main_session.cpp
```diff
@@ -364,12 +364,6 @@ void Session::uploadsStopWithConfirmation(Fn<void()> done) {
 	const auto id = _uploader->currentUploadId();
 	const auto exists = !!data().message(id);
 	auto box = Box([=](not_null<Ui::GenericBox*> box) {
-		const auto label = box->addRow(
-			object_ptr<Ui::FlatLabel>(
-				box.get(),
-				tr::lng_upload_sure_stop(),
-				st::boxLabel),
-			st::boxPadding + QMargins(0, 0, 0, st::boxPadding.bottom()));
```
23rd, line 372: No, it should be kept here. Maybe we should add the `[[maybe_unused]]` attribute.

klemensn: That or someone makes use of `label`.
Either way will fix `-Werror` builds.

I removed it since that seemed straight forward.
I'll use the attribute.

john-preston:
Ok, I guess you can resolve conflicts and we'll check the build.

PR #24066, opened 2022-02-12
Fix dir separator on windows

XiangQingW:
Fix dir separator on windows

### DIFF
--- a/Telegram/build/prepare/prepare.py
+++ b/Telegram/build/prepare/prepare.py
@@ -27,7 +27,7 @@ os.chdir(scriptPath + '/../../../..')
 
 dirSep = '\\' if win else '/'
 pathSep = ';' if win else ':'
-libsLoc = 'Libraries' if not win64 else 'Libraries/win64'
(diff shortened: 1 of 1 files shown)


XiangQingW:
The %LIBS_DIR%\\Qt-5.15.2 dir doesn't been generated after running `tdesktop\Telegram\build\prepare\win.bat`.

23rd:
Maybe it would be better to use `dirSep` instead.

XiangQingW:
@23rd I have just replaced it with `dirSep`, please review again.

john-preston:
I wonder how this fixes anything. There is code later that still uses the universal separator:
```
if not os.path.isdir(libsDir + '/' + keysLoc):
    pathlib.Path(libsDir + '/' + keysLoc).mkdir(parents=True, exist_ok=True)
```
I'm afraid the problem may be somewhere else, in the build qt / install qt steps.

ilya-fedin:
btw, there's os.path.join in python

john-preston:
Ok, I don't really understand, how this fixes anything, but the change looks benign.

PR #24119, opened 2022-02-27
[Linux] Clear read message notifications from history

vimpostor:
This fixes notifications not being cleared from history on Linux, once the chat with that person (whom the notifications originated from) is opened.
The notification reference must only be removed when the notification is really cleared - and not already when the notification has timed out.
See commit message for more details.

### DIFF
--- a/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
@@ -695,7 +695,21 @@ void NotificationData::setImage(const QString &imagePath) {
 }
 
 void NotificationData::notificationClosed(uint id, uint reason) {
-	if (id == _notificationId) {
+	/*
+	 * From: https://specifications.freedesktop.org/notification-spec/latest/ar01s09.html
+	 * The reason the notification was closed
+	 * 1 - The notification expired.
+	 * 2 - The notification was dismissed by the user.
+	 * 3 - The notification was closed by a call to CloseNotification.
+	 * 4 - Undefined/reserved reasons.
+	 *
+	 * If the notification was dismissed by the user (reason == 2), the notification is not kept in notification history.
+	 * We do not need to send a "CloseNotification" call later to clear it from history.
+	 * Therefore we can drop the notification reference now.
+	 * In all other cases we keep the notification reference so that we may clear the notification later from history,
+	 * if the message for that notification is read (e.g. chat is opened or read from another device).
+	*/
+	if (id == _notificationId && reason == 2) {
 		_manager->clearNotification(_id);
 	}
 }
--- a/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
@@ -74,6 +74,9 @@ std::unique_ptr<base::Platform::DBus::ServiceWatcher> CreateServiceWatcher() {
 				const Glib::ustring &oldOwner,
 				const Glib::ustring &newOwner) {
 				if (activatable && newOwner.empty()) {
+					crl::on_main([] {
+						Core::App().notifications().clearAll();
+					});
 					return;
 				}
 
@@ -665,7 +668,9 @@ void NotificationData::close() {
 			_notificationId,
 		}),
 		{},
-		std::string(kService));
+		std::string(kService),
+		-1,
+		Gio::DBus::CALL_FLAGS_NO_AUTO_START);
 	_manager->clearNotification(_id);
 }
 
--- a/Telegram/SourceFiles/history/history_inner_widget.cpp
+++ b/Telegram/SourceFiles/history/history_inner_widget.cpp
@@ -2570,7 +2570,7 @@ void HistoryInner::checkHistoryActivation() {
 	}
 	adjustCurrent(_visibleAreaBottom);
 	if (_history->loadedAtBottom() && _visibleAreaBottom >= height()) {
-		// Clear possible scheduled messages notifications.
+		// Clear possible message notifications.
 		Core::App().notifications().clearFromHistory(_history);
 	}
 	if (_curHistory != _history || _history->isEmpty()) {


ilya-fedin:
But when it's really cleared through? This looks like adding a memory leak...

ilya-fedin:
You also can remove Manager::clearNotification as it becomes unused.

vimpostor:
> But when it's really cleared through? This looks like adding a memory leak...

This is already explained in the commit message. It happens when the chat is opened with an user who caused a notification.
Perhaps this backtrace makes it clearer for you:

```gdb
Thread 1 "telegram-deskto" hit Breakpoint 1, Platform::Notifications::Manager::Private::clearNotification (this=0x7fff97e0cc80, id=...) at Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp:985
985     void Manager::Private::clearNotification(NotificationId id) {
(gdb) backtrace
#0  Platform::Notifications::Manager::Private::clearNotification(Window::Notifications::Manager::NotificationId)
    (this=0x7fff97e0cc80, id=...)
    at Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp:985
#1  0x000055555a9c7b02 in Platform::Notifications::Manager::clearNotification(Window::Notifications::Manager::NotificationId)
    (this=0x7fff9c7a9280, id=...)
    at Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp:1008
#2  0x000055555a9c4390 in Platform::Notifications::(anonymous namespace)::NotificationData::close() (this=0x7fff9c76c700)
    at Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp:648
#3  0x000055555a9c75e3 in Platform::Notifications::Manager::Private::clearFromHistory(gsl::not_null<History*>)
    (this=0x7fff97e0cc80, history=...)
    at Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp:960
#4  0x000055555a9c7d8b in Platform::Notifications::Manager::doClearFromHistory(gsl::not_null<History*>)
    (this=0x7fff9c7a9280, history=...)
    at Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp:1040
#5  0x000055555b158a79 in Window::Notifications::Manager::clearFromHistory(gsl::not_null<History*>)
    (this=0x7fff9c7a9280, history=...)
    at Telegram/SourceFiles/window/notifications_manager.h:239
#6  0x000055555b15152d in Window::Notifications::System::clearFromHistory(gsl::not_null<History*>) (this=
    0x7fffe38f6f80, history=...) at Telegram/SourceFiles/window/notifications_manager.cpp:321
#7  0x00005555599a1fe9 in HistoryInner::checkHistoryActivation() (this=0x7fffde75e200)
    at Telegram/SourceFiles/history/history_inner_widget.cpp:2548
```

So basically https://github.com/telegramdesktop/tdesktop/blob/feb2d3066e7143409644c5dcc3c2dc0f66dcfed6/Telegram/SourceFiles/history/history_inner_widget.cpp#L2548 is the entry point for when it is cleared again - so no memory leak.

Perhaps I could change the `// Clear possible scheduled messages notifications.` comment above that line, since it also clears notifications from chats in general, but that would just be a minor thing.

ilya-fedin:
> This is already explained in the commit message.

Oh, my bad, sorry, I checked only the PR description

ilya-fedin:
I also wonder what if the daemon doesn't have the history and reuses IDs? clearNotification can happen at random point of time and close a notification from another application...

ilya-fedin:
I believe this is a KDE bug. They shouldn't emit the signal or it's unknown whether the ID is hold by the server or not. I believe GNOME also has notification history and doesn't send a signal because as far as I heard, Telegram notifications present in history there.

john-preston:
@vimpostor How do you clear notification if the chat was read from a different device? 🤔

vimpostor:
> I see another problem through. Snap needs a StartService call to activate an activatable service before sending a request. Previously it was assumed that Close call can happen only until notification is hidden and notification daemon process exited, therefore it's not needed to be wrapped. Now, it seems the call can be done far long after the daemon has exited.

I don't understand what you mean. The Close call still happens afterwards, why would the notification daemon exit?

> I also wonder what if the daemon doesn't have the history and reuses IDs? clearNotification can happen at random point of time and close a notification from another application...

No, this is impossible, as the [Freedesktop spec](https://specifications.freedesktop.org/notification-spec/notification-spec-latest.html) explicitly prohibits this:
```
Each notification displayed is allocated a unique ID by the server.
This is unique within the session.
While the notification server is running, the ID will not be recycled unless the capacity of a uint32 is exceeded.
```

> I believe this is a KDE bug. They shouldn't emit the signal or it's unknown whether the ID is hold by the server or not. I believe GNOME also has notification history and doesn't send the signal because as far as I heard, Telegram notifications present in history there.

Nah, I think you are just misunderstanding the code. This is not a KDE bug, it is totally expected that the CloseNotification signal is emited after the 5 second timeout. In fact, it would be a bug if a notification server would not emit that. But again this is irrelevant anyway, since IDs are not recycled (see above).

vimpostor:
> @vimpostor How do you clear notification if the chat was read from a different device? thinking

Right now I don't, notifications are cleared only if they are read on the same device.
But you are right, I guess it would make sense to also clear them when they are read on another device...

ilya-fedin:
> I don't understand what you mean. The Close call still happens afterwards, why would the notification daemon exit?

Have you heard about [GApplication](https://docs.gtk.org/gio/class.Application.html) services? This type of services is very popular in gtk world, they dbus-activatable (start on demand) and close in multiple seconds (usually 15) after they have nothing to do. MATE's notification daemon is such a service. I think Ubuntu MATE has notification history connected with Unity Indicators, we should check how it works: does it still exit before the notification removed from the history, does it reuse ID then (the upstream MATE without history does), does CloseNotification call affects this somehow and does it allow activate actions in the history.

> No, this is impossible, as the [Freedesktop spec](https://specifications.freedesktop.org/notification-spec/notification-spec-latest.html) explicitly prohibits this:

As you can see, there's no guarantee it's still running. Maybe we need to add tracking of notification daemon and then remove all references when it exits...

vimpostor:
> they dbus-activatable (start on demand) and close in multiple seconds (usually 15) after they have nothing to do. MATE's notification daemon is such a service.

That sounds very broken, have you checked if this is even true?
Try the following and check if it prints different notification IDs (which would be good) or the same IDs:
```bash
gdbus call --session --dest=org.freedesktop.Notifications --object-path=/org/freedesktop/Notifications --method=org.freedesktop.Notifications.Notify 'Telegram Desktop' 0 'mail-message' "Notification 0" "Test body 0" '[]' '{}' 5000
sleep 16
gdbus call --session --dest=org.freedesktop.Notifications --object-path=/org/freedesktop/Notifications --method=org.freedesktop.Notifications.Notify 'Telegram Desktop' 0 'mail-message' "Notification 1" "Test body 1" '[]' '{}' 5000
```

I suspect that the notification daemon still does not reuse IDs, even despite it technically being restarted as you claim.

> As you can see, there's no guarantee it's still running. Maybe we need to add tracking of notification daemon and then remove all references when it exits...

That would be very cumbersome. Right now we don't know if your claim is even true, but if Mate does indeed reuse notification IDs (which I doubt) due to the service being stopped entirely, then I'd rather whitelist DE's with sane notification servers than keep track of the notification daemon.
There aren't even that many DEs outside of the major 2 that support a notification history in the first place.

ilya-fedin:
> Try the following and check if it prints different notification IDs (which would be good) or the same IDs:

It reuses, but needs a longer sleep since notifications don't immediately disappear:
```
(uint32 2,)
(uint32 1,)
```

> then I'd rather whitelist DE's with sane notification servers than keep track of the notification daemon.

This seem to follow the spec + I don't like per-DE workarounds

vimpostor:
> It reuses, but needs a longer sleep since notifications don't immediately disappear:

How long until it starts reusing?


> This seem to follow the spec + I don't like per-DE workarounds

I would hardly call "restarting the service, so that it doesn't technically count as the same session" as following the spec:
```
This [notification ID] is unique within the session.
```

Also for someone not liking per-DE workarounds, there seems to be quite a lot of DE-specific workarounds for dark mode detection in the codebase...

Well if we were to somehow track the notification daemon, how would that even be possible, I don't think there is an API for that around.

ilya-fedin:
> How long until it starts reusing?

when notification disappears + 15 seconds I think. I haven't measured how long notification is shown, but it's also a couple of seconds.

> Also for someone not liking per-DE workarounds, there seems to be quite a lot of DE-specific workarounds for dark mode detection in the codebase...

Yeah, I hate that :cry:. There were no standard at all though.

> Well if we were to somehow track the notification daemon, how would that even be possible, I don't think there is an API for that around.

There's already everything for this, see CreateServiceWatcher function in the file. I just need to check Ubuntu MATE to see how the logic should work exactly.

ilya-fedin:
> as following the spec

The next sentence says "While the notification server is running, the ID will not be recycled unless the capacity of a uint32 is exceeded." what means it works as intended. Maybe it's a bug in the spec that it's not clear, though, I don't know.

ilya-fedin:
Ok, Ubuntu MATE's history doesn't have actions, so it's just a read-only history that seem to spy d-bus socket and have no integration with notification daemon. So this should be enough:
```
diff --git a/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp b/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
index 4b83a1d19..03130947e 100644
--- a/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
@@ -74,6 +74,9 @@ std::unique_ptr<base::Platform::DBus::ServiceWatcher> CreateServiceWatcher() {
 				const Glib::ustring &oldOwner,
 				const Glib::ustring &newOwner) {
 				if (activatable && newOwner.empty()) {
+					crl::on_main([] {
+						Core::App().notifications().clearAll();
+					});
 					return;
 				}
 
```

vimpostor:
> Oh, we also need to avoid daemon startup when clearing notifications, so:

Thanks, I have added that diff in a separate commit with you as author, as I don't like claiming other peoples code as mine. ;)

vimpostor:
> @vimpostor How do you clear notification if the chat was read from a different device? thinking

@john-preston Actually this also already works without any code changes, it is already called from `History::inboxRead()`:
https://github.com/telegramdesktop/tdesktop/blob/2917a53b757a41beb2db59ac4ce2adb307b063bc/Telegram/SourceFiles/history/history.cpp#L1559

vimpostor:
Do not merge yet. I think it makes sense to only keep the notification reference in the internal manager, when the NotificationClosed reason was that it timed out. If the user explicitly closes the notification, it is not moved to the notification history anyway, so we don't need to clear it from history.

It doesn't really hurt to send a `CloseNotification` for a notification that is not in history anymore (it is a noop and not an error), but still it is an unnecessary call so I will fix this.

Aokromes:
> Do not merge yet. 

you can change it to draft while it's incomplete.

vimpostor:
> you can change it to draft while it's incomplete.

Will do it that way next time. But it is implemented and fixed now, so ready for review/merge again.

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -695,7 +700,21 @@ void NotificationData::setImage(const QString &imagePath) {
 }
 
 void NotificationData::notificationClosed(uint id, uint reason) {
-	if (id == _notificationId) {
+	/*
+	 * From: https://specifications.freedesktop.org/notification-spec/latest/ar01s09.html
```
ilya-fedin, line 704:
```
enum class CloseReason {
	TimedOut = 1,
	UserCanceled = 2,
	ApplicationHidden = 3,
	Reserved = 4,
}
```

You can put this before the ServerInformation struct, replace uint with the enum in the function argument, and do switch-case like in the Windows backend: https://github.com/telegramdesktop/tdesktop/blob/d5b9effd2e0ba1f2a0ae1941f8f8d9f481ca8362/Telegram/SourceFiles/platform/win/notifications_manager_win.cpp#L784-L794

ilya-fedin: And it would be nice then to have the id check in not-increasing-indention manner:
```
if (id != _notificationId) {
 	return;
}
```

john-preston:
Ok, let's test that.

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
vimpostor: Sorry I didn't see this before it got merged. But since these are just minor codestyle changes, feel free to make your own PR, if you still feel like this is important. ;)

ilya-fedin: It's not, maybe will do if would change something in this file

john-preston:

PR #24126, opened 2022-03-02
Introduce TDESKTOP_DISABLE_LEGACY_TGVOIP

klemensn:
Originally from Alt Linux[0], OpenBSD has so far adapted the removal of
tgvoip in the official net/tdesktop build.

tgcalls provides everything needed for calls;  audio/video/desktop
sharing calls have been working fine across different operating systems
and telegram desktop/mobile versions without problems.

0: http://www.sisyphus.ru/cgi-bin/srpm.pl/Sisyphus/telegram-desktop/getpatch/1

### DIFF
--- a/Telegram/CMakeLists.txt
+++ b/Telegram/CMakeLists.txt
@@ -31,7 +31,9 @@ get_filename_component(res_loc Resources REALPATH)
 include(cmake/telegram_options.cmake)
 include(cmake/lib_ffmpeg.cmake)
 include(cmake/lib_stripe.cmake)
-include(cmake/lib_tgvoip.cmake)
+if (NOT TDESKTOP_DISABLE_LEGACY_TGVOIP)
+    include(cmake/lib_tgvoip.cmake)
+endif()
 include(cmake/lib_tgcalls.cmake)
 include(cmake/td_export.cmake)
 include(cmake/td_mtproto.cmake)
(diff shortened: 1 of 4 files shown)


ilya-fedin:
> sharing calls have been working fine across different operating systems
> and telegram desktop/mobile versions without problems.

Even with Telegram X for Android? :thinking:

rakino (changes requested):
Rebasing against these two commits is needed: https://github.com/telegramdesktop/tdesktop/commit/2b5d61ad36e96c46ff46793004e3ebc9d59535cb https://github.com/telegramdesktop/tdesktop/commit/84af084a3bc6408a11ad6da8b285b6b3792d739b

ilya-fedin:
There's no really need to, it was discussed in the tdesktop contributions group that there will be no possibility to build without libtgvoip. Either all tdesktop builds should support it until @john-preston decides it's the time to drop the support.

rakino:
Thanks for the explanation!

Another question is that, I wonder if there's a bugreport related to this issue mentioned in the packaging guide:
> so the calls will crash due to two webrtc libraries being in one address spacey

I didn't find that while searching through, would you mind giving me a pointer?

Thanks again :)

ilya-fedin:
This shouldn't ever happen if you let tdesktop to build libtgvoip. If you're trying to build libtgvoip with its own build scripts, it's unlikely to ever work right as they're abandoned.

Issue #17061, opened 2021-10-05
[Feature Request] org.freedesktop.appearance.color-scheme support to read dark mode preference


vimpostor:
Currently the "Follow global dark theme preference" switch on Linux uses an unreliable implementation.
Recently the `org.freedesktop.appearance.color-scheme` key got standardized across all DEs, which should be more reliable. It would be cool, if Telegram supported reading this setting and super cool if it could subscribe to its changes during runtime.

Useful links:
[Gnome blog post](https://blogs.gnome.org/alexm/2021/10/04/dark-style-preference/)
[Spec](https://github.com/flatpak/xdg-desktop-portal/blob/d7a304a00697d7d608821253cd013f3b97ac0fb6/data/org.freedesktop.impl.portal.Settings.xml#L33-L45)
[KDE initial implementation](https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/merge_requests/49) + [proper implementation](https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/merge_requests/52)

ilya-fedin:
Finally. I'm tried of this dancing with DE-specific keys.

ilya-fedin:
Well, I implemented that locally, but I will push only when it would be possible to test how it works. Basically, that means 'after support of this preference will land to my desktop'. I have MATE on NixOS on host and a VM with KDE Neon. So, that means a MATE release with that feature in NixOS or a KDE release with that feature in KDE Neon.

carnage-mode:
Are those four implementations cross platform or Linux specific?

carnage-mode:
Fair enough :+1:

vimpostor:
> IsDarkMode has four implementations and it's already big. I don't really want to add a fifth implementation.

The clean way would be to make an upstream Qt patch, such that `QGuiApplication::paletteChanged` triggers with the new key.
However I am also calling bullshit here and fail to see why adding the new key and keeping the 4 other methods as fallback would be a problem. Worst-case scenario you can remove the most unreliable of the 4 methods and replace it with the new one.

> I don't feel this spec will have a big adoption outside of GNOME & KDE since it's portal-based.

Then the other DEs should implement Portal.

ilya-fedin:
> The clean way would be to make an upstream Qt patch

The feature will reach in multiple years to tdesktop, then. There's no updates for Qt 5 and there's no plan to migrate to Qt 6 in near time.

[...]

PR #24133, opened 2022-03-02
Prevent false-positive when matching dark themes

WhyNotHugo:
References: https://github.com/telegramdesktop/tdesktop/issues/17061


WhyNotHugo:
Note: I signed a DCO, **not** a CLA as the bot above indicates.

Telegram/SourceFiles/platform/linux/specific_linux.cpp
```diff
@@ -499,9 +499,12 @@ std::optional<bool> IsDarkMode() {
 
 			result = false;
 
-			if (gtkThemePortalString.contains(
+			if (gtkThemePortalString.endsWith(
```
ilya-fedin, line 502: This should be in sync with the xesttings part

WhyNotHugo: Updated.

ilya-fedin: Maybe it's better to create a helper function in private namespace? IsGtkDarkTheme or something like that.

WhyNotHugo: Yes, will do.

ilya-fedin:
```c++
bool IsGtkDarkTheme(const QString &theme) {
	const auto themeLower = theme.toLower();
	return themeLower.contains(qsl("-dark-")) || themeLower.endsWith(qsl("-dark"));
}
```

Telegram/SourceFiles/platform/linux/specific_linux.cpp
```diff
@@ -499,9 +502,7 @@ std::optional<bool> IsDarkMode() {
 
 			result = false;
 
-			if (gtkThemePortalString.contains(
-				qsl("-dark"),
-				Qt::CaseInsensitive)) {
+			if (IsGtkDarkTheme(gtkThemeX)) {
```
23rd, line 505:
```suggestion
			if (IsGtkDarkTheme(gtkThemePortalString)) {
```

Telegram/SourceFiles/platform/linux/specific_linux.cpp
```diff
@@ -398,6 +398,11 @@ QString GetIconName() {
 	return Result;
 }
 
+bool IsGtkDarkTheme(const QString &theme) {
```
23rd, line 401:
```suggestion
[[nodiscard]] bool IsGtkDarkTheme(const QString &theme) {
```

Telegram/SourceFiles/platform/linux/specific_linux.cpp
```diff
@@ -398,6 +398,11 @@ QString GetIconName() {
...
+	const auto themeLower = theme.toLower();
+	return themeLower.contains(qsl("-dark-")) || themeLower.endsWith(qsl("-dark"));
```
23rd, line 403:
```suggestion
	return themeLower.contains(qsl("-dark-"))
		|| themeLower.endsWith(qsl("-dark"));
```

Telegram/SourceFiles/platform/linux/specific_linux.cpp
WhyNotHugo: > in private namespace

I'm not very familiar with C++ and I'm not sure about this bit.

Telegram/SourceFiles/platform/linux/specific_linux.cpp
```diff
@@ -474,9 +479,7 @@ std::optional<bool> IsDarkMode() {
 		const auto gtkThemeX = xSettings->setting("Net/ThemeName");
 		if (gtkThemeX.isValid()) {
 			result = false;
-			if (gtkThemeX.toString().contains(
-				qsl("-dark"),
-				Qt::CaseInsensitive)) {
+			if (IsGtkDarkTheme(gtkThemeX)) {
```
23rd, line 482:
```suggestion
			if (IsGtkDarkTheme(gtkThemeX.toString())) {
```
?

Telegram/SourceFiles/platform/linux/specific_linux.cpp
WhyNotHugo: Thanks. I just learnt something new! :)

23rd:
You can squash all commits to the first one. 👌

WhyNotHugo:
Squashed!

WhyNotHugo:
I'm curious, why this:

```c
			if (IsGtkDarkTheme(gtkThemePortalString)) {
				result = true;
				return result;
			}
```

Instead of:

```c
			if (IsGtkDarkTheme(gtkThemePortalString)) {
				return true;
			}
```

Also, `result = false` repeats a lot, but it's not necessary since it only ever changes to `true` before `return`ing. Would this make more sense?

```diff
-	std::optional<bool> result;
+	std::optional<bool> result = false;
```

ilya-fedin:
> I'm curious, why this

We're checking all the APIs and return as soon as at least one of them returns dark mode

> Also, `result = false` repeats a lot, but it's not necessary since it only ever changes to `true` before `return`ing. Would this make more sense?

The default value of `result` is `std::nullopt`, not `false`

john-preston:
@WhyNotHugo what's DCO?

john-preston:
@WhyNotHugo I see, Developer Certificate of Origin.

The point of the CLA is to give the submitted work to the public domain so that if I need to do any relicensing in the future I won't need to contact all the people who sent PRs here.

WhyNotHugo:
Hmm, upon further inspection, there's a lot of inconsistency here:

- The bot claims that one has to sign a CLA.
- [The agreement shown has a title "Developer Certificate of Origin"](https://cla-assistant.io/telegramdesktop/tdesktop?pullRequest=24133).
- The actual body of the agreement **is not** [the DCO](https://developercertificate.org/), but something very different that releases the contribution into public domain.

I had missed the discrepancy between title and body the previous time. Isn't it misleading to call it a DCO if the text is entirely different and has different goals?

I'm also a bit confused about the implications here: Telegram is distributed under the GPL, but if all developers have signed this agreement, doesn't that imply that it's fully been released into public domain?

> The point of the CLA is to give the submitted work to the public domain so that if I need to do any relicensing in the future I won't need to contact all the people who sent PRs here.

I'm familiar. Using a CLA in this style is quite common for open source projects that later want to turn into a non-open source or non-free project.

john-preston:
@WhyNotHugo Yes, for example I've maintained a closed source version for Telegram Support volunteers, that was later merged into the current open source version. But still I leave a possibility for something like that to happen again. Sorry, it is what it is.

There are no plans in closing the source code of any Telegram app (all the apps and the client side libraries for building other apps, like TDLib, are open source and this is not planned to be changed). But still I need the full control of the source code, sorry.

john-preston:
Perhaps the phrasing could be improved indeed.

ilya-fedin:
@WhyNotHugo if you don't agree with the CLA, maybe it's better to close the PR?

WhyNotHugo:
Oh, pretty sure I did accept it, I'll do that again.

WhyNotHugo:
> Perhaps the phrasing could be improved indeed.

Including the document in the repo too would probably be good enough, making it a lot less confusing/ambiguous what people agreed to. Especially since the ACTUAL body of the agreement is hosted externally and all references made by the bot are are not quite accurate.

WhyNotHugo:
Oh, can't sign right now since bot is down anyway. Will retry later.

image

WhyNotHugo:
Looks like the CLA bot now needs a few more permissions that are, regrettably, not at all reasonable (e.g.: read-write access to several private and public resources):

image

ilya-fedin:
So, will you sign?

vimpostor:
> I do still believe that having the document in-repo would be a simpler and more transparent approach.

I agree but I guess the advantage of the current approach is that Preston can see directly in the PR if someone signed the CLA. 

If you put the CLA just in the repo, you would also need to require every contributor to signoff their commit (similar to the Linux Kernel) with `Signed-off-by:` and you still wouldn't be completely sure that they agreed to the CLA. Also I am not sure if you can apply the DCO Signoff workflow at all to a CLA and still be legally sound.

WhyNotHugo:
Given 17fcc729a2474c6a18b6c4d5086b45d066c77e4a, this is no longer relevant. Thanks @ilya-fedin for implementing that! :beers:

PR #24168, opened 2022-03-10
Fix incorrect admin log

c0re100:
One more case of #23944: `Restricted` to `Left`

Before:
{8D9CEEC0-3B35-4E51-BD7E-CE7D90EF9D58}

After:
{4453F818-BC6C-4879-A5F0-94BE72751A1E}

### DIFF
--- a/Telegram/SourceFiles/history/admin_log/history_admin_log_item.cpp
+++ b/Telegram/SourceFiles/history/admin_log/history_admin_log_item.cpp
@@ -483,7 +483,10 @@ auto GenerateParticipantChangeText(
 				user,
 				ChatRestrictionsInfo(),
 				oldRestrictions);
-		} else if (oldParticipant && oldParticipant->type() == Type::Restricted && participant.type() == Type::Member) {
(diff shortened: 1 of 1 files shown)


Telegram/SourceFiles/history/admin_log/history_admin_log_item.cpp
```diff
@@ -483,7 +483,7 @@ auto GenerateParticipantChangeText(
 				user,
 				ChatRestrictionsInfo(),
 				oldRestrictions);
-		} else if (oldParticipant && oldParticipant->type() == Type::Restricted && participant.type() == Type::Member) {
+		} else if (oldParticipant && oldParticipant->type() == Type::Restricted && (participant.type() == Type::Member || participant.type() == Type::Left)) {
```
23rd, line 486:
```suggestion
		} else if (oldParticipant
			&& oldParticipant->type() == Type::Restricted
			&& (participant.type() == Type::Member
				|| participant.type() == Type::Left)) {
```

PR #24215, opened 2022-03-18
Added options to set UI scaling to 175% and 225%

Dreista:
Windows has 175% and 225% scaling options in display settings, and these options are commonly used (for example, 175% is commonly used by 4K, 27-inch display users).

If users install Telegram Desktop from Microsoft Store, they are unable to pass `-scale` argument to Telegram Desktop.


john-preston:
This should be possible in the new 4.5.5 beta version.

Dreista:
Tested on Telegram 4.5.7 beta version. Windows scaling is set to 175%. Telegram defaults to 200% scale and the actual step of the slider is 10% instead of 5%.

PR #24254, opened 2022-03-25
Allow to add more than 3 accounts on WIN/LINUX

ShodiyorK:
Like on mac/ios.


Aokromes:
such feature belongs to https://bugs.telegram.org/

PR #24275, opened 2022-04-01
Check and link `libatomic` for `RISC-V` architecture

SpriteOvO:
Currently this project cannot be compiled for `RISC-V` architecture, `ld` linker raises an error:

```
/usr/bin/ld: /tmp/ccOIMCdU.ltrans23.ltrans.o: undefined reference to symbol '__atomic_compare_exchange_1@@LIBATOMIC_1.0'
/usr/bin/ld: /usr/lib/libatomic.so.1: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

This is because the `RISC-V` architecture does not support some 1-byte and 2-byte lock-free atomic operations(e.g. `fetch_add`, `exchange`...).

Explicitly linking to `libatomic` would fix this problem. Inside that library, unsupported operations for 1-byte and 2-byte `std::atomic` used in the project will fallback to the mutex implementation, which is allowed by C++ standard.

### DIFF
--- a/Telegram/CMakeLists.txt
+++ b/Telegram/CMakeLists.txt
@@ -39,6 +39,7 @@ include(cmake/td_lang.cmake)
 include(cmake/td_scheme.cmake)
 include(cmake/td_ui.cmake)
 include(cmake/generate_appdata_changelog.cmake)
+include(cmake/check_atomic.cmake)
 
 if (WIN32)
     include(cmake/generate_midl.cmake)
@@ -1377,6 +1378,20 @@ else()
             desktop-app::lib_waylandshells
             desktop-app::external_kwayland
         )
+    endif()
+
+    if (NOT HAVE_CXX_ATOMICS_WITHOUT_LIB)
+        if (DESKTOP_APP_USE_PACKAGED)
+            target_link_libraries(Telegram
+            PRIVATE
+                atomic
+            )
+        else()
+            target_link_libraries(Telegram
+            PRIVATE
+                atomic-static
+            )
+        endif()
     endif()
 endif()
 
--- a/Telegram/cmake/check_atomic.cmake
+++ b/Telegram/cmake/check_atomic.cmake
@@ -1,0 +1,43 @@
+# Modified from https://github.com/llvm/llvm-project/blob/main/llvm/cmake/modules/CheckAtomic.cmake
+
+include(CheckCXXSourceCompiles)
+include(CheckLibraryExists)
+
+# Sometimes linking against libatomic is required for atomic ops, if
+# the platform doesn't support lock-free atomics.
+
+function(check_working_cxx_atomics varname)
+  set(OLD_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+  set(CMAKE_REQUIRED_FLAGS "${CMAKE_REQUIRED_FLAGS} -std=c++11")
+  CHECK_CXX_SOURCE_COMPILES("
+#include <atomic>
+std::atomic<int> x;
+std::atomic<short> y;
+std::atomic<char> z;
+std::atomic<uint64_t> v;
+int main() {
+  ++z;
+  ++y;
+  uint64_t i = v.load(std::memory_order_relaxed);
+  (void)i;
+  return ++x;
+}
+" ${varname})
+  set(CMAKE_REQUIRED_FLAGS ${OLD_CMAKE_REQUIRED_FLAGS})
+endfunction(check_working_cxx_atomics)
+
+# Check for (non-64-bit) atomic operations.
+if(MSVC)
+  set(HAVE_CXX_ATOMICS_WITHOUT_LIB True)
+elseif(LLVM_COMPILER_IS_GCC_COMPATIBLE OR CMAKE_CXX_COMPILER_ID MATCHES "XL")
+  # First check if atomics work without the library.
+  check_working_cxx_atomics(HAVE_CXX_ATOMICS_WITHOUT_LIB)
+  # If not, check if the library exists, and atomics work with it.
+  if(NOT HAVE_CXX_ATOMICS_WITHOUT_LIB)
+    list(APPEND CMAKE_REQUIRED_LIBRARIES "atomic")
+    check_working_cxx_atomics(HAVE_CXX_ATOMICS_WITH_LIB)
+    if (NOT HAVE_CXX_ATOMICS_WITH_LIB)
+      message(FATAL_ERROR "Host compiler must support std::atomic!")
+    endif()
+  endif()
+endif()


ilya-fedin:
1. Such kind of a change should go to the common_options target from cmake_helpers submodule, options_linux.cmake file
2. There should be a check for DESKTOP_APP_USE_PACKAGED and when it's false, the library should be statically linked
3. Isn't -pthread supposed to handle this?

SpriteOvO:
@ilya-fedin Thanks for your review.

> 1. Such kind of a change should go to the common_options target from cmake_helpers submodule, options_linux.cmake file

I will open a PR in that repo later.

> 2. There should be a check for DESKTOP_APP_USE_PACKAGED and when it's false, the library should be statically linked

So it should be changed to this, right?

```cmake
if (NOT HAVE_CXX_ATOMICS_WITHOUT_LIB)
    if (DESKTOP_APP_USE_PACKAGED)
        target_link_libraries(Telegram
        PRIVATE
            atomic
        )
    else()
        target_link_libraries(Telegram
        PRIVATE
            atomic-static
        )
    endif()
endif()
```

> 3. Isn't -pthread supposed to handle this?

It was, until `glibc` v2.33. See [Why glibc 2.34 removed libpthread](https://developers.redhat.com/articles/2021/12/17/why-glibc-234-removed-libpthread) and https://github.com/riscv-collab/riscv-gcc/issues/12#issuecomment-1007118732

XieJiSS:
Hi @ilya-fedin, 

IMO the nature of this problem is that atomics might need to link to libatomic, depending on atomic values' length and (of course) the target architecture. cmake allows you to explicitly link to libatomic, but that could be unnecessary on platforms like x86_64, which does not require libatomic to compile 1, 2, 4 and 8 byte(s) atomics. So, developers would need to check what types of atomic values they are using (e.g. bool, uint16, int, etc.) and compose them to a detect script to be executed by cmake.

As the types of atomic values developers are using often vary across projects, I think submitting this to cmake might not be a neat way to solve the problem.

ilya-fedin:
According to the link you provided, -pthread still links to libatomic when needed, it's just cmake doesn'tpass it even though THREADS_PREFER_PTHREAD_FLAG is set. So, in my opinion this is a cmake bug that should be fixed in cmake.

SpriteOvO:
@ilya-fedin That might be a `CMake` bug, but the point is that `pthread` implies that linking to `libatomic` is not documented (I've been Google it for half an hour) and we shouldn't rely on this behavior.

But the `gcc` documentation explicitly says that in this case we should link to `libatomic`:
> Current releases use the newer `__atomic` intrinsics, which are implemented by library calls if the hardware doesn't support them. Undefined references to functions like `__atomic_is_lock_free` should be resolved by linking to `libatomic`, which is usually installed alongside libstdc++. ([ref](https://gcc.gnu.org/onlinedocs/libstdc++/manual/ext_concurrency_impl.html))

ilya-fedin:
> and we shouldn't rely on this behavior

Well, I tend to disagree. It provides a very elegant way to handle this and I would prefer it until it would stop work.

> That might be a `CMake` bug

Report to cmake, please. If they would say this is not a cmake bug, we can add a manual logic to pass -pthread.

XieJiSS:
Hi @ilya-fedin, I've investigated the cmake pthread issue before, and I'd like to share my discoveries.

cmake does test for `-pthread` flag's necessity, and the test code cmake uses are limited to libpthread's function, e.g. `pthread_join`. However, with glibc >= 2.34, all libpthread's functions, like `pthread_create`, `pthread_join`, and even pthread mutex locks, do not require `-pthread` flag to compile. libpthread.a is an empty archive now, only left for compatibility sake.

Linking to libatomic as needed when specifying pthread (on riscv64) is a temporary decision of gcc, and might change in the future; I *guess* their intention of doing so is to hide the long-lasting lack of subword atomic asm code generation support.

This is quite *smart*: atomics are useful for multithreading applications, and people will provide -pthread when they are compiling codes that uses multithread. So, if gcc silently link to libatomic (`--as-needed`), subword atomics will work, and developers won't feel the difference of compiling atomic codes between riscv64 and x86_64. The only problem is that since glibc >= 2.34, pthread works without the -pthread flag. As `-pthread` starts to disappear from build scripts, the subword atomic issue becomes sensible to some developers.

cmake is detecting pthread correctly now and then, and we used to rely on an unstable coincidence and a hiding gcc *"feature"* to make things work. We can't expect cmake to test for `std::atomic` or gcc `__atomic ` builtins in their pthread detection code. Now that pthread codes compile directly, and `-pthread` disappears from CFLAGS/CXXFLAGS, I think that testing for libatomic functions and re-specify `-pthread` when libatomic's tests fail sounds kind of weird.

Another possible solution, despite the fix suggested by this PR, is to specify `--as-needed -latomic`, but we need to make sure the `--as-needed` will not affect other arguments despite `-latomic`.

XieJiSS:
https://github.com/Kitware/CMake/blob/1a46bc0d2b8e26eb96cb05d1fa4c97781ab503c1/Modules/FindThreads.cmake#L70

this is the pthread detection script of cmake. IMO adding `__atomic` stuffs into this script does not sound like matching the original intention of the detection.

ilya-fedin:
https://github.com/Kitware/CMake/commit/68285bc8a91586ae47315679212eaef737af6cc0

This is possible. Someone should just fix this patch to avoid the bug due what it was reverted.

XieJiSS:
> [Kitware/CMake@68285bc](https://github.com/Kitware/CMake/commit/68285bc8a91586ae47315679212eaef737af6cc0)
> 
> This is possible. Someone should just fix this patch to avoid the bug due to what it was reverted.

We have compiled and test 3.23.0, and this patch has been proved to be useless. It is still calling `_check_pthreads_flag`, which refuses to add `-pthread` for the reason I have pointed out before: pure pthread codes can now compile directly.

The position of invoking `_check_pthreads_flag` doesn't matter. The content of `_check_pthreads_flag` matters, but adding tests of libatomic into `_check_pthreads_flag` doesn't sound wise.

ilya-fedin:
> Another possible solution, despite the fix suggested by this PR, is to specify `--as-needed -latomic`, but we need to make sure the `--as-needed` will not affect other arguments other than `-latomic`.

I don't think always having libatomic as a dependency is a problem, it's way more sane than enumerating every use of std::atomic in a cmake test and change it on every new std::atomic use. But what if the system uses compiler-rt rather than libatomic?

> which refuses to add `-pthread` for the reason I have pointed out before: pure pthread codes can now compile directly.

Why though? I see it only checks that the compiler returns 0 exit code with -pthread compile argument. Nothing else.

ilya-fedin:
> We have compiled and tested cmake 3.22.0-rc5 and 3.23.0 (this patch landed on these branches)

I don't see 3.22.0-rc5, I see only 3.22.0-rc3. 3.23.0 already reverted the commit.

XieJiSS:
> Why though? I see it only checks that the compiler returns 0 exit code with -pthread compile argument. Nothing else.

the HEAD of CMake is `6e1be5db`, and:

https://github.com/Kitware/CMake/blob/6e1be5dbefab3e7317502e3d0fe4b132d0162ae5/Modules/FindThreads.cmake#L163

Here it tries to compile the test code without the -pthread flag. If it success, then:

```
      set(CMAKE_THREAD_LIBS_INIT "")
      set(CMAKE_HAVE_THREADS_LIBRARY 1)
      set(Threads_FOUND TRUE)
```

The code of adding `-pthread` is after this, in the `else` branch: 

```
172 | # Check for -pthread first if enabled. This is the recommended
      # way, but not backwards compatible as one must also pass -pthread
      # as compiler flag then.
```

So, if the code compiles directly, then the `else` branch won't be reached, hence no `-pthread` flag specified.

ilya-fedin:
> the HEAD of CMake is `6e1be5db`

The commit is already reverted.

ilya-fedin:
изображение

XieJiSS:
> > We have compiled and tested cmake 3.22.0-rc5 and 3.23.0 (this patch landed on these branches)
> 
> I don't see 3.22.0-rc5, I see only 3.22.0-rc3. 3.23.0 already reverted the commit.

Sorry, mistyping, should be 3.23.0-rc5.

I will double-check by directly checking out the specific commit and compiles cmake.

ilya-fedin:
3.23.0-rc5 has reverted commit, too

ilya-fedin:
3.23.0-rc2 should be good

ilya-fedin:
> So maybe we don't need to worry about compiler-rt?

I don't think libatomic can be linked when compiler-rt is used

> However... I still think linking to libatomic by `-pthread` is weird

And I still think it's the best solution and no change should be done in tdesktop.

r-value:
Hi @ilya-fedin,
If you check the doc of [gcc/libstdc++/using concurrency](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_concurrency.html), you will find:

> Some uses of `std::atomic` also require linking to `libatomic`.

In this case this exactly what happened. `libpthread` and `libatomic` are separate libraries, if you use `-pthread`, only threads are guaranteed to work, not atomics. `-pthread` causing links against `libatomic` is only a side effect of the fact that `libpthread` depends on `libatomic`. Now this relation has disappeared, exposing the problem that `std::atomic` are used without any explicit checks and links. Acoording to the document above, you do have to check if `libatomic` is needed if you used `std::atomic`.

ilya-fedin:
> Now this relation has disappeared, exposing the problem that `std::atomic` are used without any explicit checks and links.

I don't see it's disappeared, -pthread still links libatomic just fine

r-value:
> I don't see it's disappeared

The relationship seems remain because of the **undocumented** behavior of `gcc`. In glibc 2.34, `libpthread.a` is empty. How could an empty library have any dependency?

> -pthread still links libatomic just fine

We should avoid to rely on a feature which is undocumented, internal and not supposed to expose to downstream user. Essentially this problem has nothing to do with pthread, checking and linking against `libatomic` is the **documented correct way** to use `std::atomic`.

ilya-fedin:
> We should avoid to rely on a feature which is undocumented

I already heard that. My opinion didn't change.

mymedia2:
I proposed a similar solution, but it has since disappeared for some reason. https://github.com/telegramdesktop/tdesktop/pull/7080

SpriteOvO:
Close it as we are patching the spec of gcc now. (https://github.com/felixonmars/archriscv-packages/pull/1056#issuecomment-1140479064)

PR #24317, opened 2022-04-12
Allow usage of Arrow Up and Arrow Down to edit more than just the most recent message

gregordr:
Pressing up currently allows you to edit the most recent message. However, there is no way to edit older messages this way.

This PR enables navigating messages to edit them by pressing the up or down key when the cursor is at the start/end of a message that is being edited.

### DIFF
--- a/Telegram/SourceFiles/history/history.cpp
+++ b/Telegram/SourceFiles/history/history.cpp
@@ -2763,6 +2763,51 @@ HistoryItem *History::lastEditableMessage() const {
 	return nullptr;
 }
 
+
+HistoryItem *History::editableMessageBefore(MsgId current) const {
+	if (!loadedAtBottom()) {
+		return nullptr;
+	}
+
+	bool found = false;
+	const auto now = base::unixtime::now();
+	for (const auto &block : ranges::views::reverse(blocks)) {
+		for (const auto &message : ranges::views::reverse(block->messages)) {
+			const auto item = message->data();
+			if (item->allowsEdit(now)) {
+				if(owner().groups().findItemToEdit(item)->id == current) {
+					found = true;
+				} else if(found) {
+					return owner().groups().findItemToEdit(item);
+				}
+			}
+		}
+	}
+	return nullptr;
+}
+
+HistoryItem *History::editableMessageAfter(MsgId current) const {
+	if (!loadedAtBottom()) {
+		return nullptr;
+	}
+
+	HistoryItem * previous = nullptr;
+	const auto now = base::unixtime::now();
+	for (const auto &block : ranges::views::reverse(blocks)) {
+		for (const auto &message : ranges::views::reverse(block->messages)) {
+			const auto item = message->data();
+			if (item->allowsEdit(now)) {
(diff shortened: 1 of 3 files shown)


ilya-fedin:
How to scroll the chat with the arrow keys then? :thinking:

gregordr:
Yeah I see the issue. But I think both should be doable via keyboard. Maybe pgup and pgdown are already usable for scrolling chats?

The current way to scroll with arrow keys is already a bit awkward, as every time you scroll up you start editing the last message you sent. Separating the scrolling and editing to different keys would be good.

Alternatively, maybe we can use ctrl+shift or ctrl+alt to edit messages and have arrow keys just for scrolling? Any other suggestions?

Martinligabue:
I think it makes no sense to scroll the chat if you're editing a message, you already have the same problem now that if you're on multiple rows it changes behaviour based on if you're on the top of the message or bottom, or in the middle. Removing the movement of the chat via the keys that are supposed to edit the message seems the best and more coherent option

ilya-fedin:
The UX of tdesktop will be much worse for me if scrolling with arrow keys is removed

Martinligabue:
never said to remove it completely, just when editing a message, a momentary thing

ilya-fedin:
Message edit starts on arrow up, so you literally say to remove it

Martinligabue:
no, I mean to remove the shortcut that it's just an arrow up to edit, if I want to scroll I want to scroll, not to edit, and having the same button for different things it's confusing and it will always do both things, an option could be added, or just change the buttons to something different

ilya-fedin:
Still personally I like the current behavior and it would be a bad thing to me if it's changed

arx8x:
Binding up arrow to edit and scroll up at the same time should be considered bad UX. You might just want to scroll but you can't scroll without going to edit mode for the last message. When you're done scrolling and want to type a message, the input box is now an edit box for the last message. You have to press `ESC` to get back to message input. You shouldn't need to do these just to scroll a few pixels up. 
I don't use arrows keys to scroll but I don't support removing the feature. That would be a terrible idea. I know a lot of people who use it.
I have two suggestions for this.
1
Make the keybinding to edit on up arrow optional keeping the current behavior as default. This will allow arrow key scroll to be purely scrolling. Use a different keybinding like Ctrl + Shift + Up to edit messages. Keep pressing Up while holding Ctrl + Shift to edit older messages. The focus moves  up or down depending on the key you're pressing. 
2
Have a message highlight mode, activated by a keybinding like Ctrl + E. In this mode, the last message is highlighted by default. Up arrow highlights older messages. Shift + up/down selects messages. While a message is highlighted, press Enter to edit it , Left arrow to reply to the message, Right arrow to forward it, Ctrl + C to copy contents. Text input should be disabled until you choose to edit, reply, forward or press ESC so alphanum keys can be used for other actions on the highlighted messages. 
The second option is better but I feel like most of the userbase won't use these and will need a lot more effort to implement. The first option is fairly easy.

Uchiha8:
hello dude
LGTM

yasid-ai (approved):

ilya-fedin:
Conflicts

PR #24322, opened 2022-04-13
Add Right to Left capability to Telegram for languages like arabic, urdu and persian

AMP999:
this capability can support languages like:
1.Arabic
2.Persian/Farsi
3.Azeri
4.Aramaic
5.Dhivehi/Maldivian
6.Hebrew
7.Kurdish (Sorani)
8.Urdu

i'm working on this project. till now dialogs are almost fixed


AMP999:
> If the PR is still WIP, you can use the GitHub's draft feature

A complete fix of rtl will be a huge PR. so I decided to PR it when a hunk is ready.
fix of rtl won't interfere with existing experience as some parts of the telegram are already rtl compatible.

ilya-fedin:
So you want to say your PR is ready for review?

AMP999:
> So you want to say your PR is ready for review?

Yes

Telegram/SourceFiles/core/sandbox.cpp
```diff
@@ -182,6 +182,9 @@ int Sandbox::start() {
 		return 0;
 	}
 	_started = true;
+
+	style::SetRightToLeft(TRUE);
```
ilya-fedin, line 186: What's that?

ilya-fedin: I don't think you can enable RTL layout for every user...

AMP999: I have set that true to see how my changes appear in run time. 
It should'nt hard coded here. This function should bind to languages

ilya-fedin: Why there's code that shouldn't be here if it's ready for review? :thinking:

AMP999: You're right. I'll fix it and PR it again

PR #24398, opened 2022-04-19
Avoid a second query for the current color-scheme

WhyNotHugo:
Telegram listens for a signal that indicates when the color-scheme
changes. The signal itself includes the new value, but Telegram
currently queries for the value immediately after getting the signal.
This second round-trip is unnecessary, since the signal itself contains
the same information.

This changeset avoids this follow-up query, and drops the now-unused
`Setter`.

### DIFF
--- a/Telegram/SourceFiles/platform/linux/specific_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/specific_linux.cpp
@@ -382,12 +382,6 @@ QString GetIconName() {
 std::optional<bool> IsDarkMode() {
 #ifndef DESKTOP_APP_DISABLE_DBUS_INTEGRATION
 	[[maybe_unused]] static const auto Inited = [] {
-		static const auto Setter = [] {
-			crl::on_main([] {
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
Ugh... CI won't work because my username is uppercase. :facepalm: 

I'm sure I've seen this before and working around it was an ugly hack :(

Telegram/SourceFiles/platform/linux/specific_linux.cpp
```diff
@@ -396,7 +390,11 @@ std::optional<bool> IsDarkMode() {
 				const Glib::VariantBase &value) {
 				if (group == "org.freedesktop.appearance"
 					&& key == "color-scheme") {
-					Setter();
+					const auto ivalue = base::Platform::GlibVariantCast<uint>(value);
```
ilya-fedin, line 393: GlibVariantCast can throw an exception. You should use try-catch.

john-preston:
Thanks!

john-preston:

PR #24608, opened 2022-06-15
Fix Premium Maximum Connected Accounts number in Premium Preview Box

Sventra:
With one account connected

Before the edit:
image
After the edit:
image

This puts the message and limit on par with Android and iOS limits

### DIFF
--- a/Telegram/SourceFiles/boxes/premium_preview_box.cpp
+++ b/Telegram/SourceFiles/boxes/premium_preview_box.cpp
@@ -1816,9 +1816,7 @@ void DoubledLimitsPreviewBox(
 		});
 	}
 	const auto now = session->domain().accounts().size();
-	const auto till = (now + 1 >= Main::Domain::kPremiumMaxAccounts)
-		? QString::number(Main::Domain::kPremiumMaxAccounts)
-		: (QString::number(now + 1) + QChar('+'));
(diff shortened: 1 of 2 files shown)


john-preston:
@Sventra I want the max limit to be 6, not 4.

The correct fix would be to replace `QString::number(now + 1) + QChar('+')` with `QString::number(std::max(now, kMaxAccounts) + 1) + QChar('+')`.

Sventra:
Understandable with wanting 6 instead of 4 (strangely though on android the
max is 4, the more the better)
But may I ask the reason in wanting to display 5+ instead of straight up
the max number which is 6?

john-preston:
@Sventra because it doesn't work that way. You don't buy a premium and get 6 accounts instead of 3. You have 3 "free slots" of accounts, so you buy a premium and get one slot free and can use 4 accounts. Then you can buy another premium (on one of three non-premium accounts) and you can use 5 accounts now, 2 of them premium. And finally 6 accounts if 3 are premium subscribers.

On mobiles it is 4, because through the App Store / Play Market payment / subscription system you can buy only one subscription for your app (so make only one account premium and free only one "slot").

Sventra:
Oh, that makes much more sense now, thanks for the clarification on that
one, sorry for the PR 😅

PR #24639, opened 2022-06-19
Wayland protocols code generation for Qt5

mymedia2:


mymedia2:
See replacement #24923

PR #24663, opened 2022-06-21
Fix some typos

Molkree:


23rd:
I'm afraid this will have no effect since these strings are taken from the cloud.

Aokromes:
> > these strings are taken from the cloud
> 
> Ah, I was not aware of that. Is this open sourced anywhere? @23rd

https://translations.telegram.org/

Issue #24696, opened 2022-06-24
Build fails in armv7 and aarch64


catdevnull:
### Steps to reproduce

1. Try to build tdesktop in armv7 or aarch64

### Expected behaviour

tdesktop should build succesfully

### Actual behaviour

tdesktop fails to build in both [armv7](https://gitlab.alpinelinux.org/Nulo/aports/-/jobs/751333) and [aarch64](https://gitlab.alpinelinux.org/Nulo/aports/-/jobs/751332) because of an assumption that `char` is signed:

```
/builds/Nulo/aports/community/telegram-desktop/src/tdesktop-4.0.0-full/Telegram/SourceFiles/data/data_types.h:152:31: error: enumerator value '-2' is outside the range of underlying type 'char'
  152 |         FileDownloadFailed = -2,
      |                               ^
/builds/Nulo/aports/community/telegram-desktop/src/tdesktop-4.0.0-full/Telegram/SourceFiles/data/data_types.h:153:29: error: enumerator value '-1' is outside the range of underlying type 'char'
  153 |         FileUploadFailed = -1,
      |                             ^
```

### Operating system

Alpine Linux edge

### Version of Telegram Desktop

4.0.0

### Installation source

Other (unofficial) source

PR #24706, opened 2022-06-24
Explicitly specify `signed` modifier for `char` type

SpriteOvO:
Build for RISC-V architecture using v4.0 release will fail with errors:

```
/build/telegram-desktop/src/tdesktop-4.0.0-full/Telegram/SourceFiles/data/data_types.h:152:31: error: enumerator value ‘-2’ is outside the range of underlying type ‘char’
  152 |         FileDownloadFailed = -2,
      |                               ^
/build/telegram-desktop/src/tdesktop-4.0.0-full/Telegram/SourceFiles/data/data_types.h:153:29: error: enumerator value ‘-1’ is outside the range of underlying type ‘char’
  153 |         FileUploadFailed = -1,
      |                             ^
ninja: build stopped: subcommand failed.
```

Reference to C++ standard ([cppref](https://en.cppreference.com/w/cpp/language/types#Character_types)):

> `char` - type for character representation which can be most efficiently processed on the target system (has the same representation and alignment as either `signed char` or `unsigned char`, but is always a distinct type). [Multibyte characters strings](https://en.cppreference.com/w/cpp/string/multibyte) use this type to represent code units. For every value of type `unsigned char` in range [0, 255], converting the value to `char` and then back to `unsigned char` produces the original value. (since C++11) **The signedness of `char` depends on the compiler and the target platform**: the defaults for ARM and PowerPC are typically unsigned, the defaults for x86 and x64 are typically signed.

Reference to [RISC-V Calling Convention Specification](https://riscv.org/wp-content/uploads/2015/01/riscv-calling.pdf) [18.1] C Datatypes and Alignment:

> The C types char and unsigned char are 8-bit unsigned integers and are zero-extended when stored in a RISC-V integer register.

For RISC-V, the `char` type without a modifier is defined as `unsigned` by the implementation. Explicitly adding a `signed` modifier to them would fix these errors.

### DIFF
--- a/Telegram/SourceFiles/data/data_types.h
+++ b/Telegram/SourceFiles/data/data_types.h
@@ -148,7 +148,7 @@ enum LocationType {
 	SecureFileLocation = 0xcbc7ee28, // mtpc_inputSecureFileLocation
 };
 
-enum FileStatus : char {
+enum FileStatus : signed char {
 	FileDownloadFailed = -2,
 	FileUploadFailed = -1,
 	FileReady = 1,


ilya-fedin:
Can you add `Fixes #24696`?

SpriteOvO:
@ilya-fedin Sorry, I missed issue #24696, now it can only be closed manually by @john-preston.

john-preston:

PR #24771, opened 2022-07-08
money-purchasing and money-advertising updated

nintendoit:
Changed money-purchasing from 'none' to 'intense' as the in-app purchase of telegram premium is available.
Added money-advertising and set it to 'moderate' as telegram ads has been introduced.


john-preston:
What is `intense`? Why `intense`? Is there `non-intense-but-still`?

john-preston:
Ok

PR #24878, opened 2022-08-06
Force /utf-8 on MSVC

cnSchwarzer:


john-preston:
Why? Iirc there are some non-utf8 symbols in some cpp sources.

cnSchwarzer:
> Why? Iirc there are some non-utf8 symbols in some cpp sources.

The compilation fails if the codepage is 936.

john-preston:
It is forced in cmake_helpers now.

PR #25012, opened 2022-09-01
Correct FFmpeg API usage

alex65536:
FFmpeg docs say that you must return `AVERROR_EOF` from the read callback, not zero. Still, Telegram just propagates the return value from `IODevice::read()` call, which returns zero in case of EOF.

I don't know whether this commit has any effect on the upstream build, but it fixes a bug in Debian build of Telegram, which is using FFmpeg 5.1 instead of 4.4. Still, it's also useful in the upstream, as it makes work with FFmpeg more correct.

### DIFF
--- a/Telegram/SourceFiles/media/clip/media_clip_ffmpeg.cpp
+++ b/Telegram/SourceFiles/media/clip/media_clip_ffmpeg.cpp
@@ -283,7 +283,7 @@ bool FFMpegReaderImplementation::start(Mode mode, crl::time &positionMs) {
 		return false;
 	}
 	_ioBuffer = (uchar*)av_malloc(FFmpeg::kAVBlockSize);
-	_ioContext = avio_alloc_context(_ioBuffer, FFmpeg::kAVBlockSize, 0, static_cast<void*>(this), &FFMpegReaderImplementation::_read, nullptr, &FFMpegReaderImplementation::_seek);
+	_ioContext = avio_alloc_context(_ioBuffer, FFmpeg::kAVBlockSize, 0, static_cast<void*>(this), &FFMpegReaderImplementation::Read, nullptr, &FFMpegReaderImplementation::Seek);
 	_fmtContext = avformat_alloc_context();
 	if (!_fmtContext) {
 		LOG(("Gif Error: Unable to avformat_alloc_context %1").arg(logData()));
@@ -480,7 +480,7 @@ FFMpegReaderImplementation::PacketResult FFMpegReaderImplementation::readAndProc
 	return result;
 }
 
-int FFMpegReaderImplementation::_read(void *opaque, uint8_t *buf, int buf_size) {
+int FFMpegReaderImplementation::Read(void *opaque, uint8_t *buf, int buf_size) {
 	FFMpegReaderImplementation *l = reinterpret_cast<FFMpegReaderImplementation*>(opaque);
 	int ret = l->_device->read((char*)(buf), buf_size);
 	switch (ret) {
@@ -490,7 +490,7 @@ int FFMpegReaderImplementation::_read(void *opaque, uint8_t *buf, int buf_size) 
 	}
 }
 
-int64_t FFMpegReaderImplementation::_seek(void *opaque, int64_t offset, int whence) {
+int64_t FFMpegReaderImplementation::Seek(void *opaque, int64_t offset, int whence) {
 	FFMpegReaderImplementation *l = reinterpret_cast<FFMpegReaderImplementation*>(opaque);
 
 	switch (whence) {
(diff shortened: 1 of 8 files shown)


Telegram/SourceFiles/media/clip/media_clip_ffmpeg.cpp
```diff
@@ -482,7 +482,13 @@ FFMpegReaderImplementation::PacketResult FFMpegReaderImplementation::readAndProc
 
 int FFMpegReaderImplementation::_read(void *opaque, uint8_t *buf, int buf_size) {
 	FFMpegReaderImplementation *l = reinterpret_cast<FFMpegReaderImplementation*>(opaque);
-	return int(l->_device->read((char*)(buf), buf_size));
+	FFMpegReaderImplementation *l = reinterpret_cast<FFMpegReaderImplementation*>(opaque);
```
darkkeks, line 485: Seems like the line is mistakenly duplicated now
```suggestion
```

ilya-fedin:
> but it fixes a bug in Debian build of Telegram

Can you provide any details on the bug?

ilya-fedin:
> FFmpeg docs say

Can you point on that place in the docs? I don't see this documented at all.
изображение

john-preston:
@alex65536 Hi. Thanks for your PR!

I'd like to request a couple of things. First was reviewed already, it looks like a line is duplicated now.

Second - all other uses of avio_alloc_context should be fixed as well. Those include:
- Media::AbstractFFMpegLoader::_read_data (media/audio/media_audio_ffmpeg_loader.cpp)
- Media::AbstractFFMpegLoader::_read_bytes
- Media::AbstractFFMpegLoader::_read_file
- Media::Capture::Instance::Inner::Private::_read_data (media/audio/media_audio_capture.cpp)
- Media::Streaming::File::Context::Read (media/streaming/media_streaming_file.cpp)
- FFmpeg::FrameGenerator::Impl::Read (ffmpeg/ffmpeg_frame_generator.cpp)

ilya-fedin:
There's no need to split such simple changes to separate commits

alex65536:
> Can you provide any details on the bug?

The details are as follows. Each time when I view a specific animated sticker, the animation hangs at the very end, and the app starts consuming one CPU core until restart. If I view the same sticker three times, the app will consume three CPU cores :) That's because ffmpeg receives `0` instead of `AVERROR_EOF`, thus trying to read zero-sized buffer infinitely many times.

AFAIK the change involving `AVERROR_EOF` instead of `0` was added in FFmpeg 4.0, and the official builds seem to use 4.4. So, I don't really know why this is not reproducible on the official builds.

See also: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1018065. This report is not of very high quality and describes three regressions at once, and this PR definitely fixes one of them.

alex65536:
> There's no need to split such simple changes to separate commits

I just pushed "Commit suggestion" in the UI :) If you prefer to keep the commits from PRs and not to squash them during merge, I may merge these two commits as one manually.

ilya-fedin:
> Each time when I view a specific animated sticker, the animation hangs at the very end, and the app starts consuming one CPU core until restart. If I view the same sticker three times, the app will consume three CPU cores

This was reported by other users of third-party packages, but definitely is not reproducible with any official binaries, yeah... Thanks for the fix.

But as @john-preston pointed your fix is incomplete, unfortunately.

alex65536:
> But as @john-preston pointed your fix is incomplete, unfortunately.

I am going to fix the places he mentioned above soon, and then I hope this PR can be merged.

alex65536:
> FFmpeg::FrameGenerator::Impl::Read (ffmpeg/ffmpeg_frame_generator.cpp)

I could not find such file, but there is `FFmpeg::EmojiGenerator::Impl::_read` instead

alex65536:
Now all other uses are corrected.

john-preston:
@alex65536 Yeah, that's right. Thanks!

alex65536:
Is anything else required for the PR to be merged?

john-preston:
@alex65536 Please don't rename "Read / Seek" to "_read / _seek", the second one is legacy style that I unfortunately didn't remove from the codebase :( Either you can leave the method names as they are or you can rename the legacy style to the current one (_read_file to ReadFile, etc). After that I'll merge it.

john-preston:
@alex65536 Also I've pushed a lot of changes, now there are the methods I've mentioned in the first comment :( Sorry for that, can you please rebase your changes.

alex65536:
> Either you can leave the method names as they are or you can rename the legacy style to the current one (_read_file to ReadFile, etc).

I rebased the changes and converted all the methods to new code style.

john-preston:
Thanks!

Issue #24880, opened 2022-08-06
The problem of touch control of discussions in the publication.


Vldm11r:
### Steps to reproduce

Go to the comments of any post and try to scroll through them. 

### Expected behaviour

Scrolling through comments should be done by swiping with one finger, and a long hold should bring up the context menu.

### Actual behaviour

One touch instead of flipping comments - you select them, which is not convenient and not logical, and flipping works with only two fingers.

### Operating system

Windows 11 21H2 x64

### Version of Telegram Desktop

4.0.2 x64 (windows)

### Installation source

Static binary from official website

sleirsgoevy:
Confirmed.

yuyoyuppe:
same issue on surface pro x

PR #25079, opened 2022-09-17
Handle touch events in view replies section ScrollArea to allow scrolling from touch screen

yuyoyuppe:
This fixes https://github.com/telegramdesktop/tdesktop/issues/24880 which is reproducible on my MS Surface x Pro machine running Windows 10.

Current behavior:
- try scrolling with one finger while in the comments section
- observe that comments are selected instead

New behavior:
- scrolling with one finger works
- to select comments, do a long touch, then choose "Select"


john-preston:
Hi. Thanks for reminding. But this was not the way it was supposed to be.

Message history is supposed to have custom touch handling so that long-tap + move in text starts selecting this text, if went outside the text -- starts selecting the messages + shows the menu with actions on touch release. I didn't move this logic from main chat history widget to more modern channel comments / scheduled messages / pinned messages sections. I did now.

PR #25104, opened 2022-09-21
Allow users to filter message searches by channel sender in groups

### LINKED ISSUE #25105: It is impossible to search for messages posted in chats by masqueraded channels
It is possible in telegram chats to search for messages, and to filter those search results by a specific user.

However, if a user is posting as a channel, it is impossible to use that search filter, as channels are excluded from the list of possible targets when filtering by user.

I would like to request that search-by-member be expanded to support searching by channel entities, and in this case it should present messages posted as that specific channel (excluding the user's main account and any other channels they might post as).

It is important to have this feature from a chat administrative perspective, as without it, it can be difficult to find messages posted as a channel in anything but the recent past. It is also a nuisance to ordinary users who interact with masquerading channels and wish to review their conversation history.

Additional thoughts:
- If a user spams with a channel and then sets it to private, it shouldn't make the spam harder to find and eliminate
- If the group's settings are changed to disallow posting as channels but there are already such messages in the chat history, should users still be able to search with that filter? should admins be able to?

TheWug:
Users can send messages as channels in any supergroup which either has a public username or is linked to a channel. Previously it was not possible to filter by sender for messages sent on behalf of a channel, and now it is. In order to avoid confusion, you must search by the channel's exact username, and the group must permit sending messages as channels.

It is worth considering whether chat administrators should always be able to search for messages by any channel, regardless of group or channel settings, to ensure they can handle adversarial situations, but this pull request addresses only the simple case.

closes #25105

### DIFF
--- a/Telegram/SourceFiles/boxes/peers/add_participants_box.cpp
+++ b/Telegram/SourceFiles/boxes/peers/add_participants_box.cpp
@@ -1132,7 +1132,7 @@ void AddSpecialBoxSearchController::searchGlobalDone(
 		} else {
 			return;
 		}
-		for (const auto& mtpChat : list.v) {
+		for (const auto &mtpChat : list.v) {
 			const auto peerId = mtpChat.match([](const MTPDchannel& data) {
 					return peerFromChannel(data.vid().v);
 				}, [](const MTPDchat& data) {
--- a/Telegram/SourceFiles/boxes/peers/add_participants_box.cpp
+++ b/Telegram/SourceFiles/boxes/peers/add_participants_box.cpp
@@ -415,8 +415,13 @@ std::unique_ptr<PeerListRow> AddSpecialBoxController::createSearchRow(
 	if (_excludeSelf && peer->isSelf()) {
 		return nullptr;
 	}
-	if (const auto user = peer->asUser()) {
-		return createRow(user);
+	if (_excludeBroadcasts && peer->isBroadcast())
+	{
+		return nullptr;
+	}
+
+	if (peer->asUser() || peer->asBroadcast()) {
+		return createRow(peer);
 	}
 	return nullptr;
 }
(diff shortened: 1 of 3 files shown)


Telegram/SourceFiles/boxes/peers/add_participants_box.cpp
```diff
@@ -1118,9 +1123,48 @@ void AddSpecialBoxSearchController::searchGlobalDone(
...
+	const auto feedListChannels = [&](const MTPVector<MTPChat>& list) {
+		// skip this step in groups which are not eligible for masquerading.
+		if (const auto megagroup = _peer->asMegagroup()) {
+			if (!megagroup->hasUsername() && !megagroup->linkedChat()) {
+				return;
+			}
+		} else {
+			return;
+		}
+		for (const auto& mtpChat : list.v) {
```
23rd, line 1136:
```suggestion
		for (const auto &mtpChat : list.v) {
```

Telegram/SourceFiles/boxes/peers/add_participants_box.cpp
```diff
@@ -1118,9 +1123,48 @@ void AddSpecialBoxSearchController::searchGlobalDone(
...
+			const auto peerId = mtpChat.match([](const MTPDchannel& data) {
+					return peerFromChannel(data.vid().v);
+				}, [](const MTPDchat& data) {
+					return peerFromChat(data.vid().v);
+				}, [](auto&&) {
+					return PeerId();
+				});
+			if (const auto peer = _peer->owner().peerLoaded(peerId)) {
+				bool isBroadcastMasqueradable = false;
```
23rd, line 1145:
```suggestion
				auto isBroadcastMasqueradable = false;
```

Telegram/SourceFiles/boxes/peers/add_participants_box.cpp
```diff
@@ -1118,9 +1123,48 @@ void AddSpecialBoxSearchController::searchGlobalDone(
...
+		// skip this step in groups which are not eligible for masquerading.
+		if (const auto megagroup = _peer->asMegagroup()) {
+			if (!megagroup->hasUsername() && !megagroup->linkedChat()) {
+				return;
+			}
+		} else {
+			return;
+		}
+		for (const auto& mtpChat : list.v) {
+			const auto peerId = mtpChat.match([](const MTPDchannel& data) {
```
23rd, line 1137:
```suggestion
			const auto peerId = mtpChat.match([](const MTPDchannel &data) {
```

Telegram/SourceFiles/boxes/peers/add_participants_box.cpp
```diff
@@ -1118,9 +1123,48 @@ void AddSpecialBoxSearchController::searchGlobalDone(
...
+					return peerFromChannel(data.vid().v);
+				}, [](const MTPDchat& data) {
```
23rd, line 1139:
```suggestion
				}, [](const MTPDchat &data) {
```

ilya-fedin:
CLA wasn't signed in 3 years

PR #25121, opened 2022-09-23
pt-BR translation to the Quit Telegram option

DanGLVK:
Having this option be in english only makes it conflict with the current set language on a user's computer, i'd suggest making translations for this option encouraged 

If this isn't the correct place for this then feel free to close this pull request

### DIFF
--- a/lib/xdg/telegramdesktop.desktop
+++ b/lib/xdg/telegramdesktop.desktop
@@ -18,4 +18,5 @@ X-GNOME-SingleWindow=true
 [Desktop Action Quit]
 Exec=telegram-desktop -quit
 Name=Quit Telegram
+Name[pt_BR]=Fechar Telegram
 Icon=application-exit


fitojb:
Thank you for finding the root of the problem. I’ve noticed it too; after your patch is merged, I’ll send one to add Spanish, Catalan and Asturian.

ilya-fedin:
@fitojb there's no plan to merge this AFAIIK. All the user-visible strings must come from the Telegram Translation Platform. To properly fix this, one has to write a generator that would get the strings from the transltation platform and write them to the .desktop file.

ilya-fedin:
A way easier solution would be to use some runtime API, so the application itself has to send already localized string to the system. But unfortunately there's no such API AFAIK (only the deprecated Unity Launcher API, no one implements it now) :(. It's an idea for you guys to ask for that API at XDG, GNOME or somewhere else if you have interest in this to be solved. Otherwise it's likely to stay like this forever since it's non-trivial to write such a generator given that the translation is provided via the regular MTProto Telegram API, I don't imagine anyone who wants to work on that (currently that seem to be only you guys) would have the related experience as well as I don't imagine anyone who have the related experience would like to work on that.

EmeraldCity3 (approved):

PR #25153, opened 2022-10-01
Mention docker-buildx requirement in build instructions

aequabit:
If you don't know that the buildx command is part of a plugin which isn't included on some distributions, the error it throws at you can be quite confusing:

```
$ poetry run gen_dockerfile | docker buildx build -t tdesktop:centos_env -
unknown shorthand flag: 't' in -t
See 'docker --help'.
```

Might save people some time.

### DIFF
--- a/docs/building-linux.md
+++ b/docs/building-linux.md
@@ -10,7 +10,7 @@ You will require **api_id** and **api_hash** to access the Telegram API servers.
 
 ### Prepare libraries
 
-Install [poetry](https://python-poetry.org), go to the `tdesktop/Telegram/build/docker/centos_env` directory and run
+Install [poetry](https://python-poetry.org), go to the `tdesktop/Telegram/build/docker/centos_env` directory and run (make sure the docker-buildx plugin is installed)
 
     poetry install
     poetry run gen_dockerfile | docker buildx build -t tdesktop:centos_env -


ilya-fedin:
Hmm, it works without any plugin for me, out-of-the-box

ilya-fedin:
Maybe it's better to suggest to install Docker from the official site (https://docs.docker.com/get-docker/) to ensure it's fully installed?

aequabit:
> Hmm, it works without any plugin for me, out-of-the-box

Out of the box, yes. Without any plugin, no. 


> Maybe it's better to suggest to install Docker from the official site (https://docs.docker.com/get-docker/) to ensure it's fully installed?

I don't see how Docker Desktop not being fully installed relates to buildx not being packaged with the base Docker binaries on some distributions. Docker Desktop just happens to include the plugin, but it also includes lots of things you don't need. Also, buildx is available in my package repositories, it just needs to be installed separately.

ilya-fedin:
I believe it's better to recommend to install Docker Desktop

ilya-fedin:
After the poetry commands add `If you have any problem running this command, make sure you're using the [official Docker build](https://docs.docker.com/get-docker/).`

ilya-fedin:
This would prevent any problems due to specific packaging

ilya-fedin:
> the problem is that you use a component that's not part of Docker

I'm not sure this is true, I have only docker installed it works out-of-the-box, without installing various plugins separately. I have a suspicious that various distributions separate docker to multiple packages just like they usually do with any software, so it's better just suggest installing the official package that has everything needed for sure.

john-preston:
It should not be needed anymore.

PR #25154, opened 2022-10-02
fix clang15/libc++15 build

pg83:
libc++ in newer llvm requre completely defined ScanListData


ilya-fedin:
We believe it's a bug in compiler, it's reported to gcc as https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106608, you can create a report for llvm I guess.

PR #25174, opened 2022-10-06
;)

Sestr0:
I deleted all emojis because they are fricking cringe for god's sake.


Sestr0:
good job!

Sestr0:
> good job!

Thanks!

23rd:
This removes the built-in emoji set, but does not affect those sets that can be downloaded from the settings.

PR #25288, opened 2022-11-07
fixes #6410

### LINKED ISSUE #6410: [Bug] - See who is online straight from the chat list, not disappear until there is mouse movement
The connected user does not disappear until there is mouse movement above the user (I do not know if it only happens with users with thumbtacks)
Fallo 
Linux Mint 19.2 64bit XFCE

Optiligence:
the animation callback didn’t work before

### DIFF
--- a/Telegram/SourceFiles/dialogs/dialogs_inner_widget.cpp
+++ b/Telegram/SourceFiles/dialogs/dialogs_inner_widget.cpp
@@ -3338,11 +3338,6 @@ void InnerWidget::groupHasCallUpdated(not_null<PeerData*> peer) {
 }
 
 void InnerWidget::updateRowCornerStatusShown(not_null<History*> history) {
-	const auto repaint = [=] {
-		repaintDialogRowCornerStatus(history);
-	};
-	repaint();
-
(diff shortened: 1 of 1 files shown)


john-preston:
The animation shouldn't play if the row isn't visible on screen, it should switch the dot instantly.

I hope this issue is already fixed in 4.3 version.

Optiligence:
i can add the visible toggle back; but i’m not a fan of seeing logic like this, Qt should be able to handle that

but i thought it’s not a good idea if the user scrolls it into view later on (and i wasn’t sure if it’s updated from somewhere else)
the 2 other places that call `updateCornerBadgeShown` don’t provide a proper callback, so they – as is – can’t start an animation

Optiligence:
the issue persists with 4.3.0

Optiligence:
also: 
>it should switch the dot instantly.

i don’t think that’s what the code does though
since the issue wasn’t that only the animation was missing – the dot remained…
(except if there is a repaint when comes into view)

john-preston:
It should be fixed in the next version 4.3.1, I hope to release it today.

PR #25305, opened 2022-11-08
Add an experimental "small message radius" toggle

DragoonAethis:
This toggle allows switching to the pre-4.3.0, smaller message bubble radius after an app restart. The message bubble radius styles now have to be referenced via the `Ui::BubbleRadius*` and `Ui::MsgFileThumbRadius*` wrappers to use the appropriate value.

### DIFF
--- a/Telegram/SourceFiles/ui/chat/chat_style_radius.cpp
+++ b/Telegram/SourceFiles/ui/chat/chat_style_radius.cpp
@@ -20,7 +20,7 @@ namespace {
 base::options::toggle UseSmallMsgBubbleRadius({
 	.id = kOptionUseSmallMsgBubbleRadius,
 	.name = "Use small message bubble radius",
-	.description = "Makes most message bubbles square-ish again.",
+	.description = "Makes most message bubbles square-ish.",
 	.restartRequired = true,
 });
 
--- a/Telegram/SourceFiles/ui/chat/chat_style_radius.cpp
+++ b/Telegram/SourceFiles/ui/chat/chat_style_radius.cpp
@@ -1,0 +1,61 @@
+/*
+This file is part of Telegram Desktop,
+the official desktop application for the Telegram messaging service.
+
+For license and copyright information please follow this link:
+https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
+*/
+#include "ui/chat/chat_style_radius.h"
+#include "ui/chat/chat_style.h"
+#include "base/options.h"
+
+#include "ui/chat/chat_theme.h"
+#include "ui/painter.h"
+#include "ui/ui_utility.h"
+#include "styles/style_chat.h"
+
+namespace Ui {
+namespace {
+
+base::options::toggle UseSmallMsgBubbleRadius({
+	.id = kOptionUseSmallMsgBubbleRadius,
+	.name = "Use small message bubble radius",
+	.description = "Makes most message bubbles square-ish again.",
+	.restartRequired = true,
+});
+
+} // namespace
+
+const char kOptionUseSmallMsgBubbleRadius[] = "use-small-msg-bubble-radius";
+
+int BubbleRadiusSmall() {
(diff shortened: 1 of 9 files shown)


DragoonAethis:
The toggle is disabled by default, of course. Here's how it looks:

image

bb010g:
How hard would it be to turn this into a message radius slider? E.g.
```cpp
int BubbleRadiusSmall() {
	return std::min(UseMsgBubbleRadius.value(), st::bubbleRadiusSmall);
}

int BubbleRadiusLarge() {
	return UseMsgBubbleRadius.value();
}
```

DragoonAethis:
> Why 'experimental features'?

Because it's a rather minimal hack that overrides just 2 values from the base app styles - it's likely some elements will look slightly incorrect with this enabled, and I don't want this addition to be a large maintenance burden. In Experimental Features it's okay for features to be just broken, so if fixing some oddity takes too much code, it can be safely left as is. Of course, if devs want to turn this into a properly-supported feature, this toggle can be easily moved later.

> How hard would it be to turn this into a message radius slider?

The app does a lot of manual painting and caching pixmaps that depends on these style values, so doing it live is probably non-trivial. Just having a slider doesn't sound too bad, but then you'd want a preview window to see how it looks, and that's a lot of extra code - devs in the main thread mentioned they'll ask if this is something they want to support for real, and in the meantime I want this initial version to be as simple as possible.

john-preston:
LGTM. Just remove the " again" in the description, it's not "again" for any new user of the app.

DragoonAethis:
macOS build failed on CI with:

```
2022-11-08T09:41:36.0384810Z CodeSign /Users/runner/work/tdesktop/tdesktop/tdesktop/out/Debug/Telegram.app (in target 'Telegram' from project 'Telegram')
2022-11-08T09:41:36.0385290Z     cd /Users/runner/work/tdesktop/tdesktop/tdesktop
2022-11-08T09:41:36.0385510Z     
2022-11-08T09:41:36.0385770Z     Signing Identity:     "-"
2022-11-08T09:41:36.0385980Z     
2022-11-08T09:41:36.0387410Z     /usr/bin/codesign --force --sign - --deep -o runtime --entitlements /Users/runner/work/tdesktop/tdesktop/tdesktop/out/Telegram/Telegram.build/Debug/Telegram.build/Telegram.app.xcent --timestamp\=none --generate-entitlement-der /Users/runner/work/tdesktop/tdesktop/tdesktop/out/Debug/Telegram.app
2022-11-08T09:41:36.7258010Z /Users/runner/work/tdesktop/tdesktop/tdesktop/out/Debug/Telegram.app: internal error in Code Signing subsystem
2022-11-08T09:41:36.7259320Z Command CodeSign failed with a nonzero exit code
```

Code signer doing code signer things, I guess? The same failure happens on other PRs as well.

Telegram/SourceFiles/ui/chat/chat_style_radius.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "ui/ui_utility.h"
+#include "styles/style_chat.h"
+
+namespace Ui {
+namespace {
+
+base::options::toggle UseSmallMsgBubbleRadius({
+	.id = kOptionUseSmallMsgBubbleRadius,
+	.name = "Use small message bubble radius",
+	.description = "Makes most message bubbles square-ish again.",
```
john-preston, line 23:
```suggestion
	.description = "Makes most message bubbles square-ish.",
```

john-preston (changes requested):
Looks fine, just change the phrase a bit please.

DragoonAethis:
I've committed this from GitHub, but can't amend commit/squash my branch from its UI - if you can merge w/ squash, go ahead. If not, I'll force push a squashed change after I'm back from work in a few hours.

john-preston:
Ok, thanks.

PR #25454, opened 2022-11-26
Have full path to the binary in .desktop file

ilya-fedin:
Depends on https://github.com/desktop-app/lib_base/pull/144


xvitaly:
Hardcoding paths is a very bad practice.

ilya-fedin:
@xvitaly hardcoding?

xvitaly:
> @xvitaly hardcoding?

`/usr/bin/telegram-desktop` is hardcoded path. `telegram-desktop` is always preferred.

ilya-fedin:
I don't know for who it's preferred, I heard an opinion it's a security problem to not to have full path to the binary in .desktop file. Anyway, this is not something I care about, feel free to submit a better PR to avoid creation of the URL scheme handler file that won't break official binary (just #ifndef'ing related code is not better of course).

Issue #24854, opened 2022-08-01
[Feature Request] Filter out messages from blocked users


github12101:
### Is your feature request related to a problem?

Problem:
If one decide to hide messages from one person, one can use Block function.
This will disable personal messages with that person.

However, all groups, channel comment chats are not affected. Everything is as normal and you can have conversations with blocked person like nothing happened.

### Describe the solution you'd like

Extend Block functionality so it hides all messages from blocked person in group and channel chats.

### Describe alternatives you've considered

There is no alternative. Currently, this feature is missing. Blocking person means personal messages don't work, nothing more. Group chats are working as usual and you can have conversation with blocked person all day.

Aokromes:
such feature belongs to https://bugs.telegram.org/

PR #25492, opened 2022-12-03
prevent a message from registering into group, if message comes from a blocked peer

ahxxm:
related issue #24854, and a long-time [request](https://bugs.telegram.org/c/37) in the forum

This is a preliminary attempt to "ignore blocked users in a group"

I'm very open to the following discussions, although I do want this feature:
- is it ok to assume that the user doesn't want to see the message?
- does the change interfere with group message update mechanism?(haven't read into this part yet, I can imagine the client tries to load since and consume more resources for both client and server)
- constraints violation regarding storage?
- is this the correct place?

(It seems to me that the message item is inserted into storage, then the client will try to register it, I'm not quite sure)

### DIFF
--- a/Telegram/SourceFiles/data/data_groups.cpp
+++ b/Telegram/SourceFiles/data/data_groups.cpp
@@ -47,7 +47,7 @@ void Groups::registerMessage(not_null<HistoryItem*> item) {
 	}
 	const auto i = _groups.emplace(item->groupId(), Group()).first;
 	auto &items = i->second.items;
-	if (items.size() < kMaxItemsInGroup) {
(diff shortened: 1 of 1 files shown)


john-preston:
This is not supposed to work that way, at least in any official Telegram application. Maybe in some fork (if it doesn't conflict with the platform rules, one should check first).

ahxxm:
thanks for the quick response.
Does the official Telegram application plan to support "ignoring messages from block persons" in the near future? What platform rules can I refer to?

john-preston:
@ahxxm As far as I know this feature is not planned.

ahxxm:
@john-preston Thanks! I'll try to build a mental model to filter these messages if they are unavoidable

PR #25512, opened 2022-12-06
Fix building without spellcheck

mid-kid:
Port from gentoo: https://github.com/gentoo/gentoo/blob/f62d7c6599bbb275f7863bd1fef0f5d31a673703/net-im/telegram-desktop/files/tdesktop-4.3.4-fix-disabling-spellcheck.patch

/var/tmp/portage/net-im/telegram-desktop-4.3.4/work/tdesktop-4.3.4-full/Telegram/SourceFiles/boxes/translate_box.cpp:18:10: fatal error: spellcheck/platform/platform_language.h: No such file or directory
   18 | #include "spellcheck/platform/platform_language.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.

### DIFF
--- a/Telegram/SourceFiles/boxes/translate_box.cpp
+++ b/Telegram/SourceFiles/boxes/translate_box.cpp
@@ -15,7 +15,9 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 #include "main/main_session.h"
 #include "mtproto/sender.h"
 #include "settings/settings_common.h"
+#ifndef TDESKTOP_DISABLE_SPELLCHECK
 #include "spellcheck/platform/platform_language.h"
+#endif
 #include "ui/effects/loading_element.h"
 #include "ui/layers/generic_box.h"
 #include "ui/widgets/buttons.h"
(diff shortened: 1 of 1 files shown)


ilya-fedin:
Just want to mention one thing from the original patch:

> Regular builds would always return false here, anyway, as linux isn't a
priority with these things.

That's false, Linux, just like Windows, uses CLD3 library for this.

ilya-fedin:
It seems a similar mis-statement is done in tdesktop-4.3.4-qt5-incompatibility-1.patch that removes functionality which could be enabled by a visible toggle in Experimental settings.

ilya-fedin:
And I wonder whether simple `.arg(md5Hash.constData())` would be enough instead of tdesktop-4.3.4-qt5-incompatibility-2.patch

mid-kid:
I based my observation regarding the language recognition feature on this line: <https://github.com/desktop-app/lib_spellcheck/blob/b095b4d/spellcheck/platform/linux/language_linux.cpp#L12>

I could be wrong but it sure seems like language recognition isn't implemented here.

ilya-fedin:
language_linux.cpp is not actually used by lib_spellcheck's CMakeLists.txt

mid-kid:
> And I wonder whether simple .arg(md5Hash.constData()) would be enough instead of tdesktop-4.3.4-qt5-incompatibility-2.patch

Part of the reason I've pointed that patch out in the bug tracker rather than submitting it as a PR, I have never written any C++ let alone in conjunction with Qt, so I have no idea how to interpret its error messages. Thanks for the suggestion.

ilya-fedin:
If you will confirm it works, I'll submit a PR fixing both errors

PR #25527, opened 2022-12-08
Use std::invoke_result_t not ::result_of to fix C++20 tdesktop 4.3.4

klemensn:
tdesktop defaults to (probably requires) C++20, but `std::result_of`[0]
was deprecated in C++17 and removed in C++20.

0: https://en.cppreference.com/w/cpp/types/result_of

### DIFF
--- a/Telegram/SourceFiles/platform/linux/specific_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/specific_linux.cpp
@@ -356,9 +356,10 @@ void LaunchGApplication() {
 
 			using Window::Notifications::Manager;
 			using NotificationId = Manager::NotificationId;
-			using NotificationIdTuple = std::result_of<
-				decltype(&NotificationId::toTuple)(NotificationId*)
-			>::type;
+			using NotificationIdTuple = std::invoke_result_t<
+				decltype(&NotificationId::toTuple),
+				NotificationId*
+			>;
 
 			const auto notificationIdVariantType = [] {
 				try {


Telegram/SourceFiles/platform/linux/specific_linux.cpp
```diff
@@ -356,7 +356,7 @@ void LaunchGApplication() {
 
 			using Window::Notifications::Manager;
 			using NotificationId = Manager::NotificationId;
-			using NotificationIdTuple = std::result_of<
+			using NotificationIdTuple = ranges::result_of<
 				decltype(&NotificationId::toTuple)(NotificationId*)
 			>::type;
```
ilya-fedin, line 361: Try
```suggestion
			using NotificationIdTuple = std::invoke_result_t<
				decltype(&NotificationId::toTuple),
				NotificationId*
			>;
```

klemensn: Thanks, building now.

klemensn: Updating the PR now so CI can also test this.
Surely a better fix than using the old interface through ranges.

Telegram/SourceFiles/platform/linux/specific_linux.cpp
```diff
@@ -356,9 +356,10 @@ void LaunchGApplication() {
 
 			using Window::Notifications::Manager;
 			using NotificationId = Manager::NotificationId;
-			using NotificationIdTuple = std::result_of<
-				decltype(&NotificationId::toTuple)(NotificationId*)
-			>::type;
+			using NotificationIdTuple = ranges::invoke_result_t<
```
ilya-fedin, line 359: ranges? It should be std

klemensn: Meh, my build ran with the right patch, I messed up updating the PR by hand.

Telegram/SourceFiles/platform/linux/specific_linux.cpp
klemensn: OpenBSD/amd64 with clang 13 completed a tdesktop 4.3.4 build with this PR.

Issue #25582, opened 2022-12-18
TDesktop is only telegram client which does not support 2560px side images. Is downloading them as 2560px, then down scaling them to 1280px before displaying.


Lesiuk:
### Steps to reproduce

1. Send image with 2560px side to yourself.
2. Open image and save to disk.
3. Compare them.

### Expected behaviour

We are saving to disk 2560px image like we should but before displaying image is converted to 1280px.
MacOSX, iPhone, Android clients are not converting images to 1280px before displaying them. Only tdesktop does that.

Code responsible for this behavior is inside function below. SideLimit is 1280, image is 2560x1900 image, and we are scaling it to 1280px.

```
void PhotoMedia::set(
		PhotoSize size,
		PhotoSize goodFor,
		QImage image,
		QByteArray bytes) {
	const auto index = PhotoSizeIndex(size);
	**const auto limit = PhotoData::SideLimit();
	if (image.width() > limit || image.height() > limit) {
		image = image.scaled(
			limit,
			limit,
			Qt::KeepAspectRatio,
			Qt::SmoothTransformation);
	}**
	_images[index] = PhotoImage{
		.data = std::make_unique<Image>(std::move(image)),
		.bytes = std::move(bytes),
		.goodFor = goodFor,
	};
	_owner->session().notifyDownloaderTaskFinished();
}
```

### Actual behaviour

Display downloaded images in their native resolutions. Do not downscale them right after downloading.

### Operating system

Windows 10

### Version of Telegram Desktop

4.4.1

### Installation source

Static binary from official website

Lesiuk:
TelegramSwift:
<img width="1370" alt="image" src="https://user-images.githubusercontent.com/543264/208315949-82b1d6d9-bb48-407a-8723-057a5f2a19bf.png">

Lesiuk:
tdesktop:
![image](https://user-images.githubusercontent.com/543264/208315982-c9ea90a4-2d13-4d02-aa54-6d6ef97c8daa.png)

ilya-fedin:
#6520, #2206
although I don't believe the download happens in 2560px as the whole reason tdesktop uses 1280px variant is to save the loading time

Lesiuk:
> #6520, #2206 although I don't believe the download happens in 2560px as the whole reason tdesktop uses 1280px variant is to save the loading time

I have my breakpoint inside PhotoMedia::set. TDesktop is downloading full 2560px image and then downscaling.

[...]

Issue #6520, opened 2019-09-10
Increase the photo size limit of 1280 pixels


mrogin-technic:
telegramdesktop uses an image size limit of 1280x1280 pixels although the Telegram API allows a photo sizes up to 2560 pixels. Photos sent with telegramdesktop therefore don't look good on current tablets and smartphones. Therefore, it should be possible to increase the limit from 1280 to 1920 or 2560 px using manual settings. A setting for the strength of the JPEG compression would also be very useful.
Currently the fixed downscaling limit and JPEG compression is hardcoded in: /SourceFiles/storage/localimageloader.cpp

jeffythedragonslayer:
These should be declared constexpr at the top of the file.  I will try to open a pull request once I get my build working.

mrg21:
If it should be configurable for each sending, then it would have to be integrated into this window.
https://i.vgy.me/GkxWSJ.png
With an additional line to selection the maximum size (1280/1920/2560) and JPG compression (LQ/MQ/HQ).
Otherwise simply change the default globally in the settings for all image files instead of the fixed values 1280 and 87 for JPG compression.
Here is the current source code:
```
else if (_type != SendMediaType::File) {
...
auto full = (w > 1280 || h > 1280) ? fullimage.scaled(1280, 1280, Qt::KeepAspectRatio,...
...
full.save(&buffer, "JPG", 87);
```
https://core.telegram.org/constructor/photoSize

mrogin-technic:
This enhancement should also fix the following bug https://github.com/telegramdesktop/tdesktop/issues/2206

mrogin-technic:
After more than one year, this issue is still open. Are there no plans to solve the problem of the poor image quality when viewing and sending photos in Telegram Desktop?

SmartManoj:
Does "easy pick" means "good first issue"?

ilya-fedin:
> Does the "easy pick" mean the "good first issue"?

No, theoretically it means that this is easy to fix for third-party contributors.

But as far as I know, this issue is definitely not an "easy pick", since core developers want to fix this only in a way that wouldn't increase traffic consumption without opening the image in fullscreen. In other words, fixing this issue requires a rewrite of how tdesktop downloads photos, to separate in-chat downloading and media view downloading, what also means that new UI elements are needed to show the download progress in media view.

Remu-rin:
Still relevant.

mrg21:
[...]

PR #25584, opened 2022-12-19
removed downscaling 2560px images before displaying them - #2206

### LINKED ISSUE #2206: Received photos are smaller than in Telegram Web
I have a bot that sends photos to users. Those photos are sent as regular photos, **not** files.

The photos have a resolution of 1600x1200. Telegram Web shows the photos at the original size, but Telegram Desktop receives them at a resolution of 1280x960, with a very very ugly rescaling method.

Lesiuk:
I tested it for, a day, and it looks good enough for me.  I honestly don't see the reason why we would like to downscale them here. 

If we receive 2560px image, we display it as 2560px image. Images sent from tdesktop are still downscaled to 1280px. This behavior is consistent with other telegram clients.

Fixes #2206

### DIFF
--- a/Telegram/SourceFiles/data/data_photo_media.cpp
+++ b/Telegram/SourceFiles/data/data_photo_media.cpp
@@ -101,14 +101,6 @@ void PhotoMedia::set(
 		QImage image,
 		QByteArray bytes) {
 	const auto index = PhotoSizeIndex(size);
-	const auto limit = PhotoData::SideLimit();
-	if (image.width() > limit || image.height() > limit) {
-		image = image.scaled(
-			limit,
-			limit,
-			Qt::KeepAspectRatio,
-			Qt::SmoothTransformation);
-	}
 	_images[index] = PhotoImage{
 		.data = std::make_unique<Image>(std::move(image)),
 		.bytes = std::move(bytes),


ilya-fedin:
Looks like you're mentioning a duplicate issue rather than the original one. It also has no sense to mention an issue like that, if you want github to close the issue automatically, you should use "Fixes #2206". For GitHub to correctly parse it, it should also be not in commit/PR title, but in commit/PR body.

ilya-fedin:
See an example of how this should be done
https://github.com/telegramdesktop/tdesktop/commit/4dd1f95fbfb21bdec766863c56047774ce73c0de

john-preston:
Ok, let's try that one and see if there is any noticeable difference in OOM crash reports amount 😆

Neurotoxin001:
@john-preston A little confirmation is needed. Did I understand correctly that users will now be able to send images with 2560 resolution from Telegram Desktop? Because even without this fix I could see images with this resolution normally. For example - https://t.me/tass_agency/170735

Neurotoxin001:
If this fix won't allow sending 2560px images, then why did these issues was closed?
https://github.com/telegramdesktop/tdesktop/issues/6520
https://github.com/telegramdesktop/tdesktop/issues/2206
Telegram Web K is allow sending 2560px images btw.

ilya-fedin:
> Did I understand correctly that users will now be able to send images with 2560 resolution from Telegram Desktop?

No

> Because even without this fix I could see images with this resolution normally.

Are you sure it's not a result of upscaling due to UI interface scale?

> then why did these issues was closed?

They are about receiving photos, not sending them.

Neurotoxin001:
> Are you sure it's not a result of upscaling due to UI interface scale?

Yes. 

https://user-images.githubusercontent.com/39812401/209478343-e52c295a-90ba-4d25-b6f1-8c03dda9005a.mp4

image

> They are about receiving photos, not sending them.

This https://github.com/telegramdesktop/tdesktop/issues/6520 is about sending.

Lesiuk:
Saved to disk images were 2560px before my changes, but those displayed on the UI were down scaled to 1280px. Compare the image side by side, and you will see that the one displayed on the UI is clearly lower resolution.

Lesiuk:
There is more info inside the duplicate ticket I created 
https://github.com/telegramdesktop/tdesktop/issues/25582

Neurotoxin001:
Seems like you're right, after download this https://t.me/tass_agency/170735 image it's look more detailed. But this issue is still needs to be reopened. https://github.com/telegramdesktop/tdesktop/issues/6520

PR #25603, opened 2022-12-24
Use the same check for drag'n'drop image type as for file dialog

ilya-fedin:
The code path when choosing an image in file dialog tries to read a QImage with Images::Read and then only checks that mime type starts with image/ in ValidPhotoForAlbum

Core::FileIsImage and QImageReader::canRead checks should provide the same behavior

### DIFF
--- a/Telegram/SourceFiles/storage/storage_media_prepare.cpp
+++ b/Telegram/SourceFiles/storage/storage_media_prepare.cpp
@@ -12,7 +12,6 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 #include "storage/localimageloader.h"
 #include "core/mime_type.h"
 #include "ui/image/image_prepare.h"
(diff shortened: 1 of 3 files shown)


novomesk:
Mime type detection for AVIF and JPEG XL works on Windows when additional .XML file is installed.

More details in https://gitlab.com/lspies/photoqt/-/issues/169

ilya-fedin:
The XML is embedded in Qt sources AFAIK. The only thing may be required is a Qt patch updating the XML.

ilya-fedin:
A patch with these two commits squashed should help
https://github.com/qt/qtbase/commit/0cbbba2aa5b472241d45b4be6959a792062fbc30
https://github.com/qt/qtbase/commit/4872392925b5d4df3e307b71a9a575b7d26e4c7e

novomesk:
image/jxl is in shared-mime-info 2.2

PR #25764, opened 2023-01-21
Update installer script

bovirus:
Add Wizard Style modern
Add UsePreviousLangue=no to detect all new languages added in new version Separated gorup of simile setting with a new line
Add version settings missing
Add variable for starting copyright year.

### DIFF
--- a/Telegram/build/setup.iss
+++ b/Telegram/build/setup.iss
@@ -1,35 +1,47 @@
-#define MyAppShortName "Telegram"
-#define MyAppName "Telegram Desktop"
-#define MyAppPublisher "Telegram FZ-LLC"
-#define MyAppURL "https://desktop.telegram.org"
-#define MyAppExeName "Telegram.exe"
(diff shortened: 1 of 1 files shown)


Telegram/build/setup.iss
```diff
@@ -1,35 +1,47 @@
...
+
+UninstallDisplayName={#MyAppName}
+UninstallDisplayIcon={app}\Telegram.exe
+
+VersionInfoDescription={#MyAppName} installer
+VersionInfoProductName={#MyAppName}
+VersionInfoVersion={#MyAppVersion}
+
+WizardStyle=Modern
+UsePreviousLanguage=no
```
john-preston, line 34: Please remove "UsePreviousLangue=no", it works as expected with the default value.

john-preston (changes requested):

Telegram/build/setup.iss
bovirus: I believe that was already removed

PR #25869, opened 2023-02-09
Optimizing images size with oxipng

Alex-S1:
Optimized with [oxipng](https://github.com/shssoichiro/oxipng) v8.0.0

oxipng -o max --dir Compressed --strip all -a -i 0 -v -f 0-9 -t 2 -Z *.png
oxipng -o max --dir Compressed_zlib --strip all -a -i 0 -v -f 0-9 --zc 12 -t 2 *.png

1761 files compressed (lossless, without loss)

Before: 4085863
After: 2668310

-1417553 bytes (-1,35 MB)


23rd:
Instead of manually compressing each newly added png, it would be better to have a Github action that would check for new png files at every push commit, compress them, and open a pull request automatically.

ilya-fedin:
https://github.com/telegramdesktop/tdesktop/pull/4068#issuecomment-349868861

john-preston:
This is right, see the comment link above - this won't have any effect except the repository-with-full-history size increase.

PR #25894, opened 2023-02-15
Upgrade libde265, libheif, libjxl in Windows build and improvement in HEIF plug-in

novomesk:
These are mostly security improvements.

### DIFF
--- a/Telegram/build/prepare/prepare.py
+++ b/Telegram/build/prepare/prepare.py
@@ -710,9 +710,8 @@ release:
 
 stage('libde265', """
 win:
-    git clone https://github.com/strukturag/libde265.git
+    git clone --depth 1 -b v1.0.11 https://github.com/strukturag/libde265.git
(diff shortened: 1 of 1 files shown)


.gitmodules
```diff
@@ -94,6 +94,7 @@
 [submodule "Telegram/ThirdParty/kimageformats"]
 	path = Telegram/ThirdParty/kimageformats
 	url = https://github.com/KDE/kimageformats.git
+	branch = kf5
```
ilya-fedin, line 97: Is this really necessary? No other submodules specify a branch, they just checkout the commit directly

novomesk: I think it is, because master branch of kimageformats removed support for Qt5.
We can remove the branch when we use Qt6 only.

novomesk: OK, it can work also without mentioning the branch explicitly.

PR #25899, opened 2023-02-16
Upgrade highway, libde265, libheif, libjxl in Linux build

novomesk:


novomesk:
> libjxl is also present in snapcraft.yaml

That one I don't fully understand yet.

ilya-fedin:
@novomesk just update libjxl version and add the same build arg

ilya-fedin:
thank you a lot

PR #25900, opened 2023-02-17
Fix event loop for non-first instance on Linux

ilya-fedin:


Telegram/SourceFiles/platform/linux/specific_linux.cpp
```diff
@@ -245,6 +245,22 @@ void LaunchGApplication() {
...
+				return false;
+			}
+
+			try {
+				return base::Platform::DBus::NameHasOwner(
+					Gio::DBus::Connection::get_sync(
+						Gio::DBus::BusType::SESSION),
+					appId);
+			} catch (...) {
+			}
```
mymedia2, line 259: The idea of ignoring all exceptions by ellipsis is not considered a good practice. It is better to catch only expected exceptions and log them to facilitate easier debugging.

Issue #8361, opened 2020-08-01
[Question] Some errors occurred while building Telegram.According to the guidelines, it has been generated Telegram.sln，but something went wrong.


zhangqingfeng0105:
错误	MSB6006	“cmd.exe”已退出，代码为 -2147483645。	lib_ui_styles	D:\software\VS\Microsoft Visual Studio\2019\Professional\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets	231	
错误	C1083	无法打开源文件: “D:\TBuild\tdesktop\out\Telegram\lib_ui\lib_ui_autogen\mocs_compilation.cpp”: No such file or directory	lib_ui
错误	C1083	无法打开源文件: “D:\TBuild\tdesktop\out\Telegram\lib_ui\gen\styles\palette.cpp”: No such file or directory	lib_ui
错误	C1083	无法打开源文件: “D:\TBuild\tdesktop\out\Telegram\lib_ui\gen\styles\style_basic.cpp”: No such file or directory	lib_ui
错误	C1083	无法打开源文件: “D:\TBuild\tdesktop\out\Telegram\lib_ui\gen\styles\style_layers.cpp”: No such file or directory	lib_ui
错误	C1083	无法打开源文件: “D:\TBuild\tdesktop\out\Telegram\lib_ui\gen\emoji_suggestions_data.cpp”: No such file or directory	lib_ui
错误	C1083	无法打开源文件: “D:\TBuild\tdesktop\out\Telegram\lib_ui\gen\styles\style_widgets.cpp”: No such file or directory	lib_ui
错误	C1083	无法打开源文件: “D:\TBuild\tdesktop\out\Telegram\lib_ui\gen\emoji.cpp”: No such file or directory	lib_ui

错误	C1083	无法打开包括文件: “styles/style_widgets.h”: No such file or directory (编译源文件 D:\TBuild\tdesktop\Telegram\lib_ui\ui\effects\ripple_animation.cpp)	lib_ui	D:\TBuild\tdesktop\Telegram\lib_ui\ui\effects\ripple_animation.h	9	

错误	C1083	无法打开包括文件: “styles/style_widgets.h”: No such file or directory (编译源文件 D:\TBuild\tdesktop\Telegram\lib_ui\ui\effects\cross_animation.cpp)	lib_ui	D:\TBuild\tdesktop\Telegram\lib_ui\ui\effects\cross_animation.h	9	

[...]

PR #26095, opened 2023-03-25
CMake: Define the SDK (Windows) version

temajm:
I suggest that you define the SDK (Windows) version in CMakeLists.txt so that CMake will set this version when generating code.

The version you propose to use in https://github.com/telegramdesktop/tdesktop/blob/dev/docs/building-win.md:
image

This PR, I assume, will solve the problem:
https://github.com/telegramdesktop/tdesktop/issues/8361
https://github.com/desktop-app/lib_ui/issues/112

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,10 +25,6 @@ if (APPLE)
 endif()
 
 if (WIN32)
-	set(CMAKE_SYSTEM_VERSION 10.0.22000.0)
-endif()
-
-if (WIN32)
 	if(NOT ${CMAKE_GENERATOR} STREQUAL "Visual Studio 17 2022")
 		message(SEND_ERROR "It will not be possible to build codegen_emoji and therefore lib_ui on ${CMAKE_GENERATOR} (Please use Visual Studio 17 2022)")
 	endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,10 @@ if (APPLE)
     set(project_langs C CXX OBJC OBJCXX)
 endif()
 
+if (WIN32)
+	set(CMAKE_SYSTEM_VERSION 10.0.22000.0)
+endif()
+
 project(Telegram
     LANGUAGES ${project_langs}
     VERSION ${desktop_app_version_cmake}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,12 @@ if (WIN32)
 	set(CMAKE_SYSTEM_VERSION 10.0.22000.0)
 endif()
 
+if (WIN32)
+	if(NOT ${CMAKE_GENERATOR} STREQUAL "Visual Studio 17 2022")
+		message(SEND_ERROR "It will not be possible to build codegen_emoji and therefore lib_ui on ${CMAKE_GENERATOR} (Please use Visual Studio 17 2022)")
+	endif()
+endif()
+
 project(Telegram
     LANGUAGES ${project_langs}
     VERSION ${desktop_app_version_cmake}


ilya-fedin:
Those errors look like from the IntelliSense rather than from build which is completely expected as those files are generated by codegens.

temajm:
> Those errors look like from the IntelliSense rather than from build which is completely expected as those files are generated by the codegens.

Their errors are caused by builds, due to incompatibilities between codegen_emoji and other versions of the SDK, so codegen cannot build codegen_emoji.exe and other files. (You can check this for yourself by changing the SDK version of the codegen_emoji project and running the build)

image

ilya-fedin:
> Their errors are caused by builds, namely because **codegen_emoji** is incompatible with other versions of the SDK, so codegen cannot build **codegen_emoji.exe** and other files. (You can check this for yourself by changing the SDK version of the **codegen_emoji** project and running the build)

I'm highly doubt this is true as I'm highly doubt this file is updated every time @john-preston update Visual Studio on his computer, so he is likely building with other version of SDK nowadays.

ilya-fedin:
I'd also say we can't know which error exactly they encounter as they provide IntelliSense errors instead of the build ones.

ilya-fedin:
Googling your error seem to say that you have updated SDK but haven't rebuilt all the dependencies, so you're getting a mix of standard libraries ABI usages.

temajm:
Вы правы, проблема заключается в другом. Необходимо использовать VS 17 2022 для удачной сборки codegen_emoji & lib_ui. (Хоть это не точно поможет решить проблему указанных авторов issues, но я более чем уверен, что у них как раз проблема из-за этого)

image
image

ilya-fedin:
I have a feel hardcoding SDK version will lead to problems

temajm:
> I have a feel hardcoding SDK version will lead to problems

You don't mean the SDK, but Visual Studio?

ilya-fedin:
No, I mean SDK you're hardcoding in CMAKE_SYSTEM_VERSION

temajm:
> No, I mean SDK you're hardcoding in CMAKE_SYSTEM_VERSION

Oh, I'm sorry, I only just noticed I forgot to put it away. It's no longer required. :)

ilya-fedin:
I also believe that the problem is not that you need exactly VS 17 2022, but that you have multiple versions of Visual Studio installed and you use another SDK version when building dependencies.

ilya-fedin:
This is definitely __not__ the right fix for the error in question. The right fix would be to somehow detect SDK version of the dependencies, ensure they all are built with the same SDK and then ensuring we're using the same SDK. But I'm not sure it worth the effort.

temajm:
В двух вариантах сборки я использовал SDK v10.0.22621.0

Visual Studio 17 2022 (v143): image

Visual Studio 16 2019 (v142): 
image

ilya-fedin:
Were you using the same Visual Studio/SDK for the dependencies built with prepare.py at each step?

temajm:
Edited it manually

image

ilya-fedin:
I'm not sure what you wanted to say with that. prepare.py is being launched from the command line, so it can use completely another compiler/standard library version if you have multiple toolchains installed. And it's your job (at least currently) to ensure both prepare.py and tdesktop use the same compiler/standard library version.

temajm:
I want to inform the developer that in a version other than VS 17 2022, the codegen_emoji project will not be built and thus neither will lib_ui. And to implement this, I need to leave this message when generating the root CMakeLists.txt. All dependencies and other use the same toolkit specified when generating the root CMakeLists.txt

ilya-fedin:
@temajm in case of a real incompatibility you should have a compile-time error. But the error on your screenshot comes from the linker saying about different compilers were used for building some of the static libraries.

ilya-fedin:
> Visual Studio 16 2019 (v142):

that screenshot points exactly on tg_angle from prepare.py being the culprit, as you can see, it uses some function from the standard library that is not present in the standard library used in your tdesktop build. I.e. some (if not all) libraries built by prepare.py were built by VS 2022 and VS 2019's stdlib doesn't have such symbol yet in its ABI.

temajm:
@ilya-fedin You're right)

temajm:
An entirely different solution is required. A more effective one

PR #26190, opened 2023-04-19
fixed typo

ErvinCret:


Aokromes:
i think that was intended @john-preston

john-preston:
It is the side limit (each image side shouldn't be more than 2560px).

PR #26199, opened 2023-04-21
Option to export Data as HTML and JSON

mrbesen:
This should save users (and telegram servers), that want both: time, disk space and traffic.

This is how the menu in the general export looks:
grafik

I have exported everything with it. The resulting html and json files looked fine to me.


Telegram/SourceFiles/export/output/export_output_both.cpp
```diff
@@ -0,0 +1,111 @@
...
+	_html = std::make_shared<HtmlWriter>();
+	_json = std::make_shared<JsonWriter>();
+}
+
+Result BothWriter::start(
+		const Settings &settings,
+		const Environment &environment,
+		Stats *stats) {
+	Expects(_html != nullptr);
+	Expects(_html != nullptr);
```
fenderchamp, line 27: shouldn't one of these be _json?

mrbesen: yes, i fixed that.

mrbesen:
Mentioned here: https://bugs.telegram.org/c/27133

Aokromes:
https://github.com/telegramdesktop/tdesktop/commit/29bfe4338629be950e0dd32f5a77d00771fca4ea already merged another version?

OTC15203:
OK HOW

Issue #24014, opened 2022-02-03
tdesktop 3.5.x and 3.6.x compilation with clang 13-14 failed


h0tc0d3:
### Steps to reproduce

Build with clang 13. The previous version tdesktop worked fine.

### Expected behaviour

-

### Actual behaviour

```
[ 53%] Building CXX object Telegram/CMakeFiles/Telegram.dir/SourceFiles/api/api_hash.cpp.o
[ 53%] Building CXX object Telegram/CMakeFiles/Telegram.dir/SourceFiles/api/api_polls.cpp.o
[ 53%] Building CXX object Telegram/CMakeFiles/Telegram.dir/SourceFiles/api/api_sending.cpp.o
In file included from <built-in>:485:
In file included from /build/telegram-desktop/src/build/Telegram/CMakeFiles/Telegram.dir/cmake_pch.hxx:5:
In file included from /build/telegram-desktop/src/tdesktop-3.5.0-full/Telegram/SourceFiles/stdafx.h:102:
In file included from /usr/include/range/v3/all.hpp:17:
In file included from /usr/include/range/v3/action.hpp:32:
In file included from /usr/include/range/v3/action/split.hpp:28:
/usr/include/range/v3/range/conversion.hpp:379:29: error: alias template 'QVector' requires template arguments; argument deduction only allowed for class templates
                -> decltype(ContT(range_cpp17_iterator_t<Rng>{},
                            ^
/usr/include/meta/meta.hpp:541:5: note: in instantiation of template class 'ranges::detail::from_range<QVector>' requested here
    using invoke = typename Fn::template invoke<Args...>;
    ^
/usr/include/range/v3/range/conversion.hpp:46:13: note: in instantiation of template type alias 'invoke' requested here
            using container_t = meta::invoke<MetaFn, Rng>;
            ^
/usr/include/range/v3/range/conversion.hpp:338:46: note: in instantiation of template type alias 'container_t' requested here
                    convertible_to_cont<Rng, container_t<MetaFn, Rng>>)
                                             ^
/usr/include/range/v3/detail/prologue.hpp:35:26: note: expanded from macro 'template'
    template<__VA_ARGS__ CPP_TEMPLATE_AUX_                                                               ^
/usr/include/range/v3/range/conversion.hpp:304:13: note: in instantiation of template class 'ranges::detail::to_container::fn<ranges::detail::from_range<QVector>>' requested here
          , Fn
            ^
[...]

PR #26219, opened 2023-04-24
Explicitly specify QVector element type to fix build with clang13+range-v3+qt6

klemensn:
Same story as 2b383a4 for #24014.
Fixes 4.8.0 build on OpenBSD/amd64 7.3 with clang 13.0.0, range-v3 0.12.0 and Qt 6.5.0.


john-preston:
That's a shame :(

Issue #26040, opened 2023-03-15
Telegram Winget package is outdated


mimi89999:
### Steps to reproduce

```
PS C:\Users\Michel> winget show Telegram.TelegramDesktop
Trouvé Telegram Desktop [Telegram.TelegramDesktop]
Version : 4.4.1
Publisher : Telegram FZ-LLC
ID de l’Editeur https://telegram.org
URL de support de l’éditeur : https://telegram.org/support
Auteur : Telegram FZ-LLC
Description : Telegram is a cloud-based mobile and desktop messaging app with a focus on security and speed.
Page d’accueil : https://desktop.telegram.org
Licence : GNU-3.0-only
URL de la licence : https://raw.githubusercontent.com/telegramdesktop/tdesktop/master/LICENSE
URL de confidentialité : https://telegram.org/privacy
URL de copyright : https://raw.githubusercontent.com/telegramdesktop/tdesktop/master/LICENSE
Mots-clés :
  chat
  messaging
  messenger
Programme d'installation :
  Type du programme d’installation : inno
  URL du programme d’installation : https://github.com/telegramdesktop/tdesktop/releases/download/v4.4.1/tsetup-x64.4.4.1.exe
  SHA256 du programme d’installation : 4cce2dcd18cd536719c0ac6bfefefc11c45faaa4736ddf67431c20bfe8f7e00d
```

### Expected behaviour

Winget should have the latest version

### Actual behaviour

The version in winget is `4.4.1`

### Operating system

Microsoft Windows 11 Professionnel

### Version of Telegram Desktop

4.6.5

### Installation source

Other (unofficial) source

Aokromes:
https://github.com/microsoft/winget-pkgs/commit/175c6f0164e39278595eee34c69ee426930713ab it looks someone else maintains it.

Aokromes:
@vedantmgoyal2009 maybe your tool is broken now?

Aokromes:
@mimi89999 check if now it's updated.

mimi89999:
Now I see `4.7.1` in Winget, but the latest tag in the repo is `v4.8.1`.

Aokromes:
repo was updated again, can you check today or tomorrow? :)

PR #26324, opened 2023-05-21
Fix installer selection for winget-releaser

vedantmgoyal2009:
* Fixes https://github.com/telegramdesktop/tdesktop/issues/26040


vedantmgoyal2009:
cc @john-preston

john-preston:

PR #26389, opened 2023-06-09
snap: connect to the network-bind interface

kenvandine:
snap: connect to the network-bind interface, necessary on systems where desktop-legacy isn't supported like Ubuntu Core Desktop.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -31,6 +31,7 @@ apps:
       - hardware-observe
       - home
       - network
+      - network-bind
       - network-status
       - opengl
       - removable-media


ilya-fedin:
But tdesktop doesn't bind to anything...

kenvandine:
> But tdesktop doesn't bind to anything...

It's related to use of a unix domain socket, listen is caught in the seccomp filter..  It's allowed with the desktop-legacy interface in snapd on classic systems, but on Ubuntu Core nothing provides desktop-legacy.

ilya-fedin:
desktop-legacy provides other important permissions, like StatusNotifierItem for the tray icon, how does tray work on Ubuntu Core Desktop?

ilya-fedin:
I tend to think notifications will misfunction as well as desktop-legacy is the interface providing access to StartServiceByName

PR #26415, opened 2023-06-17
Use absolute path of sed instead

CaledoniaProject:
I installed and linked gnu-sed and it would fail here. I need to explicitly specify /usr/bin/sed to resolve this.

### DIFF
--- a/Telegram/build/prepare/prepare.py
+++ b/Telegram/build/prepare/prepare.py
@@ -489,7 +489,7 @@ stage('xz', """
 !win:
     git clone -b v5.2.9 https://git.tukaani.org/xz.git
     cd xz
-    sed -i '' '\\@check_symbol_exists(futimens "sys/types.h;sys/stat.h" HAVE_FUTIMENS)@d' CMakeLists.txt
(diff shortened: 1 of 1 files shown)


ilya-fedin:
I'm not sure it's a good idea. It may be not gnu-sed and it could be installed in various ways being in any folder. I believe it's better for you to fix your PATH. Or maybe even check what you did wrong that it's not in PATH automatically.

CaledoniaProject:
@ilya-fedin The command inside prepare.py works with only Mac builtin sed, not gnu sed. And since MacOS has SIP enabled by default /usr/bin/sed will not usually be modified.

In my case I installed gnu-sed and it fails to run, because the command line syntax. It's better to use /usr/bin/sed to resolve this issue.

ilya-fedin:
I'm highly doubt it's better. You can fix your PATH so it has sed. It also works just fine in the CI now.

ilya-fedin:
There's likely a lot of other similar reimplementations of standard tools people can install, does that mean all the prepare.py should be modified to use absolute paths? I don't think it's a good idea to open that pandora box.

Issue #26414, opened 2023-06-17
Failed to build, opus not found with pkg-config


CaledoniaProject:
### Steps to reproduce

I'm following exactly the steps here
https://github.com/telegramdesktop/tdesktop/blob/dev/docs/building-win-x64.md

### Expected behaviour

I expect tdesktop\Telegram\build\prepare\win.bat to finish successfully

### Actual behaviour

The prepare script failed

![111](https://github.com/telegramdesktop/tdesktop/assets/1357701/429d2d46-b025-494a-8958-5767164b688e)

### Operating system

Windows Server 2022

### Version of Telegram Desktop

latest

### Installation source

Other (unofficial) source

### Logs

```shell
Packages (1) pkgconf-1.9.4-1

Total Download Size:   0.06 MiB
Total Installed Size:  0.29 MiB

:: Proceed with installation? [Y/n]
:: Retrieving packages...
 pkgconf-1.9.4-1-x86_64                          61.6 KiB  65.7 KiB/s 00:01 [#########################################] 100%
(1/1) checking keys in keyring                                              [#########################################] 100%
(1/1) checking package integrity                                            [#########################################] 100%
(1/1) loading package files                                                 [#########################################] 100%
(1/1) checking for file conflicts                                           [#########################################] 100%
(1/1) checking available disk space                                         [#########################################] 100%
:: Processing package changes...
(1/1) installing pkgconf                                                    [#########################################] 100%
sed 's#@@PREFIX@@#C:/Users/Administrator/Desktop/Libraries/win64/local#' ffnvcodec.pc.in > ffnvcodec.pc
install -m 0755 -d '/c/Users/Administrator/Desktop/Libraries/win64/patches/../local/include/ffnvcodec'
install -m 0644 include/ffnvcodec/*.h '/c/Users/Administrator/Desktop/Libraries/win64/patches/../local/include/ffnvcodec'
install -m 0755 -d '/c/Users/Administrator/Desktop/Libraries/win64/patches/../local/lib/pkgconfig'
install -m 0644 ffnvcodec.pc '/c/Users/Administrator/Desktop/Libraries/win64/patches/../local/lib/pkgconfig'
ERROR: opus not found using pkg-config

If you think configure made a mistake, make sure you are using the latest
version from Git.  If the latest version fails, report the problem to the
ffmpeg-user@ffmpeg.org mailing list or IRC #ffmpeg on irc.libera.chat.
[...]

PR #26424, opened 2023-06-18
Update building-win-x64.md

CaledoniaProject:
Fix issues mentioned here https://github.com/telegramdesktop/tdesktop/issues/26414

### DIFF
--- a/docs/building-win-x64.md
+++ b/docs/building-win-x64.md
@@ -8,7 +8,7 @@
 
 ## Prepare folder
 
-The build is done in **Visual Studio 2022** with **10.0.22000.0** SDK version.
+The build is done in **Visual Studio 2022** with **10.0.22000.0** SDK version, do not install **C++ CMake tools for Windows**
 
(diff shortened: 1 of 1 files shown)


ilya-fedin:
I'm not sure it's a good idea given that you haven't confirmed it's the source of your problem

Aokromes:
oh i have seen people having this issue on trinitycore too, and yes, installing separate cmake instead C++ CMake tools for Windows fixes a lot of issues.

ilya-fedin:
Maybe they have updated their cmake with time though? Haven't seen such reports for a long time.

docs/building-win-x64.md
```diff
@@ -8,7 +8,7 @@
 
 ## Prepare folder
 
-The build is done in **Visual Studio 2022** with **10.0.22000.0** SDK version.
+The build is done in **Visual Studio 2022** with **10.0.22000.0** SDK version, do not install **C++ CMake tools for Windows**
```
ilya-fedin, line 11: So I have removed manually installed cmake altogether and switched to cmake from Visual Studio. No issues. This line doesn't need any changes.

PR #26517, opened 2023-07-17
Upgrade dav1d, libde265, libheif, libjxl

novomesk:
So that Linux build uses same versions like Windows build.

### DIFF
--- a/Telegram/build/docker/centos_env/Dockerfile
+++ b/Telegram/build/docker/centos_env/Dockerfile
@@ -196,7 +196,7 @@ RUN git clone -b v1.3.1 --depth=1 {{ GIT }}/xiph/opus.git \
 FROM builder AS dav1d
 COPY --link --from=nasm {{ LibrariesPath }}/nasm-cache /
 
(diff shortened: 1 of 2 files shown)


ilya-fedin:
I wonder whether there are new versions of lcms2, brotli and highway? :thinking: 
While the Windows script probably relies on libjxl builds script to build them, Dockerfile does that manually...

novomesk:
There are newer versions of other libraries too.

However I think it is not always necessary to jump to the latest version quickly. Sometimes new versions have new problems.

Upgrades to libde265, libheif, libjxl solve discovered vulnerabilities and I use dav1d 1.2.1 for some time too.

I am submitting changes I am confident with.

ilya-fedin:
I thought those libraries are participating in the file processing and so are likely to be the cause of vulnerabilities? I don't monitor that though and don't know whether there are vulnerabilities caused by those libraries (which could be considered a part of libjxl as they're submodules in libjxl repo).

novomesk:
I understand.

That vulnerability fixed in libjxl 0.8.2 is only related to libjxl itself, it is not related to submodule libraries.

PR #26522, opened 2023-07-18
Create SECURITY.md

dimas1980200:


john-preston:
In tdesktop there is a single sequence of versions, so the first suggested part of the file doesn't look relevant. The second - the official faq at

https://telegram.org/faq#q-why-should-i-trust-you

suggests writing to security@telegram.org

PR #26565, opened 2023-07-24
Use absolute executable path for dbus service

Avimitin:
After the v4.8.7 release, I cannot launch telegram desktop because of the recent dbus support introduced in
https://github.com/telegramdesktop/tdesktop/commit/2dfe85832717ccfd9268a5b767f03eb4e8665b8f. And it was fixed after I changed the value of the Exec field to absolute path to the telegram-desktop executable.

In [dbus specification](https://dbus.freedesktop.org/doc/dbus-specification.html#message-bus-starting-services) they don't mention that they will find executable from PATH variable, so this is a UB.
After this commit, cmake will use the full install path to generate the correct service file and fix the above issue.


ilya-fedin:
Hmm, what the error you get as-is?

Telegram/CMakeLists.txt
```diff
@@ -1759,6 +1759,7 @@ endif()
 if (LINUX AND DESKTOP_APP_USE_PACKAGED)
     include(GNUInstallDirs)
     configure_file("../lib/xdg/org.telegram.desktop.metainfo.xml" "${CMAKE_CURRENT_BINARY_DIR}/org.telegram.desktop.metainfo.xml" @ONLY)
+    configure_file("../lib/xdg/org.telegram.desktop.service" "${CMAKE_CURRENT_BINARY_DIR}/org.telegram.desktop.service" @ONLY)
```
ilya-fedin, line 1762: Please place this before the previous line

Avimitin:
> Hmm, what the error you get as-is?

When launching telegram-desktop through KDE krunner because it fail to find excutable specify in the Exec field in the service file. It will report that "no such file or directory".

image


> I'm not sure having .desktop and D-Bus service files having different executable path is a good thing to do...

In freedesktop spec, they specify that ["If no full path is provided the executable is looked up in the $PATH environment variable used by the desktop environment."](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#exec-variables), so it is ok to use only program name for the .desktop file. However the `Exec` key has different bahavior in dbus spec.

ilya-fedin:
> In freedesktop spec, they specify that ["If no full path is provided the executable is looked up in the $PATH environment variable used by the desktop environment."](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#exec-variables), so it is ok to use only program name for the .desktop file. However the `Exec` key has different bahavior in dbus spec.

I mean this could lead to different executables launched if there are multiple telegram-desktop binaries in different directories of the $PATH.

Telegram/CMakeLists.txt
Avimitin: Fixed and amended in the latest commit.

Avimitin:
> I mean this could lead to different executables launched if there are multiple telegram-desktop binaries in different directories of the $PATH.

What if we force them all use the full install path as the executable path? I think this can avoid a lot of problem as you mentioned.

ilya-fedin:
This is not trivial as this means some logic should be changed because .desktop's Exec value must be the same as the value provided to various functions generating the .desktop files in runtime (for tg:// handling, for autostart, etc). It also complicates things on distros like NixOS where the runtime-gotten binary path is unlikely to be the same what will lead to problems like generating broken .desktop files.

ilya-fedin:
Having full path in those files is really a big pain in the ass.

ilya-fedin:
> When launching telegram-desktop through KDE krunner because it fail to find excutable specify in the Exec field in the service file. It will report that "no such file or directory".

Thanks, what does this command return: `busctl --user call org.telegram.desktop /org/telegram/desktop org.freedesktop.Application Activate 'a{sv}' 0`? Do you use dbus-broker or dbus-daemon?

ilya-fedin:
> Call failed: Failed to execute program org.telegram.desktop: No such file or directory

What about the second question? :thinking:

Avimitin:
> What about the second question? thinking

I am using dbus-daemon.

ilya-fedin:
Oh, I was hoping that maybe it's only with dbus-broker and maybe we can apply some systemd magic as it starts all the services via systemd...

Looks like nautilus had full path in the d-bus service file but not .desktop file so maybe it's good to go...

Avimitin:
Also note worthy that maybe we don't need to change the .desktop file. In the freedesktop manual they mentioned that: " The Exec key is required if DBusActivatable is not set to true in the main desktop entry group. Even if DBusActivatable is true, Exec should be specified for compatibility with implementations that do not understand DBusActivatable."

So it seems that changing only the dbus service file and leaving the .desktop file unchanged will not break other functionality. And DbusActivation only affect how the program is launched.

ilya-fedin:
D-Bus activation is exactly about launching the program when it's *not* launched

Avimitin:
> This is not trivial as this means some logic should be changed because .desktop's Exec value must be the same as the value provided to various functions generating the .desktop files in runtime (for tg:// handling, for autostart, etc). 

So will all these functions break if we force the use of absolute path in the dbus service file but leave desktop file unchanged?

ilya-fedin:
> So will all these functions break if we force the use of absolute path in the dbus service file but leave desktop file unchanged?

No, they will if the path in .desktop file would be introduced without adapting the code. The only weird thing that could happen with the current variant if there are multiple telegram-desktop binaries in the path but I believe it's minor in comparison to headache full path in .desktop file would introduce.

Avimitin:
> The only weird thing that could happen with the current variant if there are multiple telegram-desktop binaries in the path but I believe it's minor in comparison to headache full path in .desktop file would introduce.

Yeah, I don't think people will install and use multiple versions of telegram in their daily life. The only scenario I can think of is debugging and comparing telegram behavior between different versions, then I think these users will know how to do this correctly.

PR #27007, opened 2023-10-30
Update README.md

catterpiler74:
add link for old version with blue icon for people who need this (95% need this)


john-preston:
The upcoming version will include a setting for turning on the old tray icon back, it'll be on for the old installations and off by default for new installations.

PR #27035, opened 2023-11-02
Add symbolic app icon

jsparber:
GNOME Shell uses the symbolic icon in some places.


ilya-fedin:
Can you explain better why should this be merged? As-is it's not clear what this change would affect.

jsparber:
> Can you explain better why should this be merged? As-is it's not clear what this change would affect.

GNOME uses the symbolic icon in the topbar next to the app menu and in notifications when the app doesn't provide a icon for the notification.

ilya-fedin:
So you're proposing a window icon change. I'm not sure that's allowed...
@john-preston what do you think?

jsparber:
> So you're proposing a window icon change. I'm not sure that's allowed... @john-preston what do you think?

No it's not. It just uses the symbolic icon in specific places, defined by the platforms theme. For example currently in GNOME the hicolor icon is used in the top bar but made monochrome, because it Telegram doesn't provide a symbolic variant and it would look bad on the platform. 

How it will look like
image

ilya-fedin:
You say it's not but I see on your screenshot that the window icon is changed

ilya-fedin:
You say it's not but I see on your screenshot that the window icon is changed

jsparber:
> You say it's not but I see on your screenshot that the window icon is changed

This is not the window icon, it's the icon used in the topbar.

Btw this is how it looks currently:
image

ilya-fedin:
This is the same thing, it's just you call it an other way

ilya-fedin:
This is the same thing, it's just you call it an other way

john-preston:
I've added our svg icon to the repository:

https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/Resources/icons/tray_monochrome.svg

Maybe it could be used instead?

ilya-fedin:
Please rebase rather than merging

jsparber:
> Please rebase rather than merging

Sorry about that. I updated the commit using the icon you added.

ilya-fedin:
@john-preston you maybe (unless you don't want this for official builds) merged this too early as the PR doesn't contain code to install the icon for official build

jsparber:
> @john-preston you maybe (unless you don't want this for official builds) merged this too early as the PR doesn't contain code to install the icon for official build

what do you mean?

ilya-fedin:
Official build copies the icons in runtime with the code in specific_linux.cpp. This code is no-op for the official build.

PR #27038, opened 2023-11-03
Fix macos build, use xcode 15

agl-1984:
MacOS workflow is failing for few days already. 
CMake sets -ld_classic flag, that exists in xcode 15 only, but default is xcode 14.2 / 14.3.

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -40,7 +40,7 @@ jobs:
 
   macos:
     name: MacOS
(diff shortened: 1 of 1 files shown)


ilya-fedin:
After discussing with @john-preston, it was decuded the right fix is to port https://github.com/desktop-app/cmake_helpers/commit/78098ede77a09e41da6233d765d02b43a60e7138 to desktop-app/crashpad

Issue #24003, opened 2022-02-01
Problems with non-sticker webm with VP9


Remu-rin:
### Steps to reproduce

1. Open https://t.me/webm_tdesktop_bug

### Expected behaviour

Should play normally.

### Actual behaviour

VP9 + vorbis/opus are shown as video, but won't play internally. They are downloaded and opened in external player.
VP9 without audio plays, but displayed duration is bugged and/or bugs/slows when seeking.

VP8 + vorbis/opus seems to work.

### Operating system

Windows 10

### Version of Telegram Desktop

3.5

### Installation source

Static binary from official website

MatthewRobertDunn:
Interestingly the example given all 100% works for me on windows 10 version 3.7.3

Some of my friends using the same version of telegram NONE of them work.

![image](https://user-images.githubusercontent.com/15967209/171077125-d04352d8-4278-4113-aca0-1ad7c6a93dab.png)

Aokromes:
maybe the bug is lack of system codecs?

try downloading https://get.daum.net/PotPlayer/Codec/OpenCodecSetup64.exe or https://get.daum.net/PotPlayer/Codec/OpenCodecSetup.exe

Remu-rin:
I thought it is a server-side thing now. When I created this issue, vp9 was bugged/not supported in tdesktop. Then vp9 got fixed, but video/file behavior became inconsistent. Sometimes all posts in that channel are files, sometimes all are videos. And most of the time part of them are videos and part of them are files. Even vp8 ones are fluctuating.

On Android webm are always still a file, so if anyone wants webm as video on every Telegram client, vote here: https://bugs.telegram.org/c/2364

egzakharovich:
TDesktop 4.3.3 Windows 10
It's even worse at this moment, webm's don't play as videos at all.

![image](https://user-images.githubusercontent.com/68099035/204229298-c4497324-b4a5-4c86-8ee6-45c503382264.png)

john-preston:
This is regulated by the server :( It has attributes with a thumbnail and with a flag that the file is streamable. For now it returns no thumbnail and no flag, I guess because Android in some versions was showing any such file as a sticker and it was bad.

Issue #10000, opened 2020-12-24
[Feature Request] Uploading and Playing MKV as video


AhyaH-YHWH:
https://t.me/TelegramDesktopTalk/53212
TDesktop 2.5.1 windows Internal player can't play MKV videos unlike Linux client

Also bots can upload MKV videos as stream-able videos, 
please add ability to upload MKV videos with desktop as stream-able video

Neurognostic:
I am using tdesktop 2.5.1 on Linux and MKV videos don't play either. Would be great to support MKV container format, everytime I share a video on Telegram I have to change the container to mp4.

RememberTheAir:
unstale

soredake:
Not stale.

lcomrade:
We need MKV.

brainrom:
Still actual. Even renaming to mp4 helps to send it as streamable video.

VetaLioSTV:
Totally agree, it should be added.

For now **4.6.2 x64**(and i think for previous versions also) -> you can use workaround, just rename mkv file to **.mp4** and send it through telegram Desktop, it will work fine without convertation to proper format .mp4, since underhood file remains as matroska container. The only problem which remains is that in iOS client at least video will not work fully, but still usable for desktop client.

filhocf:
I tried to upload a mkv video now and view it in my android device, without success. It starts download the file to see.

This workaround works only in the desktop and not in mobile devices.

GatoSoft-Software:
We still need MKV full support. Changing file extension to .mp4 can lead to confusion and also for the sake of adding features like selecting streams embedded in .mkv files.

ccfelix:
Changing file extension don't work anymore. We still need MKV support.

mikkovedru:
Everybody who is interested in this feature should go to Telegram's official bug/feature site and upvote all the .mkv related issues they have: https://bugs.telegram.org/?sort=rate&query=mkv

PR #27117, opened 2023-11-17
Fix embedding for .mkv .webm

PeterNespesny:
Related:  https://github.com/telegramdesktop/tdesktop/issues/10000
Client doesn't treat .mkv as a video file type when user tries to upload one.
mkv_noembed
Fixing that by returning "video/mp4" in MimeTypeForFile
mkv_embed

Related:  https://github.com/telegramdesktop/tdesktop/issues/24003
Client fools the user by displaying uploaded .webm videos as embedded for some time, but they do not get embedded server-side and [turn into documents](https://github.com/telegramdesktop/tdesktop/issues/24003#issuecomment-1328707916). Why not fix it with the most logical solution of supplying .webm videos with MIME type "video/mp4" and changing the filename: filename.webm to filename[webm].mp4
webm_embed


23rd:
What if *.mkv file has 5 video and 10 audio tracks inside? What the client should to do?

Telegram/SourceFiles/core/mime_type.cpp
```diff
@@ -111,6 +111,9 @@ MimeType MimeTypeForFile(const QFileInfo &file) {
 		return MimeType(MimeType::Known::TDesktopTheme);
 	} else if (path.endsWith(u".tdesktop-palette"_q, Qt::CaseInsensitive)) {
 		return MimeType(MimeType::Known::TDesktopPalette);
+	} else if (path.endsWith(u".mkv"_q, Qt::CaseInsensitive)) {
+		// This will ensure .mkv is detected as a video file by the client when drag-n-dropped
+		return MimeType(QMimeDatabase().mimeTypeForName("video/mp4"));
```
ilya-fedin, line 116: This doesn't look like a right change: if you want to allow .mkv to be videos then do exactly that, don't pretend it to be .mp4. But I guess @john-preston won't allow that. Also, most mkv videos are unlikely to be played by the official builds as they have ffmpeg built only with require codecs enabled.

Telegram/SourceFiles/data/data_document.cpp
```diff
@@ -875,6 +875,11 @@ ChatRestriction DocumentData::requiredSendRight() const {
 void DocumentData::setFileName(const QString &remoteFileName) {
 	_filename = remoteFileName;
 
+	// Needed in order to force the server to embed the video
+	if (_filename.toLower().endsWith(".webm")) {
```
ilya-fedin, line 879: This really looks like a candidate for various forks as getting through explicit server-side restrictions in an official client looks very weird.

PeterNespesny: 1) Restricting .webm embedding server-side doesn't make any sense
2) User can achieve the same result by changing the file extension manually

ilya-fedin: This doesn't mean official client is allowed to help the user to overcome server-side restriction

Telegram/SourceFiles/core/mime_type.cpp
PeterNespesny: > if you want to allow .mkv to be videos then do exactly that, don't pretend it to be .mp4

Yes, I do believe adding .mkv to extensions in FileLoadTask::CheckForVideo would achieve the same result.
`static const auto extensions = {
		u".mp4"_q,
		u".mov"_q,
		u".m4v"_q,
		u".webm"_q,
};`

ilya-fedin: You can also add the mimetype. Searching by `video/mp4` it's not the only place where it has to be added for full-feature support.

Aokromes:
> What if *.mkv file has 5 video and 10 audio tracks inside? What the client should to do?

imho, prefer defaults :)

ilya-fedin:
@Aokromes I guess @23rd means tdesktop doesn't handle that and the UX with mkvs is likely to be poor

Telegram/SourceFiles/data/data_document.cpp
PeterNespesny: If there's such a server-side restriction, then `FileLoadTask::CheckForVideo` probably shouldn't return `true` for webms resulting in [broken embedding](https://github.com/telegramdesktop/tdesktop/issues/24003#issuecomment-1141588641)

Issue #26917, opened 2023-10-10
Nvidia Shadowplay video crash


kukuruzka165:
### Steps to reproduce

1. Click on [this](https://t.me/TelegramDesktopTalk/106448) video

### Expected behaviour

Video just play as normal

### Actual behaviour

The client crashes

### Operating system

Windows 11

### Version of Telegram Desktop

v4.10.2

### Installation source

Static binary from official website

### Logs

```shell
[2023.10.10 22:55:02] Assertion Failed! "!from.size.isEmpty()" media_streaming_player.cpp:60
```

kukuruzka165:
I have asked people who have this crash or don't and:

- The crash occurs on all OSs, but does not occur on new nvidia GPUs (probably due to video transcoding or something else, idk)
- The crash is caused by a large SAR value.
ffprobe says: 
> Stream #0:0[0x1](und): Video: hevc (Main) (hvc1 / 0x31637668), yuvj420p(pc, bt709), 2154x1250, 16887 kb/s, SAR 1346875:1347327 DAR 2155:1251, 59.88 fps, 59.94 tbr, 59874 tbn (default)

If this value is edited through a hex editor, it stops crashing

PR #27160, opened 2023-11-28
Fix crash on some videos

kukuruzka165:
This fixes issue #26917 when the client crashed playing a video with a non-standard SAR (pixel size) due to an incorrect calculation of the video width, it was -1034


john-preston:
Thanks!

PR #27161, opened 2023-11-28
add bash shebang

mrbesen:
I was having problems when running the script with zsh. Adding a bash shebang avoids that.

```
./Telegram/build/prepare/linux.sh: 3: pushd: not found
```


ilya-fedin:
Most scripts don't have shebangs in the repo. The ones that do use `#!/bin/bash`.

mrbesen:
i changed it to `#!/bin/bash` now

Issue #2198, opened 2016-06-27
Prefer IPv6 if its available


nshopik:
### Steps to reproduce
1. Open Telegram on dual-stacked host
2. Telegram tries connect via IPv4 and IPv6
3. Then drops IPv6 connection and stick to IPv4 only
### Expected behaviour

It should stick to IPv6 if its available and not crippled, similar to happy eyeballs **RFC6555**
### Actual behaviour

Always prefers IPv4, only disabling IPv4 make it stick to IPv6.
### Configuration

Xubuntu 16.04
`4.4.0-24-generic #43-Ubuntu SMP Wed Jun 8 19:27:37 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux`

v 0.9.56 alpha
### Logs

startup log
https://gist.github.com/nshopik/31528efa4cce7dbccdd7d99f5ef9e862

XenHat:
I would like to add that Telegram will also stay on IPV4 mode regardless of the user preference when behind an ISP-provided 6to4 tunnel (Videotron Ltee in my case). I have only tested this on Windows x64 so far.

nshopik:
@XenHat This is per design on most OSes, 2002::/16 prefix usually have lower precedence
https://tools.ietf.org/html/rfc6724#section-2.1 so it only selected to route within same prefix.

nshopik:
@john-preston This also somewhat improve security, as CGNAT become more common, IPv4 address from which somebody logged into your account could be same as yours and you could completely ignore that by mistake.

Thus connecting via IPv6 preferably give you better track your sessions

stek29:
Happens on OS X too by the way.

vitorgrs:
This still happens on Windows.

yangm97:
Still happens on macOS.

john-preston:
Yes, currently IPv4 is preferred to IPv6, because IPv4 shows itself better, is more stable from both our side and local provider side.

nshopik:
@john-preston Is there any practical reason why checkbox `try connecting through IPv6` unchecked by default then? 

> Yes, currently IPv4 is preferred to IPv6, because IPv4 shows itself better, is more stable from both our side and local provider side.

I would disagree on that, since iOS version and macOS native version prefer IPv6 and work flaweless

john-preston:
[...]

PR #27178, opened 2023-12-01
Prefer IPv6 over IPv4 for connection test

FlyGoat:
It is almost 2024 and there is no reason to perfer IPv4 over IPv6, especially in tdesktop user still need to toggle a switch to enable IPv6.

For tdesktop's connect test, if user does not have a routable IPv6 address, onDisconnect will be called promptly to fallback to another connection. If user's IPv6 connection is somehow filtering out connection to telegram's server, it will wait until kWaitForBetterTimeout and then fallback to another connection.

This implementation is actually in spirit of happy eyeballs recommandation.

Fixes: https://github.com/telegramdesktop/tdesktop/issues/2198

### DIFF
--- a/Telegram/SourceFiles/mtproto/session_private.cpp
+++ b/Telegram/SourceFiles/mtproto/session_private.cpp
@@ -185,7 +185,7 @@ void SessionPrivate::appendTestConnection(
 		const bytes::vector &protocolSecret) {
 	QWriteLocker lock(&_stateMutex);
 
-	const auto priority = (qthelp::is_ipv6(ip) ? 0 : 1)
+	const auto priority = (qthelp::is_ipv6(ip) ? 1 : 0)
 		+ (protocol == DcOptions::Variants::Tcp ? 1 : 0)
 		+ (protocolSecret.empty() ? 0 : 1);
 	_testConnections.push_back({


FlyGoat:
@anacondaq Well, I'm doing this PR because I'm living in a place where I need to pay extra for IPv4 connection, even if I pay extra for IPv4 connection is still worse than IPv6 one because of DS-Lite/CLAT overhead.

This is at least true for some parts of Europe and East Asia, all telecom infras are moving to IPv6.

I understand you don't like some design decisions made by IPv6, so do I. But we must move forward, because it's wide usage in real world had already proofed that it's not something born dead, it's the trend.

FlyGoat:
For this very specific use case, as we are doing connection test before actually use IPv6 connection, there is almost no chance that any regression can be introduced.

john-preston:
@FlyGoat I was told still that we really want to prefer IPv4 over IPv6 always (which means - always try to work through IPv4 if it works, use IPv6 connection otherwise).

I'd accept an experimental setting of "prefer IPv6" or even "use only IPv6" if someone PRd it, because after all that's for you, user, to decide in the end, but nothing in the normal connectivity options and certainly not by default.

chandr1000:
> > It is almost 2024 and there is no reason to perfer IPv4 over IPv6, especially in tdesktop user still need to toggle a switch to enable IPv6.
> 
> It's time to wake up and smell the rotting carcass of IPv6! This so-called "next-gen" protocol is nothing but a festering heap of bugs and failures that should have been buried alongside floppy disks and dial-up internet.
> 
> Let's cut the crap and lay down the harsh reality. IPv6 is a colossal waste of time, resources, and brain cells. You think it's all rainbows and unicorns because of its supposed abundance of addresses? Think again! IPv6 is a breeding ground for vulnerabilities and headaches that make your worst nightmares look like a walk in the park.
> 
> First off, let's talk about compatibility. Ever tried setting up IPv6 on your network? Good luck with that! It's like trying to fit a square peg into a round hole while blindfolded. Sure, some devices might support it, but don't be surprised when half of your gadgets start throwing tantrums and refusing to play nice.
> 
> And don't even get me started on security. IPv6 is like leaving the front door of your house wide open and inviting hackers to come in and have a field day. It's riddled with security holes big enough to drive a truck through, and no amount of patching can plug them all.
> 
> But wait, there's more! Let's talk about performance. Ever noticed how your internet connection slows to a crawl whenever you try to use IPv6? That's because it's about as efficient as a one-legged horse trying to win the Kentucky Derby. It chews up bandwidth like it's going out of style and leaves you wondering why you bothered upgrading in the first place.
> 
> And don't even get me started on compatibility with existing tools and software. Half the time, you'll spend more hours troubleshooting compatibility issues than actually getting any work done. It's like trying to teach a fish to ride a bicycle – frustrating and utterly pointless.
> 
> So here's the bottom line: IPv6 is a dead-end technology that should have been left to rot in the annals of history. Instead of wasting our time and energy trying to salvage this sinking ship, let's stick with good old IPv4 and build upon its solid foundation. It may not be perfect, but at least it doesn't suck as hard as IPv6.
> 
> My post not about "hey, ipv4 better, and bla bla bla". No, absolutely not. I have some experience in developing/patching/working with ipv6, this is some kind of nightmare. You have almost zero apps on your OS that working without using any hidden pitfalls that utilize ipv6 over ipv4, and there is a reason for that. IPv6 in my personal opinion - one of the most awful network technology, and it's born dead.
> 
> Try yourself to check different things here and there: https://github.com/search?q=ipv6+issues&type=issues
> 
>  ipv6.mp4

It's not
IPv6 solves me double CGNAT, or even triple CGNAT nightmare & worst CAPTCHA looping due to IPv4 being shared over 1000 ISP subscriber

Seems like you're having issue with your ISP

PR #27325, opened 2024-01-07
Add systemd unit to setup application scope and add hardening.

shanechko:
After enabling support for D-Bus activation  2dfe858327  application may started by the dbus service.
In this case application is placed into the incorrect cgroup scope inherited from dbus service.

This change provides systemd user unit for application to put application to the applications scope. Also, add some hardening to prevent unnecessary permissions.


ilya-fedin:
> In this case application is placed into the incorrect cgroup scope inherited from dbus service.

That's perhaps not worse than being placed in the cgroup from the application the launches tdesktop given that mime-apps-spec implementations usually don't create a cgroup. I'd rather encourage people to switch to dbus-broker if that's something important to them, instead of maintaining a unit file as a part of tdesktop.

xximwon (approved):

PR #27533, opened 2024-03-01
Fix libraries build (qt5)

agl-1984:
update new script location in qt repo
https://github.com/qt/qt5/commit/1971e8075d8f55256bfbd38b28165b47d3cf8cff

PR #27539, opened 2024-03-04
Fix Qt5 build on clean Windows

EricKotato:
This PR fixes building of Windows version on clean system.

## The problem

Perl is now used from MSYS instead of Strawberry, which leads to Qt 5 build failure, because it wasn't found in `PATH` environment variable:

Those who built TDesktop prior this change (including me) haven't spotted the difference, since Strawberry Perl is probably in their `PATH`.

## The solution

This can be solved by temporarily adding MSYS's `/usr/bin/` folder to system `PATH`, like in FFmpeg or libvpx build. `SET CHERE_INVOKING=enabled_from_arguments` and `SET MSYS2_PATH_TYPE=inherit` are not needed, since Perl script only initializes Qt repo, and is not used in building.

I've also tried to replace `perl` with `%THIRDPARTY_DIR%\\msys64\\usr\\bin\\perl.exe`, but this won't work since `dirname` at minimum is used inside the Perl script. Restoring PATH is also important because otherwise `link.exe` will be used from MSYS, not from Visual Studio, which will lead to Qt build failure.

## Attribution

@phagleb for screenshot and testing.

### DIFF
--- a/Telegram/build/prepare/prepare.py
+++ b/Telegram/build/prepare/prepare.py
@@ -1302,7 +1302,14 @@ if buildQt5:
     stage('qt_5_15_12', """
     git clone https://github.com/qt/qt5.git qt_5_15_12
     cd qt_5_15_12
+win:
+    SET PATH_BACKUP_=%PATH%
+    SET PATH=%ROOT_DIR%\\ThirdParty\\msys64\\usr\\bin;%PATH%
+common:
     perl init-repository --module-subset=qtbase,qtimageformats,qtsvg
+win:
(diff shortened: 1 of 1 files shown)


ilya-fedin:
The error from screenshot is clearly from perl so changing the PATH to help Windows find perl couldn't really fix this. The right fix is #27533.

EricKotato:
> The error on screenshot is clearly from perl so changing the PATH to help Windows find perl couldn't really fix this.

It did fixed it, but #27533 also did it and it's much simpler, so I'm closing this.

PR #27591, opened 2024-03-14
Using system font stack for Instant View

EricKotato:
To ensure that IV looks good enough on different platforms, the system font stack is used.

It's taken from https://systemfontstack.com/ with some tweaks:

- Tahoma is used before Arial, since Tahoma was widely used in Windows before Segoe UI was introduced (not sure if it matters, since Windows 7 includes Segoe UI anyway, and earlier OSes are not supported);
- Georgia font is used before Times New Roman as it was used in IV before this PR (and it looks better than Times New Roman);
- Cascadia Code is used before Consolas, since it's new Microsoft's font that is used primarily in Windows Terminal.


ilya-fedin:
>     * Cascadia Code is used before Consolas, since it's new Microsoft's font that is used primarily in Windows Terminal.

It's also the preferred monospace font on Windows in tdesktop

PR #27601, opened 2024-03-15
Upgrade dav1d, libavif, libde265, libheif

novomesk:
I updated these libraries because they contains various fixes for discovered bugs/vulnerabilities.

I tested local build on Windows.


john-preston:
Thanks!

PR #27674, opened 2024-03-31
Fix build Qt error on non-english windows os

wention:
Fixed the issue where Perl command running error on non-English systems caused QT compilation failure.


ilya-fedin:
That looks quite dangerous. There's perhaps a reason this code uses the script. Couldn't it be fixed rather than removed?

wention:
Qt document showed that the perl command could be replace by command "git submodule update"
> refer: https://wiki.qt.io/Building_Qt_5_from_Git

PR #27694, opened 2024-04-02
Fix tdesktop 4.15.6 build on OpenBSD

klemensn:
7.5-current with clang 16.0.6 and Qt6 fails, not entirely sure why...


klemensn:
Also needs https://github.com/desktop-app/lib_base/pull/247

PR #27737, opened 2024-04-10
Fixed python zipapp extension on windows in IsExecutableName

el-garro:
The correct python zipapp extension on windows is pyzw, this typo could lead to executing code in the client device without proper warning


john-preston:
Thank you!

HSSkyBoy (approved):

juli:
denylists alone is risky.

PR #27744, opened 2024-04-12
Update data_document_resolver.cpp

el-garro:
Added the LEXE and WLUA extensions to the IsExecutableName function for Windows. Those are registered by default on Windows devices with LUA installed


john-preston:
Thanks.

DeadMaX:
You might add MSI file extention too :)

john-preston:
@DeadMaX It's there, isn't it?

el-garro:
@DeadMaX Is there, the only one I can think of is .au3 but that's even more niche, given very few people even know what autoit is

PR #27841, opened 2024-05-02
fix: Incorrect package name

popey:
Noble has libvpx9, not vpx8.

https://packages.ubuntu.com/search?keywords=libvpx9

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -193,7 +193,7 @@ parts:
       - dpkg-dev
       - gtk-update-icon-cache
       - libglib2.0-bin
-      - libgtk-3-0
+      - libgtk-3-0t64
     stage-packages:
       - libc-bin
       - libgdk-pixbuf-2.0-0
@@ -211,7 +211,7 @@ parts:
           update-icon-caches "$theme"
         fi
       done
-      GTK_PATH=$PWD/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/gtk-3.0 /usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libgtk-3-0/gtk-query-immodules-3.0 > usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/gtk-3.0/3.0.0/immodules/immodules.cache
+      GTK_PATH=$PWD/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/gtk-3.0 /usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libgtk-3-0t64/gtk-query-immodules-3.0 > usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/gtk-3.0/3.0.0/immodules/immodules.cache
     after:
       - telegram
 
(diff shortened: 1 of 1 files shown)


Aokromes:
https://packages.debian.org/sid/libvpx8 exists.

popey:
Ok, but the snap builds inside Ubuntu usually, this change only affects the snap.

ilya-fedin:
@popey btw maybe you know how to enable riscv64 builds on launchpad and build.snapcraft.io? i see core22 and core24 have riscv64 builds but canonical builders don't even try to build tdesktop for it despite snapcraft.yml has no architecture restrictions.

I thought you have quit snap for flatpak but if you returned... maybe I can ask you more questions privately via Telegram (what's your username?)?

snap/snapcraft.yaml
```diff
@@ -193,7 +193,7 @@ parts:
       - dpkg-dev
       - gtk-update-icon-cache
       - libglib2.0-bin
-      - libgtk-3-0
+      - libgtk-3-0t64
```
ilya-fedin, line 196: That is not the only place libgtk-3-0 mentioned. There's also in the stage packages of qt part.

popey: That all built fine. It was only desktop-qt that barfed here.

ilya-fedin: using different packages for gtk smells like a bug. Either both must be `libgtk-3-0` or both must be `libgtk-3-0t64`.

ilya-fedin: > but I haven't test-built it like that

It's ok. I don't see how it could build like it's now given that there's no `libgtk-3-0` package for noble.

popey:
On the subject of riscv64:

`snapcraft remote-build --build-for=riscv64`

I just kicked one off as a test - which I _fully_ expect to fail. It's 'currently building' on launchpad right now. 

https://launchpad.net/~popey/popey-craft-remote-build/+snap/snapcraft-telegram-asahi-db4b99270893b00bc1a00a30fe11bf4e

Screenshot, just because if (when) the build fails the above link will likely break.

image

I'm popeydc on telegram. Feel free to dm me about this kind of thing. Happy to help. I have no idea how to make flatpaks :D

snap/snapcraft.yaml
popey:
```
$ apt-cache show libgtk-3-0t64 | grep Replaces
Replaces: libgtk-3-0 (<< 3.24.41-1.1)
```

Working as designed :D

ilya-fedin:
There's no sign of riscv64 here: https://launchpad.net/~telegram.desktop/+snap/telegram-desktop-master/

And preston was sharing screenshots of build.snapcraft.io where it doesn't even try to build for riscv64 just as well

popey:
I suspect that riscv64 isn't on by default for builds which use 'all' architectures. 
The [docs](https://snapcraft.io/docs/remote-build) say you can force it with `--build-for=amd64,arm64,armhf,ppcel64,s390x,riscv64`

Be aware the riscv64 builds take 3-5 times longer than other architectures. They're very slow - I suspect qemu - build bots. 

*however* there appears to be a bug here.

Found the bug: https://github.com/canonical/snapcraft/issues/4780

ilya-fedin:
> I suspect that riscv64 isn't on by default for builds which use 'all' architectures.

It's ok then... Thought maybe there's some architecture choice on launchpad/snapcraft.io as well...

PR #27850, opened 2024-05-04
Use jom build openssl on windows

GkvJwa:
After using openssl3, I found that compiling on windows is very slow. nmake still does not seem to support parallel make, but we can use jom to enable multi-core compilation to speed up

[refer to](https://github.com/openssl/openssl/issues/9931)


john-preston:
Thanks!

PR #27920, opened 2024-05-14
Fix typo in localimageloader.cpp

feriman:
Perhaps it meant "size limit", not "side limit".


23rd:
https://github.com/telegramdesktop/tdesktop/pull/27473

PR #27993, opened 2024-05-29
add option to export html and json in global export

mrbesen:
As started in #26199 and since [v4.9.6](https://github.com/telegramdesktop/tdesktop/releases/tag/v4.9.6) there is the option to export all data as html and json simultaneously.

This option is already implemented for exports of single chats. But the ui entry for the selection is still missing if all chats are going to be exported.


mrbesen:
@john-preston is there a problem with this change?

mrbesen:
> Was this tested on variety of options for the whole export, that isn't used in a single chat export? Like, contacts, userpics, additional info etc..

I tested it with "all options maxed" (except File sizes)  and the results look fine to me.

john-preston:
@mrbesen Ok. Forgot to merge into this beta, sorry.

john-preston:
Thanks!

PR #28048, opened 2024-06-14
Add xlsb to list of executable extensions

pandaninjas:
`xlsb` is Excel's binary file format. `xlsb` files support macros similar to `xlsm`. 

Since `xlsm` is included in the list of executable file extensions, `xlsb` probably should be included too.


john-preston:
Thanks!

PR #28129, opened 2024-07-04
Upgrade libjxl to v0.10.3

novomesk:


novomesk:
I tested only on Windows. Let's see if it builds correctly on MacOS too.

PR #28179, opened 2024-07-18
Added logic in Dockerfile and gen_dockerfile.py to support aarch64 security compilation fla…

looptc:
…gs equivalent to the actual ones.

Hi!

As we were talking in the issure, here is the PR.

I sent another PR to cmake_helpers because some .cmake files need also a modification.

Thanks!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -42,7 +42,7 @@
 	url = https://github.com/desktop-app/lib_storage.git
 [submodule "cmake"]
 	path = cmake
-	url = https://github.com/desktop-app/cmake_helpers.git
+	url = git@github.com:looptc/cmake_helpers.git
 [submodule "Telegram/ThirdParty/expected"]
 	path = Telegram/ThirdParty/expected
 	url = https://github.com/TartanLlama/expected
--- a/Telegram/build/docker/centos_env/Dockerfile
+++ b/Telegram/build/docker/centos_env/Dockerfile
@@ -745,6 +745,14 @@ RUN git clone -b v2023.06.01 --depth=1 https://chromium.googlesource.com/breakpa
 	&& cd .. \
 	&& rm -rf breakpad
 
+FROM builder AS openh264 
+RUN git clone https://github.com/cisco/openh264.git \
+	&& cd openh264 \
+	&& make -j$(nproc) OS=linux ARCH={{ ARCH }} \
+	&& make DESTDIR="{{ LibrariesPath }}/openh264-cache" install \
+	&& cd .. \
+	&& rm -rf openh264
+
 FROM builder AS webrtc
 COPY --link --from=opus {{ LibrariesPath }}/opus-cache /
 COPY --link --from=libvpx {{ LibrariesPath }}/libvpx-cache /
@@ -753,6 +761,7 @@ COPY --link --from=ffmpeg {{ LibrariesPath }}/ffmpeg-cache /
 COPY --link --from=openssl {{ LibrariesPath }}/openssl-cache /
 COPY --link --from=libXtst {{ LibrariesPath }}/libXtst-cache /
 COPY --link --from=pipewire {{ LibrariesPath }}/pipewire-cache /
+COPY --link --from=openh264 {{ LibrariesPath }}/openh264-cache /
 
 # Shallow clone on a specific commit.
 RUN git init tg_owt \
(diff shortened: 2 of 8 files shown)


.gitmodules
```diff
@@ -42,7 +42,7 @@
 	url = https://github.com/desktop-app/lib_storage.git
 [submodule "cmake"]
 	path = cmake
-	url = https://github.com/desktop-app/cmake_helpers.git
+	url = git@github.com:looptc/cmake_helpers.git
```
ilya-fedin, line 45: This doesn't look right

Telegram/build/docker/centos_env/Dockerfile
```diff
@@ -2,7 +2,12 @@
 {%- set GIT_FREEDESKTOP = GIT ~ "/gitlab-freedesktop-mirrors" -%}
 {%- set QT = "6.7.2" -%}
 {%- set QT_TAG = "v" ~ QT -%}
+{%- set CFLAGS = "-O3 -pipe -fPIC -fno-omit-frame-pointer -fstack-protector-all -fstack-clash-protection -fcf-protection -DNDEBUG -D_FORTIFY_SOURCE=3 -D_GLIBCXX_ASSERTIONS" -%}
 {%- set CFLAGS_DEBUG = "-g -pipe -fPIC -fstack-protector-all -fstack-clash-protection -fcf-protection -D_GLIBCXX_ASSERTIONS" -%}
+{%- if ARCH == "aarch64" -%}
+    {%- set CFLAGS = CFLAGS | replace("-fcf-protection", "-msign-return-address=all -mbranch-protection=standard") -%}
```
ilya-fedin, line 8: 1. Isn't -msign-return-address deprecated by -mbranch-protection?
2. Isn't `-mbranch-protection=standard` the default?

Telegram/build/docker/centos_env/gen_dockerfile.py
```diff
@@ -3,11 +3,14 @@
...
 from jinja2 import Environment, FileSystemLoader
 
+import platform
+
 def checkEnv(envName, defaultValue):
     return bool(len(environ[envName])) if envName in environ else defaultValue
 
 def main():
     print(Environment(loader=FileSystemLoader(dirname(__file__))).get_template("Dockerfile").render(
+	ARCH=platform.machine(),
```
ilya-fedin, line 13: Can't we do it in a similar fashion to target_compile_options_if_exists? Try `echo 'int main() {return 0;}' | g++ -o /dev/null -x c++ -fcf-protection -`, if it fails for you then you can use it to check whether the argument is supported.

Telegram/build/docker/centos_env/Dockerfile
looptc: Okay, so I remove the flags because they're by default on.

.gitmodules
looptc: Sorry, I commit the mistake to share the pull request from the same repository where I am using my personal fork. Let me fix that.

Telegram/build/docker/centos_env/gen_dockerfile.py
looptc: I have been pulling submodules from master and dev branches, and webrtc is requesting openh264, which needs make OS=linux ARCH=aarch64 to compile. I can remove this, leaving only a removement of the flag for the compiler following your fashion, but maybe in a future we need to have the keyword 'aarch64' in some place accesible for jinja2. What do you think?

ilya-fedin: Let's don't solve problems we haven't encountered yet

looptc: Okay, give me some time to fix this PR.

looptc: I've created and exported a function to later user with jinja2, it runs the command you suggested with the option one by one and sets a final variable. Is this way the right one?

Telegram/build/docker/centos_env/Dockerfile
looptc: There's a problem if we use -mbranch-protection=pac-ret+leaf with dav1d, it doesn't support leaf functions. May we use an specific compiling options for dav1d, for example -mbranch-protection=standard that works great, or may we use 'standard' for the whole project?

Telegram/build/docker/centos_env/gen_dockerfile.py
ilya-fedin: Looks like you haven't pushed it? I don't see any changes

looptc: Sorry, I sent to review but I just wanted to send a simple comment. I'm preparing the docker image, it's finishing soon, then I send changes when it works.

PR #28321, opened 2024-08-26
chore: update mtproto_dc_options.cpp

eltociear:
overriden -> overridden


john-preston:

PR #28351, opened 2024-09-04
Changed dev name in org.telegram.desktop.metainfo.xml

Viliansh:
In flathub, and it's front-ends like KDE's Discover or GNOME Software, Telegram shows "By John Preston", and some users were hesitant to install it afraid that was "fake", so i changed it to Telegram, this would also make it consistent with snap, thanks!

### DIFF
--- a/lib/xdg/org.telegram.desktop.metainfo.xml
+++ b/lib/xdg/org.telegram.desktop.metainfo.xml
@@ -24,7 +24,7 @@
         <category>InstantMessaging</category>
     </categories>
     <update_contact>johnprestonmail_at_gmail.com</update_contact>
-    <developer_name>John Preston</developer_name>
+    <developer_name>Telegram FZ-LLC</developer_name>
     <url type="homepage">https://desktop.telegram.org/</url>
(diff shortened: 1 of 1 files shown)


ilya-fedin:
Maybe it's also a good idea to remove the update_contact tag as @john-preston doesn't actually read it (@john-preston right?)... It's optional anyway.

@Viliansh can you rebase the PR to get rid of that merge branch commit?

lib/xdg/org.telegram.desktop.metainfo.xml
```diff
@@ -24,7 +24,7 @@
         <category>InstantMessaging</category>
     </categories>
     <update_contact>johnprestonmail_at_gmail.com</update_contact>
-    <developer_name>John Preston</developer_name>
+    <developer_name>Telegram FZ-LLC</developer_name>
```
ilya-fedin, line 27:
```suggestion
    <developer id="org.telegram" name="Telegram FZ-LLC"/>
```

ilya-fedin:
Nice. I was thinking what ID to set for John Preston for a long time but this idea solves it very easily.

ilya-fedin:
> yeah, anything else?

wait preston

PR #28395, opened 2024-09-11
Fix restrict user in recent actions

c0re100:
`if (const auto since = participant.restrictedSince())`
Cause `since` is always false when user is not restricted,
[restrictedSince() return 0](https://github.com/telegramdesktop/tdesktop/blob/dev/Telegram/SourceFiles/api/api_chat_participants.cpp#L372)
then EditRestrictedBox wouldn't pop up.

### DIFF
--- a/Telegram/SourceFiles/history/admin_log/history_admin_log_inner.cpp
+++ b/Telegram/SourceFiles/history/admin_log/history_admin_log_inner.cpp
@@ -1572,7 +1572,8 @@ void InnerWidget::suggestRestrictParticipant(
 				if (participant.type() == Type::Creator
 					|| participant.type() == Type::Admin) {
 					editRestrictions(true, {}, nullptr, 0);
-				} else if (const auto since = participant.restrictedSince()) {
+				} else {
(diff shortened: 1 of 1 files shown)


23rd:
Not `else` but `else if (participant.type() == Type::Restricted || participant.type() == Type::Banned)` (see 686e9643ad2fdaa31e261e0704cc2ae3b7d99dcb).

But still you are right.

Telegram/SourceFiles/history/admin_log/history_admin_log_inner.cpp
```diff
@@ -1572,7 +1572,8 @@ void InnerWidget::suggestRestrictParticipant(
 				if (participant.type() == Type::Creator
 					|| participant.type() == Type::Admin) {
 					editRestrictions(true, {}, nullptr, 0);
-				} else if (const auto since = participant.restrictedSince()) {
+				} else {
```
23rd, line 1575:
```suggestion
				} else if (participant.type() == Type::Restricted
					|| participant.type() == Type::Banned) {
```

PR #28458, opened 2024-09-24
Update prepare.py

Donovoi:
different capitalization for different versions of OS


ilya-fedin:
Doesn't that make all following `os.environ['Platform']` usages invalid?

john-preston:

PR #28524, opened 2024-10-13
Linux: fix notification urgency

lenemter:
Telegram currently stores notifications data in GLib.Notification which uses Notification portal under the hood. But sends the data from it to notification server on `org.freedesktop.Notifications`.

These two methods use different specifications. Portal uses [this one](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Notification.html#) and `org.freedesktop.Notifications` [this one](https://specifications.freedesktop.org/notification-spec/1.2/index.html). And in conversion from portal version to fd.o urgency HIGH with byte number 2 becomes URGENT priority with the same byte number 2.

PS: I haven't tested these changes, because I don't know how to use make.

### DIFF
--- a/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
@@ -249,8 +249,8 @@ bool NotificationData::init(
 			Gio::ThemedIcon::new_(base::IconName().toStdString()));
 
 		// for chat messages, according to
-		// https://docs.gtk.org/gio/enum.NotificationPriority.html
-		_notification.set_priority(Gio::NotificationPriority::HIGH_);
+		// https://specifications.freedesktop.org/notification-spec/1.2/urgency-levels.html
+		_notification.set_priority(Gio::NotificationPriority::NORMAL_);
 
 		// glib 2.70+, we keep glib 2.56+ compatibility
 		static const auto set_category = [] {


ilya-fedin:
That's really a bug in your portal notification backend and you should make a PR there instead. The current priority is right according to the API documentation and I don't think we should change it unless the documentation changes

I.e. if you want to change the code in tdesktop, please make a PR to change GIO documentation first. If they would accept it, it could be considered.

ilya-fedin:
E.g. glib itself has the right conversation code: https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/gfdonotificationbackend.c#L269

Telegram/SourceFiles/platform/linux/notifications_manager_linux.cpp
```diff
@@ -249,8 +249,8 @@ bool NotificationData::init(
 			Gio::ThemedIcon::new_(base::IconName().toStdString()));
 
 		// for chat messages, according to
-		// https://docs.gtk.org/gio/enum.NotificationPriority.html
-		_notification.set_priority(Gio::NotificationPriority::HIGH_);
+		// https://specifications.freedesktop.org/notification-spec/1.2/urgency-levels.html
```
ilya-fedin, line 252: Using freedesktop urgency levels with GIO APIs is wrong. Code using GIO API should use GIO notification priority. For freedesktop urgency to be used, freedesktop API has to be used directly. This change is completely wrong.

lenemter:
Found out the issue is not Telegram, but xdg-desktop-portal-gtk instead. The fix was introduced in https://github.com/flatpak/xdg-desktop-portal-gtk/commit/5caa26dcc864b77292114acf2fd2ae68e1b53735 but hasn't got into Ubuntu 24.04 yet.

ilya-fedin:
> but hasn't got into Ubuntu 24.04 yet

I don't think it will get into it at all unless someone would create a bug report on ubuntu bugtracker asking to backport the commit.

PR #28567, opened 2024-10-23
Build: Speed up third party libraries cloning

difhel:
When I tried to build Telegram Desktop from source code, I realized it was running pretty darn slow.

When I decided to look into it, I suddenly found out that the preparation script clones the entire git repository index of each third party library used, not just the last commit we need to build. So I looked at which stages `tdesktop` uses just a branch (without explicitly checkingout to a specific commit) and added the `--depth 1` parameter. This greatly reduces build preparation time, as well as disk space and traffic used, since some of the libraries used are very large.

### DIFF
--- a/Telegram/build/prepare/prepare.py
+++ b/Telegram/build/prepare/prepare.py
@@ -61,6 +61,7 @@ optionsList = [
     'qt6',
     'skip-release',
     'build-stackwalk',
+    'no-deps-git-history'
 ]
 options = []
 runCommand = []
@@ -326,6 +327,17 @@ def run(commands):
     else:
         return subprocess.run("set -e\n" + commands, shell=True, env=modifiedEnv).returncode == 0
 
+no_deps_git_history = "no-deps-git-history" in options
+if no_deps_git_history:
+    print("Notice: Some dependency repositories will be downloaded without full history (via git clone --depth 1) to reduce time and disk space.")
+
+def git(command, args):
+    result = f"git {command}"
+    if no_deps_git_history:
+        result += " --depth 1"
+    result = result + " " + args
+    return result
+
 # Thanks https://stackoverflow.com/a/510364
 class _Getch:
     """Gets a single character from standard input.  Does not echo to the
@@ -500,9 +512,9 @@ win:
     del jom.zip
 """, 'ThirdParty')
 
-stage('depot_tools', """
+stage('depot_tools', f"""
 mac:
-    git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
+    {git('clone', 'https://chromium.googlesource.com/chromium/tools/depot_tools.git')}
     cd depot_tools
     ./update_depot_tools
 """, 'ThirdParty')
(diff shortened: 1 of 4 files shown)


john-preston:
If this is to be added it should be non-default and somehow configurable, because I really enjoy full repository histories in the dependencies.

docs/building-linux.md
```diff
@@ -15,6 +15,10 @@ Install [poetry](https://python-poetry.org), go to ***BuildPath*** and run
     git clone --recursive https://github.com/telegramdesktop/tdesktop.git
     ./tdesktop/Telegram/build/prepare/linux.sh
 
+You might not need Telegram Desktop's git dependency history, in which case you can run this command and save disk space and speed up cloning quite a bit:
+
+    ./tdesktop/Telegram/build/prepare/linux.sh no-deps-git-history
```
ilya-fedin, line 20: This won't work, look at linux.sh

docs/building-win.md
```diff
@@ -30,6 +30,10 @@ Open **x86 Native Tools Command Prompt for VS 2022.bat**, go to ***BuildPath***
     git clone --recursive https://github.com/telegramdesktop/tdesktop.git
     tdesktop\Telegram\build\prepare\win.bat
 
+You might not need Telegram Desktop's git dependency history, in which case you can run this command and save disk space and speed up cloning quite a bit:
```
ilya-fedin, line 33: Maybe make it clear somehow that it has to be done instead of the last command? Given how much people ask stupid questions about build, I won't be surprised if people end up calling win.bat two times.

ilya-fedin:
Maybe squash 1st, 2nd and 4th commits?

docs/building-linux.md
difhel: увы

docs/building-win.md
```diff
@@ -30,6 +30,11 @@ Open **x86 Native Tools Command Prompt for VS 2022.bat**, go to ***BuildPath***
     git clone --recursive https://github.com/telegramdesktop/tdesktop.git
     tdesktop\Telegram\build\prepare\win.bat
 
+> [!IMPORTANT]
+> You might not need Telegram Desktop's git dependency history, in which case you can run this command **instead of the one above** and save disk space and speed up cloning quite a bit:
```
ilya-fedin, line 34: This seems clearer (imo)
```suggestion
> You might not need Telegram Desktop's git dependency history, in which case you can run this command **instead of the last one** and save disk space and speed up cloning quite a bit:
```

ilya-fedin: Can you apply to other files?

Telegram/build/prepare/prepare.py
```diff
@@ -326,6 +327,17 @@ def run(commands):
     else:
         return subprocess.run("set -e\n" + commands, shell=True, env=modifiedEnv).returncode == 0
 
+no_deps_git_history = "no-deps-git-history" in options
+if no_deps_git_history:
+    print("Notice: Some dependency repositories will be downloaded without full history (via git clone --depth 1) to reduce time and disk space.")
+
+def git(command, args):
```
john-preston, line 334: This does what I asked, but not the way other things are done here. It looks to me that more consistent with how the script already works would be another approach:
- if 'no-deps-git-history' is passed as an option you add something like 'DEPTH' -> '--depth 1' to the 'environment' map, otherwise set it to an empty string
- in all places necessary just add $DEPTH in the git command so that the '--depth 1' is added if necessary..

If this way it works as expected, then I'd like that approach to be implemented. If somehow I missed something and this doesn't work then we'll proceed with the one you proposed.

john-preston (changes requested):

ilya-fedin:
Conflicts

PR #28580, opened 2024-10-27
refactor: construct QString with QStringLiteral to improve performance

Integral-Tech:


ilya-fedin:
Please use https://github.com/desktop-app/lib_base/blob/21d1ac8bfcca03f67d7f6df75e265cd5597dc101/base/basic_types.h#L162-L174 literals instead of QStringLiteral

Telegram/SourceFiles/api/api_premium_option.cpp
```diff
@@ -27,7 +27,7 @@ Data::PremiumSubscriptionOption CreateSubscriptionOption(
 	return {
 		.duration = Ui::FormatTTL(months * 86400 * 31),
 		.discount = discount
-			? QString::fromUtf8("\xe2\x88\x92%1%").arg(discount)
+			? QStringLiteral("\xe2\x88\x92%1%").arg(discount)
```
ilya-fedin, line 30: This change (as well as all other `QString::fromUtf8 -> QStringLiteral` ones) is suspicious. Have you checked that the data is not corrupted now? I'm pretty sure QStringLiteral interprets the data as UTF-16, not UTF-8.

john-preston:

In all new code I try using the u"..."_q literal (it's similar to QStringLiteral macro, but with better readability), but replacing all literals in the code base should be done really careful, for example ::fromUtf8() should not be replaced that way.

Integral-Tech:
> Sorry, but I've decided not to merge any pull requests that don't affect end user experience in a positive way (except for some exceptions on an individual basis).
> 
> In all new code I try using the u"..."_q literal (it's similar to QStringLiteral macro, but with better readability), but replacing all literals in the code base should be done really careful, for example ::fromUtf8() should not be replaced that way.

What about replacing string literals with ""_q, except for `QString::fromUtf8()`?

Issue #26735, opened 2023-09-01
Command list is not displayed (depending)


DO97:
### Steps to reproduce

Some of my groups don't show the command list of bots. It doesn't matter the bot, because even if it's the same, some groups have it and some not.

It seems like my most recently created groups are not affected, but I'm not 100% sure.

Do you know why? How can I fix this bug? It's been a year now, but it's still present.

### Expected behaviour

Command list should work for any group.

### Actual behaviour

Command list icon, when clicked, doesn't show up anything. It only generates a "/".

### Operating system

Windows 10

### Version of Telegram Desktop

4.9.4

### Installation source

Static binary from official website

sufficientforce:
I also see this with certain groups.

The bot's command list shows up when typing '/' in other groups in which the bot has been added...

...AND it shows up when typing '/' in the mobile app for the problematic group...

...but for the problematic group on Windows desktop telegram, typing slash does not show the bot's commands.

sam0324sam:
same quetion.....please fix it.

DO97:
It Is like this for over a year. I don't know how they didn't notice it yet.

Issue #25027, opened 2022-09-04
[Question] Why aren't Bot commands displayed in my group?


moshaoli688:
Why aren't Bot commands displayed in my group？？
It is normal for the same bots to be added to other groups. However, the BOT commands never show up properly in my group。
However, using the third party client UNIGRAM can display the command of BOT in my group normally.
Mobile APP can also display BOT commands normally. But official Telegram customers cannot display bot commands.

bot id:
           @FFQMiaokoBot
           @FFQAntiChannelSpamBot
           @MissRose_bot 
           @ffqMiaoBot
telegram:
![image](https://user-images.githubusercontent.com/85819540/188311406-e2ffacf9-7b40-4782-a13f-2acfc55f86e5.png)
unigram:
![image](https://user-images.githubusercontent.com/85819540/188311448-c1568e06-a828-4c67-bfb2-dec2559d260a.png)

Senderman:
Reproducible in Telegram Desktop 4.6.2 Linux

aNNiMON:
This issue is reproducible on 4.6.2

We have two groups. In the first group we have 5 bots with commands (let's say bot A, bot B, bot C, bot D and bot E), but tdesktop shows only commands from 4 of them. In the second group we have 7 bots: bot A and 6 another bots with commands. Tdesktop shows command from these 6 bots, but not for bot A. However, in the first group bot A commands are showing in the list.
 
Then we tried the following steps in the second group:

№ | Action | Result | Expected? 
--- | --- | --- | -- 
1 | Remove all bots | Commands from 6 bots are still exist in the list (list is not refreshing after removing bots) | ⛔
2 | Restart tdesktop | Empty commands list | ✅ 
3 | Add bot A | Bot A commands are showing | ✅
4 | Add 3 (of 6 kicked) bots to the group again | Commands from 3 bots are showing, but Bot A commands aren't | ⛔

aNNiMON:
Still not fixed

aNNiMON:
The issue is still present in 4.14.13

sam0324sam:
The issue is still present in 4.15 x64

andrey-helldar:
I've run into this problem too. The same bot in some groups shows commands and in some groups it does not.

moshaoli688:
This problem seems to exist more on the desktop client, and the mobile client seems to be fine at present.
So far, the bot command in my windows client cannot be fully displayed.

andrey-helldar:
[...]

PR #28631, opened 2024-11-09
Fix bug in ChatBotCommands::update when commands empty

ziplantil:
Certain bots do not show commands in particular groups at random. This is a known issue, see e.g. #25027, #26735.

The issue is that we try to assign a value through an iterator even after we have called `remove`. This will cause the iterator to point to some other element, and then we copy an empty list of commands into that other list. This means that the bot list of some other bot, that is the bot with the next higher user ID, gets cleared instead.

This fixes the bug by making sure that we do not try to assign anything after calling `remove`.

### DIFF
--- a/Telegram/SourceFiles/data/data_peer_bot_commands.cpp
+++ b/Telegram/SourceFiles/data/data_peer_bot_commands.cpp
@@ -18,10 +18,12 @@ ChatBotCommands::Changed ChatBotCommands::update(
 	} else {
 		for (const auto &commands : list) {
 			auto &value = operator[](commands.userId);
-			changed |= commands.commands.empty()
-				? remove(commands.userId)
-				: !ranges::equal(value, commands.commands);
-			value = commands.commands;
+			if (commands.commands.empty()) {
+				changed |= remove(commands.userId);
+			} else {
+				changed |= !ranges::equal(value, commands.commands);
+				value = commands.commands;
+			}
 		}
 	}
 	return changed;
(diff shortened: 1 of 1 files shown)


Telegram/SourceFiles/data/data_peer_bot_commands.cpp
```diff
@@ -18,10 +18,12 @@ ChatBotCommands::Changed ChatBotCommands::update(
 	} else {
 		for (const auto &commands : list) {
 			auto &value = operator[](commands.userId);
```
23rd, line 20:
```suggestion
```

Telegram/SourceFiles/data/data_peer_bot_commands.cpp
```diff
@@ -18,10 +18,12 @@ ChatBotCommands::Changed ChatBotCommands::update(
...
-			changed |= commands.commands.empty()
-				? remove(commands.userId)
-				: !ranges::equal(value, commands.commands);
-			value = commands.commands;
+			if (commands.commands.empty()) {
+				changed |= remove(commands.userId);
+			} else {
+				changed |= !ranges::equal(value, commands.commands);
+				value = commands.commands;
```
23rd, line 25:
```suggestion
				auto &value = operator[](commands.userId);
				const auto isEqual = ranges::equal(value, commands.commands);
				changed |= !isEqual;
				if (!isEqual) {
					value = commands.commands;
				}
```

john-preston:
Thanks!

PR #28693, opened 2024-11-24
Update kimageformats submodule

novomesk:
There is an important fix in HEIF plug-in and minor improvements in AVIF plug-in.


john-preston:
Thanks!

PR #28718, opened 2024-11-30
Issue 28717 suggestion

Maksimonk:
New elements should be pushed in the beginning of _items


john-preston:
This change breaks the thumbnails of all groups, changing the order to the opposite. For example of photo albums.

ilya-fedin:
No review response in 2 years

PR #28758, opened 2024-12-13
Update kimageformats, libheif, libjxl

novomesk:
kimageformats - contains tiny fix for animated JXL images

libheif - newer version supports images from latest iPhones. Note: 1.18.2 is not newest today, but I consider it better choice now than v1.19.5, which has lot of new features and maybe needs more time to test.

libjxl - update with relatively important fixes/improvements (CVE-2024-11498, avoiding abort in release build).

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -254,7 +254,7 @@ parts:
   libjxl:
     source: https://github.com/libjxl/libjxl.git
     source-depth: 1
-    source-tag: v0.10.3
+    source-tag: v0.11.1
     plugin: cmake
     build-environment:
       - LDFLAGS: ${LDFLAGS:+$LDFLAGS} -s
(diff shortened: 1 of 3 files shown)


snap/snapcraft.yaml
```diff
@@ -276,7 +276,6 @@ parts:
       - -DJPEGXL_ENABLE_SKCMS=OFF
     override-pull: |
       craftctl default
-      git apply $CRAFT_STAGE/patches/libjxl.patch
```
ilya-fedin, line 279: Entire override-pull section has no sense after that

novomesk: So you are suggesting that override-pull: can be removed, right?

ilya-fedin: Yes

PR #28815, opened 2025-01-01
Fix notifications when passed through with a cursor too quickly

### LINKED ISSUE #28811: Crossing a notification with a cursor too fast makes it freeze
### Steps to reproduce

1. Wait for a notification
2. Pass it quickly with the cursor (i.e., enter and exit the box)

### Expected behaviour

After a timeout, the notification fades out

### Actual behaviour

The notification hangs there, until cleared with another action (e.g., mouse click)

Demo:

https://github.com/user-attachments/assets/8e8d2e7b-1fa1-439c-bdce-ad0d04143915

### Operating system

Arch Linux, i3 version 4.24

5.9.2

### Installation source

Other (unofficial) source

### Crash ID

_No response_

### Logs

_No response_

kolayne:
In this PR, for `Window::Notification::Default::internal::Widget` I:
- Rename `.isShowing()` to `.isFadingIn()` to clarify its behavior;
- Add a call to `Ui::Animations::Simple::stop` before `::start` in notification hiding animation.

~~Is blocked by desktop-app/lib_ui#254~~. Fixes #28811.

Kindly, see the commit messages for details.

### DIFF
--- a/Telegram/SourceFiles/window/notifications_manager_default.cpp
+++ b/Telegram/SourceFiles/window/notifications_manager_default.cpp
@@ -188,16 +188,10 @@ void Manager::checkLastInput() {
 
 void Manager::startAllHiding() {
 	if (!hasReplyingNotification()) {
-		int notHidingCount = 0;
 		for (const auto &notification : _notifications) {
-			if (notification->isShowing()) {
-				++notHidingCount;
-			} else {
-				notification->startHiding();
-			}
+			notification->startHiding();
 		}
-		notHidingCount += _queuedNotifications.size();
-		if (_hideAll && notHidingCount < 2) {
+		if (_hideAll && _queuedNotifications.size() < 2) {
 			_hideAll->startHiding();
 		}
 	}
--- a/Telegram/SourceFiles/window/notifications_manager_default.cpp
+++ b/Telegram/SourceFiles/window/notifications_manager_default.cpp
@@ -549,6 +549,10 @@ void Widget::hideStop() {
 
 void Widget::hideAnimated(float64 duration, const anim::transition &func) {
 	_hiding = true;
+	// Stop the previous animation so as to make sure that the notification
+	// is fully restored before hiding it again.
+	// Relates to https://github.com/telegramdesktop/tdesktop/issues/28811.
+	_a_opacity.stop();
 	_a_opacity.start([this] { opacityAnimationCallback(); }, 1., 0., duration, func);
 }
 
--- a/Telegram/SourceFiles/window/notifications_manager_default.h
+++ b/Telegram/SourceFiles/window/notifications_manager_default.h
@@ -143,7 +143,7 @@ public:
 		int shift,
 		Direction shiftDirection);
 
-	bool isShowing() const {
+	bool isFadingIn() const {
 		return _a_opacity.animating() && !_hiding;
 	}
 


Telegram/lib_ui
kolayne, line 1: I am not sure how I'm supposed to handle the lib_ui submodule in my commits (both of which require corresponding changes in lib_ui). Please, tell me if/how I should fix it.

kolayne:
By the way, I believe that with these changes, in the `Widget` constructor there is no need to call `setWindowOpacity(0.);` anymore, as it will anyway start the animation at the end. Is that right? Should I remove it?

https://github.com/telegramdesktop/tdesktop/blob/4505a2bf2dbb73185b4bb8b18b2aab721765bb7e/Telegram/SourceFiles/window/notifications_manager_default.cpp#L472-L497

ilya-fedin:
You say in one of commit messages:
```
Previously, `Window::Notifications::Default::internal::Widget` was using
`Ui::Animations::Basic`'s `.animating()` method to tell if a
notification was in the showing state. That is incorrect: a notification
may be in the animating state whether it is fading out (i.e., not
showing) or fading in (i.e., to be shown).
```

But that's not true, here's the function in question:
```
	bool isShowing() const {
		return _a_opacity.animating() && !_hiding;
	}
```
It does check exactly hiding, it can't return `true` on showing.

ilya-fedin:
I also suspect the issue you mark here as being fixed couldn't be fixed by changes in this PR, as they likely stem from the same problem of broken QWidget::leaveEvent you try to work around in #28816?

Telegram/SourceFiles/window/notifications_manager_default.cpp
```diff
@@ -543,12 +543,17 @@ void Widget::hideStop() {
 	if (_hiding) {
 		_hiding = false;
 		_hidingDelayed = {};
-		_a_opacity.start([this] { opacityAnimationCallback(); }, 0., 1., st::notifyFastAnim);
+		// Using `.change` instead of `.start` because we want to start
+		// animation from the current opacity level, not blink it.
```
ilya-fedin, line 547: I see Animations::start already has some handling of unfinished animation?

kolayne:
> It does check exactly !_hiding, it can't return true on hiding.

Yes, it can:

`_hiding` is set to true inside `Widget::hideSlow`, which is called by `Notification::startHiding`, which is called by the manager (`Manager::startAllHiding`) _only if_ `notification->isShowing()` is false. So, the order is the other way around: first, the manager will check if the notification is showing (which involves checking the value of `_hiding`), and then conditionally call `startHiding` to change the value of `_hiding`. I feel like the member `_hiding` is intended to just avoid double hiding notifications.

Relevant code:
First, the manager checks the notification status:
```cpp
void Manager::startAllHiding() {
    if (!hasReplyingNotification()) {
        int notHidingCount = 0;
        for (const auto &notification : _notifications) {
            if (notification->isShowing()) {
                ++notHidingCount;
            } else {
                notification->startHiding();
            }
        }
        // ...
```
and that involves the `_hiding` member:
```cpp
	bool isShowing() const {
		return _a_opacity.animating() && !_hiding;
	}
```
then the manager conditionally calls `notification->startHiding()`:
```cpp
void Notification::startHiding() {
    if (!_history) return;
    hideSlow();
}
```
which calls `Widget::hideSlow()`:
```cpp
void Widget::hideSlow() {
    if (anim::Disabled()) {
        _hiding = true;
        base::call_delayed(
            st::notifySlowHide,
            this,
            [=, guard = _hidingDelayed.make_guard()] {
                if (guard && _hiding) { 
                    hideFast();
                }
            });
    } else {
        hideAnimated(st::notifySlowHide, anim::easeInCirc);
    }
}
```
and that's where `_hiding` is set, either directly, or through `hideAnimated`:
```cpp
void Widget::hideAnimated(float64 duration, const anim::transition &func) {
    _hiding = true;
    _a_opacity.start([this] { opacityAnimationCallback(); }, 1., 0., duration, func);
}
```

kolayne:
> I also suspect the issue you mark here as being fixed couldn't be fixed by changes in this PR, as they likely stem from the same problem of broken QWidget::leaveEvent you try to work around in #28816?

As I understand, you are saying that (1) this is the same issue as in #28816 and (2) this PR does not fix it?

1. This is definitely a different issue. By either adding debug output in the code or running a debug build of Telegram in `gdb`, one can make sure that event hooks are generated properly here. The problem here is the incorrect behavior of `isShowing()`, see the explanation above. Using the same methods, one can verify that in the scenario in #28816 the hook is _not_ triggered, so these are two very different issues.
If that would help, let me know and I will design a demo to show clearly that the hook is triggered and the decision made by `Manager::hideAllAnimating()` is incorrect.
2. I can reproduce the issue every time with the official build. I couldn't ever reproduce it with the new build. This, and the explanation in the previous message, are the reasons I believe this PR fixes the problem.

Telegram/SourceFiles/window/notifications_manager_default.cpp
kolayne: Unfortunately, the lib_ui was not documented, so I can't be sure about that, but I assumed that:
- `Ui::Animations::Simple::start`, when called on an ongoing animation, should reset it completely, set opacity to `from`, and go from there to `to`
- `Ui::Animations::Simple::change`, when called on an ongoing animation, should not reset the `from`: it starts from where the previous animation got, and continues from there.

If my assumptions are correct, I think, you should agree that here we need `.change()` (i.e., restore the notification from whatever state it was in) rather than `.start()` (blink the notification by first setting its opacity to zero and then restoring it).

If my assumptions are wrong, I kindly ask you to explain the intended use of the `.start()` and `.change()` methods of an animation in lib_ui.

ilya-fedin:
> `_hiding` is set to true inside `Widget::hideSlow`, which is called by `Notification::startHiding`, which is called by the manager (`Manager::startAllHiding`) _only if_ `notification->isShowing()` is false. So, the order is the other way around: first, the manager will check if the notification is showing (which involves checking the value of `_hiding`), and then conditionally call `startHiding` to change the value of `_hiding`.

I'm not sure what's wrong here? The hiding animation gets started only after _hiding is set to true. I don't see a moment when it could hide without _hiding being set.

ilya-fedin:
It's also weird that after the change it's `_a_opacity.fadingOut() && !_hiding`, i.e. "hiding and not hiding". Perhaps it's meant to be just `_a_opacity.fadingIn()`?

ilya-fedin:
I believe what you try to fix here needs a method that would do
```
	if (!_waitingForInput) {
		_hideTimer.start(st::notifyWaitLongHide);
	}
```
and then call it when isShowing() returns true. With no other changes.

kolayne:
> It's also weird that it's `_a_opacity.fadingOut() && !_hiding` after the change, i.e. "hiding and not hiding".

Oh, you're right. I just realized I misunderstood the purpose of `Notification::isShowing`. I agree, `_a_opacity.animating() && _hiding` is the condition for fading in. My solution is incorrect.

But then, @ilya-fedin, could you help me understand the following piece of code:
```cpp
void Manager::startAllHiding() {
    if (!hasReplyingNotification()) {
        int notHidingCount = 0;
        for (const auto &notification : _notifications) {
            if (notification->isShowing()) {
                ++notHidingCount;
            } else {
                notification->startHiding();
            }
        }
// ...
```

What is the purpose of the inner condition? Why does the manager refuse to hide a notification that is fading in? Seems like it is exactly what is causing the issue: if the cursor leaves the notification too quickly, `leaveEventHook` calls `Manager::startAllHiding` but because the fade in animation is not finished, the manager won't start hiding that notification, so it will never be gone, unless interacted with again.

> I believe what you try to fix here needs a method that would do (perhaps name it startHidingTimer?)
> 
> ```cpp
> 	if (!_waitingForInput) {
> 		_hideTimer.start(st::notifyWaitLongHide);
> 	}
> ```
> 
> and then call it when isShowing() returns true. With no other changes.

But why should the behavior be different depending on whether the notification is fadingIn or not?

kolayne:
Btw, I would argue that `isShowing()` is a confusing name for this. The way I understood it initially was "is displayed and is not fading out". If you agree (and we still need this method), I would rename `Notification::isShowing()` to `Notification::isFadingIn()`

ilya-fedin:
> What is the purpose of the inner condition?

I guess its purpose is exactly preventing hiding while it's still showing. Apparently to give the user more time to decide whether he wants to click on the notification. The current behavior of not hiding at all may just be intended then...

> The way I understood it initially was "is displayed and is not fading out".

I understand it as "in the process of showing". For your explanation, I would expect the naming to be `isShown`. And, well, actually there's `shown()` in the RpWidget class it derives from.

kolayne:
@ilya-fedin, I have fixed the discussed issues (and updated both desktop-app/lib_ui#254 and this PR). Please, see the new changes

Telegram/SourceFiles/window/notifications_manager_default.h
```diff
@@ -143,7 +143,7 @@ class Widget : public Ui::RpWidget {
 		int shift,
 		Direction shiftDirection);
 
-	bool isShowing() const {
+	bool isFadingIn() const {
 		return _a_opacity.animating() && !_hiding;
 	}
```
kolayne, line 148: As far as I can see, this method is not used anywhere anymore. Should we remove it?

ilya-fedin:
Do you really want to have animation APIs change in the scope of this PR? I haven't noticed any problems without animation changes, it continues to disappear just fine from the opacity it had.

Your changes might affect other animation API usages and encounter regressions in places where the old behavior assumed...

kolayne:
> I haven't noticed any problems without animation changes, it continues to disappear just fine from the opacity it had.

There is one: if while a notification is disappearing (e.g., halfway through the fade out), it is again crossed with the cursor, it will be fading in while the cursor is on it, but if the cursor leaves too quickly (before 150ms), the notification is not fully restored yet at that point.

Without the lib_ui change, in that case, the notification will start fading out from the partially restored state; with the change, the moment the cursor leaves, the notification is reset to opacity 1, regardless of how long the cursor spent in the notification area.

> Your changes might affect other animation API usages and encounter regressions in places where the old behavior assumed...

I guess, the only ways to ensure that is by either checking other uses, or testing. For the last few days I am running a patched version of Telegram Desktop (with my fixes applied), and it seems fine.

I want to additionally note that the behavior is only changed in the case when there was an old animation, which has not finished yet, and another one is started with `.start()`.

ilya-fedin:
> There is one: if while a notification is disappearing (i.e., halfway through the fade out), it is again crossed with the cursor, it will be fading in while the cursor is on it, but if the cursor leaves too quickly (before 150ms), the notification is not fully restored yet at that point.

And?.. I built with such change:
```
diff --git a/Telegram/SourceFiles/window/notifications_manager_default.cpp b/Telegram/SourceFiles/window/notifications_manager_default.cpp
index 198418b585..4b078d5d11 100644
--- a/Telegram/SourceFiles/window/notifications_manager_default.cpp
+++ b/Telegram/SourceFiles/window/notifications_manager_default.cpp
@@ -190,7 +190,7 @@ void Manager::startAllHiding() {
 	if (!hasReplyingNotification()) {
 		int notHidingCount = 0;
 		for (const auto &notification : _notifications) {
-			if (notification->isShowing()) {
+			if (false && notification->isShowing()) {
 				++notHidingCount;
 			} else {
 				notification->startHiding();
@@ -500,6 +500,7 @@ void Widget::opacityAnimationCallback() {
 	updateOpacity();
 	update();
 	if (!_a_opacity.animating() && _hiding) {
+		manager()->startAllHiding();
 		manager()->removeWidget(this);
 	}
 }
```

Here's what I see:

https://github.com/user-attachments/assets/5ef5a945-711d-4302-a9f9-56bf6e248ccc

I don't see any issue with the animation...

kolayne:
> Here's what I see:
> 
>  _20250103_212752.mp4 
> I don't see any issue with the animation...

Same for me. I don't think this behavior is correct. In my understanding, if my cursor has crossed a notification, the notification should reset completely and start fading out from the initial state (i.e., opacity 1). That's what former experience with Telegram Desktop felt like: if a notification is fading out already and I want to keep it on the screen but my cursor happens to be too far, I can just rush and cross the notification (without making the it precise enough to stay in the notification area) and that will keep it.

ilya-fedin:
> if my cursor has crossed a notification, the notification should reset completely and start fading out from the initial state (i.e., opacity 1).

Hm. How your change makes it happen? There's just no time for it to fade in completely. And, well, that's why I proposed to make a new method that would start the _hideTimer: it starts hiding later, after it's completely shown.

kolayne:
> Hm. How your change makes it happen?

Do you mean how I make notifications fade in completely? I fix the behavior of the `Ui::Animations::Simple::start()` method to make sure that the `from` opacity is set before the animation starts. Previously, that was not the case (i.e., `.start()` would behave similar to `.change()` if there is an active animation) but this bug was not visible because Telegram Desktop would not even attempt to hide notifications that were crossed with a cursor too quickly.

> There's just no time for it to fade in completely. And, well, that's why I proposed to make a new method that would start the _hideTimer: it starts hiding later, after it's completely shown.

I agree that this would also lead to the desired behavior. But I wonder if this is the right solution. I guess, the key question here is what kind of behavior of lib_ui is right.

In my understanding, if there is an ongoing animation, and the `.start()` method is called on it instead of `.change()` (so, the opacity values `from` and `to` are provided as arguments), the caller likely expects that the animation to go from `from` to `to`, rather than from the current opacity value to `to`. If that's the case, then there's a bug in lib_ui, and the timer solution just works around that bug. If that's not the case and the previous behavior is desired (i.e., the `from` parameter of `.start()` shall be ignored when there is a previous animation), then my PR to lib_ui is irrelevant, and the timer solution is the right one.

ilya-fedin:
> I fix the behavior of the `Ui::Animations::Simple::start()` method to make sure that the `from` opacity is set before the animation starts.

I.e. you make it jump to opaque rather animate?

kolayne:
It switches instantly to the value of `from` and then animates to `to`. But there is no visual disturbance when I try this with notifications. I can record a demo when I'm back home.

ilya-fedin:
No need to, I already tried your branch

ilya-fedin:
You have to wait @john-preston's review now

john-preston:
image

I don't understand that change. It looks like it doesn't change anything. Please explain.

kolayne:
> image
> 
> I don't understand that change. It looks like it doesn't change anything. Please explain.

Before the change, `Simple::prepare` ignores `from` if there is an ongoing animation (i.e., `_data != nullptr`); after the change, it will set `_data->value = from` in this case.

`Simple::change` continues to work as before; `Simple::start` will now always set the `from` value at the beginning of the animation, even if it replaces the previous unfinished animation.

john-preston:
@kolayne I don't like this really, currently start() doesn't change the current value, so there won't be sudden changes if you call it.

If you really want to enforce a value and don't mind glitches that sudden value change may cause you can always call .stop before .start and get the desired behavior.

I believe there are some places in the codebase which use .stop(); .start(...);, while a huge amount of other places don't and it's a tricky (and unnecessary) thing to check all of them and think if it needs to be replaced with .change or not.

Telegram/SourceFiles/window/notifications_manager_default.cpp
```diff
@@ -549,6 +543,9 @@ void Widget::hideStop() {
 
 void Widget::hideAnimated(float64 duration, const anim::transition &func) {
 	_hiding = true;
+	// Stop the previous animation so as to make sure that the notification is fully restored
+	// before hiding it again. Related to #28811.
```
ilya-fedin, line 547:
```suggestion
	// Stop the previous animation so as to make sure that the notification
	// is fully restored before hiding it again.
	// Related to https://github.com/telegramdesktop/tdesktop/issues/28811.
```

kolayne: Fixed

ilya-fedin:
btw, the second commit's message seem to be wrong now, it doesn't change `.start()`/`.change()` now

kolayne:
Fixed the commit message, thx

john-preston:
Thanks!

PR #28816, opened 2025-01-01
Fix handling of notification disappearing under cursor

### LINKED ISSUE #28813: Closing the topmost notification prevents other notifications from hiding
### Steps to reproduce

1. Receive multiple notifications, so that more than one is displayed in the corner
2. Hover the topmost notification (for the bottom corner) or the bottommost notification (for the top corner)
3. Either close the notification via right-click or the X button, or open the message with the hovered notification from another device.

### Expected behaviour

As the notification is gone from under the cursor, other notifications start fading out

### Actual behaviour

Nothing happens until a cursor passes through notifications again. This is likely because `Notification::leaveEventHook` is not triggered in any of the cases from point 3.

### Operating system

Arch Linux, i3 version 4.24

5.9.2

### Installation source

Other (unofficial) source

### Crash ID

_No response_

### Logs

_No response_

kolayne:
Previously, notifications disappearing under cursor (e.g., because closed manually or open from another device) did not notify the manager properly, as the leaveEventHook was not triggered. This could lead to notifications staying around when not supposed to (see #28813).

This commit fixes that by explicitly notifying manager when the notification widget disappears under the cursor.

Fixes #28813.

### DIFF
--- a/Telegram/SourceFiles/window/notifications_manager_default.cpp
+++ b/Telegram/SourceFiles/window/notifications_manager_default.cpp
@@ -500,7 +500,13 @@ void Widget::opacityAnimationCallback() {
 	updateOpacity();
 	update();
 	if (!_a_opacity.animating() && _hiding) {
-		manager()->removeWidget(this);
+		if (underMouse()) {
+			// The notification is leaving from under the cursor, but in such case leave hook is not
+			// triggered automatically. But we still want the manager to start hiding notifications
+			// (see #28813).
+			manager()->startAllHiding();
+		}
+		manager()->removeWidget(this);  // Deletes `this`
 	}
 }
 


Telegram/SourceFiles/window/notifications_manager_default.cpp
```diff
@@ -553,7 +553,14 @@ void Widget::hideAnimated(float64 duration, const anim::transition &func) {
 }
 
 void Widget::updateOpacity() {
-	setWindowOpacity(_a_opacity.value(_hiding ? 0. : 1.) * _manager->demoMasterOpacity());
+	const auto opacity = _a_opacity.value(_hiding ? 0. : 1.) * _manager->demoMasterOpacity();
+	if (opacity == 0. && underMouse()) {
+		// The notification is leaving from under the cursor, but in such case leave hook is not
```
ilya-fedin, line 558: This doesn't sound correct. The Qt documentation:

> A leave event is sent to the widget when the mouse cursor leaves the widget.

https://doc.qt.io/qt-5/qwidget.html#leaveEvent

Telegram/SourceFiles/window/notifications_manager_default.cpp
```diff
@@ -553,7 +553,14 @@ void Widget::hideAnimated(float64 duration, const anim::transition &func) {
...
+		// triggered automatically. But we still want the manager to start hiding notifications
+		// (see #28813).
+		manager()->startAllHiding();
```
ilya-fedin, line 561: As far as I can see from the code, this function gets called from hide animation. As the animation is already playing, it means the hiding has already started and but you call it again. This doesn't look correct.

ilya-fedin:
I believe the right fix here is finding why leave event is not being emitted for you and fixing that.

Telegram/SourceFiles/window/notifications_manager_default.cpp
kolayne: I suppose, in Qt's terminology the widget becoming invisible is not "mouse cursor leaves the widget"?..

kolayne:
> I believe the right fix here is finding why leave event is not being emitted for you and fixing that.

Makes sense

Telegram/SourceFiles/window/notifications_manager_default.cpp
ilya-fedin: Not sure. I guess it might be platform-dependent if it's not mentioned in documentation.

kolayne: In that case, our code should not rely on that, should it?

ilya-fedin: If I got your idea right, just adding `manager()->startAllHiding();` before `manager()->removeWidget(this);` in `opacityAnimationCallback` should be enough

kolayne: Huh. That makes sense, let me try that.

But I don't think `->startAllHiding()` should be before `->removeWidget(this)`: that would restart the animation on the already hidden notification. We should probably first escape from the manager's control, and then request that it starts hiding notifications, shouldn't we?

ilya-fedin: > that would restart the animation on the already hidden notification

Your code actually does the same right now. Just adding the same condition to other method.

kolayne: Seems so! And, like you said in another comment:

> this function As far as I can see from the code, this function gets called from hide animation. As the animation is already playing, it means the hiding has already started but you call it again. This doesn't look correct.

Anyway, I'll try it now and see if it works as expected.

ilya-fedin: I didn't understand that it's hiding due to being closed back then. It's getting destroyed so it's logical to start hiding again in case mouse over it was preventing hiding.

kolayne: I feel like it's even more logical to start hiding the rest of the notifications when this one is already destroyed (from the manager's perspective)

ilya-fedin: yeah, personally I'm fine with either ordering

kolayne: Omg, does `removeWidget` free the notification from memory? It is then very surprising that it is called from inside a notification. Is it even defined behavior if an object is freed/deleted from within it's member method, even if it's the last statement?

ilya-fedin: Yeah, look at `Manager::removeWidget`, all it does is finds the notification in the vector, erases it and calls `showNextFromQueue()`

kolayne:
@ilya-fedin, I fixed the previously discussed issues. Additionally, I added a comment at the call to `Manager::removeWidget` to note that it deletes the object.

Note, that I only call `manager()->startAllHiding()` if the disappeared notification is `underMouse()` (if the notification disappeared while not under cursor, we don't want any special actions about other notifications' animations, it workds fine already). The fact that the widget reports that it is `underMouse()` makes me think that it is the correct behavior of Qt that `leaveEventHook` is not triggered: from Qt's perspective, the cursor has not left, and this is consistent across different Qt interfaces (i.e., event-driven and direct).

PR #28904, opened 2025-01-28
Fix memory and etc.

RuslanSem:
Fix memory leak and etc.


Aokromes:
it's recommended to do one PR for each fix, the  chance to be merged are higher.

PR #28933, opened 2025-02-04
Added forward as copy(Remove senders name)

pachax001:
Don't Know how this works. Right click context menu also got the save as copy function. Don't know how it happened. Need to check this. If can please add this option to original. It is a much needed one.


ilya-fedin:
You can already do this?

pachax001:
> You can already do this?

ilya-fedin:
If this also adds possibility to choose multiple recipients, that's something you should have been  specify in the commit message I believe

MariuxReloaded:
I second that: the "multi-forward", which allows to send a message to multiple chats all together at once, and the "forward as copy" are features present in 2-3 Telegram Desktop forks (Kotatogram, 64Gram and FAGram) that could be added as very useful functions in the official Telegram Desktop

ilya-fedin:
Conflicts

PR #28943, opened 2025-02-08
Include media sizes to exports

anton-ryzhov:
I hope this suggestion could be classified as "fixing the existing user experience" and therefore considered.

Currently it is possible to include various media types and set a size limit. But it's impossible to predict how much size and time the export will take, and how much data was below and above the limit.

This change saves media sizes even for skipped items, so allows to get this insights.

I've tested my change:
```json
  {
   "id": 111,
   "type": "message",
   "date": "2025-01-30T14:53:04",
   "date_unixtime": "1738245184",
   "from": "",
   "from_id": "",
   "photo": "(File not included. Change data exporting settings to download.)",
   "photo_size": 178404,
   "width": 960,
   "height": 1280,
   "text": "",
   "text_entities": []
  }
```

### DIFF
--- a/Telegram/SourceFiles/export/output/export_output_json.cpp
+++ b/Telegram/SourceFiles/export/output/export_output_json.cpp
@@ -385,6 +385,7 @@ QByteArray SerializeMessage(
 	};
 	const auto pushPhoto = [&](const Image &image) {
 		pushPath(image.file, "photo");
+		push("photo_file_size", image.file.size);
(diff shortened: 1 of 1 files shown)


john-preston:
Maybe photo_file_size? Because photo_size reads like dimensions..

anton-ryzhov:
I also noticed this confusion, but tried to keep it consistent. How about now?

john-preston:
Thanks!

Issue #28981, opened 2025-02-19
Invisible links in a video caption can be clicked through the video player in full screen mode


comdev1337:
### Steps to reproduce

Open a video with links in its caption in full screen mode
![Image](https://github.com/user-attachments/assets/17531313-8fa8-44dc-8103-1213094ab70f)

### Expected behaviour

A user shouldn't be able to click a link which is not visible in full screen mode

### Actual behaviour

Even though the video caption gets hidden, links can still be clicked
![Image](https://github.com/user-attachments/assets/819daf22-f7d9-4322-bb9d-ff9c372babd0)

### Operating system

Windows 11

### Version of Telegram Desktop

5.11.1 x64

### Installation source

Static binary from official website

### Logs

```shell

```

john-preston:
@comdev1337 thanks!

PR #29000, opened 2025-02-24
Video Player: Fix caption elements being clickable in video fullscreen

davidholio:
Fixes https://github.com/telegramdesktop/tdesktop/issues/28981


john-preston:
My "thanks" in the issue meant that I've fixed it in (different part of) code already, in not-yet-pushed-here branch :) But thank you.

Issue #29008, opened 2025-02-27
Touchscreen support on linux is broken after 5.10.7


vipaoL:
### Steps to reproduce

- Install 5.10.8+ (I compiled "5.10.8 beta" from sources and it is not ok, 5.10.7 was ok)
- Open a chat
- Tap on a photo/video/video message/spoiler/answer or long tap on the message

### Expected behaviour

The photo should open after a tap, a context menu should show up after a long tap

### Actual behaviour

The click/long tap does nothing

### Operating system

PostmarketOS edge, Gnome (Desktop) 47.4, Wayland

### Version of Telegram Desktop

5.11.1

### Installation source

Other (unofficial) source

### Logs

```shell

```

Aokromes:
plz try downloading official 5.10.8 from github releases.

ilya-fedin:
> I even was not able to press the "Start Messaging" button without a keyboard.

As you could notice, the appdata mentions that you a keyboard is required to use this software. I assume you're using an unsupported device (perhaps a phone?) so the broken-ness is absolutely expected. Sorry but you have to use a different software for your device (which doesn't necessarily exist, in which case someone interested from the community has to make one)

vipaoL:
Anyway, it is a regression. I found the commit that breaks it:
```
7ac849ab12c766149206a84fd2bd150f4dac8538 is the first bad commit
commit 7ac849ab12c766149206a84fd2bd150f4dac8538
Author: John Preston <johnprestonmail@gmail.com>
Date:   Tue Feb 4 20:16:25 2025 +0400

    Don't update mouse in non-active window.

 Telegram/SourceFiles/history/history_inner_widget.cpp | 6 +++++-
 Telegram/SourceFiles/history/history_inner_widget.h   | 1 +
 2 files changed, 6 insertions(+), 1 deletion(-)
```

ilya-fedin:
I'm doubt anyone will look into it unless it reproduces on Windows with a laptop

ilya-fedin:
The web version might be a better fit for a phone. It's developed to be used on phones in mind AFAIK, they even ditched the idea of using react and tdlib as they were too slow on phones.

Jaymz665:
@ilya-fedin 

It also doesn't work on my touchscreen laptop (void linux)
And on my smartphone (postmarketos)
Maybe it seems to me, but you closed this problem early, especially since have already found where the error is growing from.

Jaymz665:
[...]

PR #29016, opened 2025-02-28
Revert "Don't update mouse in non-active window."

hitechshell:
This reverts commit 7ac849ab12c766149206a84fd2bd150f4dac8538.

This commit introduce bug on mobile linux with Wayland

1) the photo did not open after clicking
2) context menu did not appear after a long press
3) jumping to the original message after clicking on a quote didn't work

fix for https://github.com/telegramdesktop/tdesktop/issues/29008

### DIFF
--- a/Telegram/SourceFiles/history/history_inner_widget.cpp
+++ b/Telegram/SourceFiles/history/history_inner_widget.cpp
@@ -1577,7 +1577,6 @@ void HistoryInner::mouseMoveEvent(QMouseEvent *e) {
 		mouseReleaseEvent(e);
 	}
 	if (reallyMoved) {
-		_mouseActive = true;
 		lastGlobalPosition = e->globalPos();
 		if (!buttonsPressed || (_scrollDateLink && ClickHandler::getPressed() == _scrollDateLink)) {
 			keepScrollDateForNow();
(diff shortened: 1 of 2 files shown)


NekoCWD:
Yey! I'll can update tg on my phone 🐈‍⬛ ❤️

ilya-fedin:
I'm doubt you can just revert it

john-preston:
@hitechshell Because this change was introduced to fix some problems in different scenarios and just reverting the change returns the problems with autoplaying videos receiving mouse events when they are in an inactive window below media viewer where the video was opened. There is a small chance the original problems were fixed more robustly already and that specific revert won't bring them back, but the chance is small and it needs to be tested first and if the problems return we'll need to find some other workarounds.

john-preston:
It's a problem that the system in your case sends mouse events without Enter event / Leave event for the widgets, this isn't supposed to work really.

ilya-fedin:
> It's a problem that the system in your case sends mouse events without Enter event / Leave event for the widgets, this isn't supposed to work really.

Aren't mouse events synthesized from ScrollArea with touch? Are enter/leave events supposed to arrive in this case?

john-preston:
I've checked the revert and unfortunately it returns bad glitches in auto-playing videos that I was fixing with this commit.

I'll add a different workaround for the next version, the input will still be processed if the window is active, even if no Enter event was caught.

ilya-fedin:
> Are enter/leave events supposed to arrive in this case?

FWIW, I checked the code and looks like the answer is yes, they are. So it's absolutely unknown why it doesn't work. And, moreover, could indicate that Qt's compatibility with the system in question is bad and touch support generally being broken.

PR #29021, opened 2025-03-01
Respect "Do not preserve zone information in file attachments" policy on Windows

phys-winner:
**Problem:**
Telegram Desktop on Windows does not respect the system policy "Do not preserve zone information in file attachments". As a result, even when this policy is enabled, an NTFS alternate data stream (`Zone.Identifier:$DATA`) is still created for every downloaded file.

**Solution:**
This change ensures that the Zone.Identifier stream is not created when the system policy explicitly restricts this behavior, aligning Telegram Desktop's behavior with Windows security settings.

### DIFF
--- a/Telegram/SourceFiles/platform/win/file_utilities_win.cpp
+++ b/Telegram/SourceFiles/platform/win/file_utilities_win.cpp
@@ -117,6 +117,28 @@ HBITMAP IconToBitmap(LPWSTR icon, int iconindex) {
 	return (HBITMAP)CopyImage(result, IMAGE_BITMAP, 0, 0, LR_DEFAULTSIZE | LR_CREATEDIBSECTION);
 }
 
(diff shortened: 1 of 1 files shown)


Telegram/SourceFiles/platform/win/file_utilities_win.cpp
```diff
@@ -276,7 +276,36 @@ void UnsafeLaunch(const QString &filepath) {
 	ShellExecute(0, L"open", wstringPath.c_str(), 0, 0, SW_SHOWNORMAL);
 }
 
+bool ShouldSaveZoneInformation() {
```
ilya-fedin, line 279: Anything not exported should go to private namespace

phys-winner: Fixed.

PR #29097, opened 2025-03-17
update ada-url to v3.2.1

anonrig:
I'm one of the authors of ada-url. I realized that telegram is using an old version of Ada, and wanted to contribute the following change to Telegram. The only major change I added was adding "ADA_INCLUDE_URL_PATTERN" flag which removes the URLPattern implementation, which Telegram doesn't depend on, since it's available after v3.

### DIFF
--- a/Telegram/build/prepare/prepare.py
+++ b/Telegram/build/prepare/prepare.py
@@ -1864,7 +1864,7 @@ release:
 """)
 
 stage('ada', """
-    git clone -b 3.2.1 https://github.com/ada-url/ada.git
+    git clone -b v3.2.1 https://github.com/ada-url/ada.git
     cd ada
 win:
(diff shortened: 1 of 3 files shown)


Telegram/build/prepare/prepare.py
```diff
@@ -1864,13 +1864,14 @@ def runStages():
 """)
 
 stage('ada', """
-    git clone -b v2.9.0 https://github.com/ada-url/ada.git
+    git clone -b 3.2.1 https://github.com/ada-url/ada.git
```
anonrig, line 1867:
```suggestion
    git clone -b v3.2.1 https://github.com/ada-url/ada.git
```

john-preston:
Thanks!

Issue #28926, opened 2025-02-03
Washed out colors in videos


magicxor:
### Steps to reproduce

1. Actively use Telegram Desktop on Windows for a long period of time (~1 day or more)
2. Open any video

### Expected behaviour

Video is displayed correctly. No washed out colors, no magenta overlay.

https://github.com/user-attachments/assets/53c7aaea-f9fa-4df2-805a-d3e57142db07

### Actual behaviour

Colors are washed out. The entire image has a strong pink or magenta overlay.

https://github.com/user-attachments/assets/6cf03521-ca54-4182-9a34-0989489a2712

### Operating system

Windows 11 Pro 24H2, Version 10.0.26100.2894, Windows Feature Experience Pack 1000.26100.36.0

### Version of Telegram Desktop

5.10.7 x64

### Installation source

Static binary from official website

### Logs

```shell

```

ilya-fedin:
Are you sure you haven't enabled hardware video decoding in advanced settings? It's highly bugged.

magicxor:
Thanks for the suggestion! I'll turn it off and see if the problem persists.

magicxor:
Can't reproduce the bug with hardware video decoding = off. Thanks.

PR #29105, opened 2025-03-20
Enable video hardware acceleration by default

OriginCode:
Hardware acceleration is widely available.

### DIFF
--- a/Telegram/SourceFiles/core/core_settings.h
+++ b/Telegram/SourceFiles/core/core_settings.h
@@ -1048,11 +1048,7 @@ private:
 	rpl::variable<Media::OrderMode> _playerOrderMode;
 	bool _macWarnBeforeQuit = true;
 	std::vector<uint64> _accountsOrder;
-#ifdef Q_OS_MAC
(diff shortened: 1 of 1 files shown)


Aokromes:
i don't think this will be afected because some people have bugs with it enabled.

ilya-fedin:
Hardware video acceleration is very unstable outside of macOS. There are periodic reports about green tint on Windows and about crashes on both Windows and Linux. I also remember other types of issue but don't remember which they were exactly are as they were reported rarely. All these issues have to be solved before the option could be enabled. Do you have patches for them?

ilya-fedin:
The last hardware decoding-related report is https://github.com/telegramdesktop/tdesktop/issues/28926

ilya-fedin:
https://github.com/telegramdesktop/tdesktop/issues?q=hardware%20decoding

OriginCode:
Got it, will leave it as a downstream feature patch for now. Thank you.

farouk2025 (commented):
Thanks so much

PR #29308, opened 2025-05-14
click_handler_types.cpp UrlRequiresConfirmation(): add linkedin.com

vitaly-zdanevich:
I think Linkedin is safe enough.


Aokromes:
it lookt that list only contains telegram owned domains.

Justfunnyboy1 (approved):

PR #29464, opened 2025-06-17
Change `const T&&` parameters to `T&&` to enable proper move semantics

Sea-n:
Replace `const T&&` parameters with correct `T&&` (non-const rvalue references) so that `move()` semantics actually take effect.

### `const T&&` disables moves
Move constructors/operators are declared as `widget(T&&)` / `widget& operator=(T&&)`. When a parameter is `const T&&`, the argument becomes an immutable rvalue, so the move constructor cannot be called, only the copy constructor is viable.

### `std::move()` is pointless on `const` rvalues
`std::move` merely casts to exactly the same reference type it receives; if the variable is const, the result is still `const T&&`.

### Unnecessary copies & misleading code
Large objects are copied instead of moved, wasting memory/CPU and hiding the real cost from programmers.

## Fix

Drop the `const` qualifier on all rvalue-reference parameters:

```diff
 ItemImage::ItemImage(
- const QPixmap &&pixmap,
+ QPixmap &&pixmap,
  ItemBase::Data data)
 : ItemBase(std::move(data))
 , _pixmap(std::move(pixmap)) {
 }
```

## Impact

- Restores true move semantics, eliminating extra deep/shallow copies.
- Reduces temporary allocations when loading images or constructing message items.
- Aligns code intent with behavior, improving maintainability and readability.

## References

- I found this improvement by reading "Effective Modern C++" book, please refer to page 160.
- StackOverflow: [Why can we use `std::move` on a `const` object?](https://stackoverflow.com/a/28595207/5201431)
- PVS Code Analyzer: [V833. Using 'std::move' function's with const object disables move semantics](https://pvs-studio.com/en/docs/warnings/v833/)


john-preston:
Last one changes const lvalue reference to non const lvalue reference 🤔

john-preston:
And it even broke the build as you can see here:

https://github.com/telegramdesktop/tdesktop/actions/runs/15718851021/job/44295293711?pr=29464

Sea-n:
Oh, thanks for pointing out [`rpl::start_with_next()`](https://github.com/desktop-app/lib_rpl/blob/9a3ce435/rpl/producer.h#L596) takes const lvalue reference, which is different thing with this one. 😅

PR #29708, opened 2025-08-27
fix: Add shmem needed for webkitgtk and gstreamer use

kenvandine:


ilya-fedin:
Closing since this doesn't really fix the issue, what it really wants is /var/tmp

PR #29832, opened 2025-09-29
Add "Always show proxy button" to experimental setting

### LINKED ISSUE #29766: Feature Request: Option to Keep Proxy Button Always Visible

Yes. The proxy button (round button in the bottom-left corner of Telegram Desktop) only appears when a proxy is enabled. Sometimes, when Telegram cannot connect at all, this button does not appear either, which makes it impossible to quickly access proxy settings.

I would like to have an option that keeps the proxy button always visible, regardless of whether a proxy is enabled or not. This would allow users to quickly open and change proxy settings at any time.

Currently, the only way is to go through the main settings menu, which is slower and not always accessible when the app cannot connect. Having the button always visible would be a much faster and more convenient solution.

### Additional context

_No response_

yueneiqi:
Close https://github.com/telegramdesktop/tdesktop/issues/29766

---
**This PR was primarily authored with Codex using GPT-5-Codex and then hand-reviewed by me. I AM responsible for every change made in this PR. I aimed to keep it aligned with our goals, though I may have missed minor issues. Please flag anything that feels off, I'll fix it quickly.**

### DIFF
--- a/Telegram/SourceFiles/settings/settings_experimental.cpp
+++ b/Telegram/SourceFiles/settings/settings_experimental.cpp
@@ -31,6 +31,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 #include "window/window_peer_menu.h"
 #include "window/window_session_controller.h"
 #include "window/window_controller.h"
+#include "window/window_connecting_widget.h"
 #include "window/notifications_manager.h"
 #include "storage/localimageloader.h"
 #include "data/data_document_resolver.h"
@@ -83,7 +84,11 @@ void AddOption(
 				tr::lng_settings_experimental_irrelevant(tr::now));
 			return;
 		}
+		const auto isProxyAlwaysVisible = (&option == &base::options::lookup<bool>(Window::kOptionProxyAlwaysVisible));
 		option.set(toggled);
+		if (isProxyAlwaysVisible) {
+			Window::NotifyProxyAlwaysVisibleChange();
+		}
 		if (restarter) {
 			restarter->callOnce(st::settingsButtonNoIcon.toggle.duration);
 		}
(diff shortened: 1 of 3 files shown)


23rd:
First of all, the project doesn't build and has errors.
Secondly, it doesn't work very well in some cases. During different account switches and when toggling the proxy itself on/off, the icon still disappears despite enabling this new experimental feature.

ilya-fedin:
Reading the linked issue

> Sometimes, when Telegram cannot connect at all, this button does not appear either, which makes it impossible to quickly access proxy settings.

It sounds like a new option isn't needed at all, just the bug when Telegram can't connect has to be fixed as the icon is supposed to be shown when connecting even if a proxy isn't set

yueneiqi:
> First of all, the project doesn't build and has errors. Secondly, it doesn't work very well in some cases. During different account switches and when toggling the proxy itself on/off, the icon still disappears despite enabling this new experimental feature.

Sorry about that. My initial test was just on my local Windows setup, and it seemed fine(actually not). I've now run it on a fresh 64-bit Windows build and WSL.

Win64:
win64


WSL:
wsl

yueneiqi:
> Reading the linked issue
> 
> > Sometimes, when Telegram cannot connect at all, this button does not appear either, which makes it impossible to quickly access proxy settings.
> 
> It sounds like a new option isn't needed at all, just the bug when Telegram can't connect has to be fixed as the icon is supposed to be shown when connecting even if a proxy isn't set

You're right that this option doesn't resolve the issue entirely. However, it does provide users with a helpful shortcut to their proxy settings.

Regarding the bug, I haven't been able to reproduce it on my end. My hope is that this shortcut will give users better context, enabling them to provide a more detailed report if they encounter it again. I will be sure to update this issue (or PR) if I find a way to reproduce or fix it.

Telegram/SourceFiles/window/window_proxy_button_visibility.cpp
```diff
@@ -0,0 +1,40 @@
+/*
```
ilya-fedin: You don't have to create a separate file. As you could notice, all other experimental options are defined in the file they're used from. In this case, it will be window_connecting_widget.{cpp,h}.

ilya-fedin:
Conflicts

PR #29869, opened 2025-10-07
enchant: allow toggling individual dictionaries

WhyNotHugo:
Allow disabling and enabling individual dictionaries with the Enchant/Hunspell backend.

This is mostly relevant in shared systems (where different users want to enable different sets of languages) or on distributions which install an exorbitant amount of dictionaries by default.

Depends-On: https://github.com/desktop-app/lib_spellcheck/pull/32

### DIFF
--- a/Telegram/SourceFiles/boxes/dictionaries_manager.cpp
+++ b/Telegram/SourceFiles/boxes/dictionaries_manager.cpp
@@ -19,6 +19,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 #include "mainwidget.h"
 #include "mtproto/dedicated_file_loader.h"
 #include "spellcheck/spellcheck_utils.h"
+#include "spellcheck/platform/platform_spellcheck.h"
 #include "ui/wrap/vertical_layout.h"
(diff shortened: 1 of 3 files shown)


WhyNotHugo:
Rocky Linux builds failed because my changes don't build when **not** using the Enchant backend. I'll need to review this further.

Build error on Windows is be because I'm using `Platform::Spellchecker::AvailableLanguages` which I only implemented for the Enchant backend. I'll have to rethink some details around this.

Build errors on macOS seem unrelated to my changes (fails installing dependencies). If it weren't for that, it would likely have failed with the same error as the Windows build.

WhyNotHugo:
I'll keep this as a draft while I re-think some of this.

WhyNotHugo:
New version should build okay on other platforms. Will keep as draft until CI finishes.

WhyNotHugo:
Pretty sure that the macOS failure is unrelated: https://github.com/telegramdesktop/tdesktop/actions/runs/18478111936/job/52647143537

WhyNotHugo:
Yeah, `dev` branch is also failing for that target.

PR #29900, opened 2025-10-15
Refactor core settings clean

omnizs:
Refactors the `core_settings` module: improves code quality, readability, and fixes minor issues.

### DIFF
--- a/Telegram/SourceFiles/core/core_settings.cpp
+++ b/Telegram/SourceFiles/core/core_settings.cpp
@@ -1291,8 +1291,7 @@ void Settings::incrementRecentEmoji(RecentEmojiId id) {
 			}
 			std::swap(*i, *(i - 1));
 		}
-	}
-	else {
+	} else {
 		_recentEmoji.push_back({ id, 1 });
 		for (i = _recentEmoji.end() - 1; i != _recentEmoji.begin(); --i) {
 			if ((i - 1)->rating > i->rating) {
--- a/Telegram/SourceFiles/core/core_settings.cpp
+++ b/Telegram/SourceFiles/core/core_settings.cpp
@@ -214,7 +214,7 @@ QByteArray Settings::serialize() const {
 		+ Serialize::bytearraySize(proxy)
 		+ sizeof(qint32) * 2
 		+ Serialize::bytearraySize(_photoEditorBrush)
-		+ sizeof(qint32) * 3
+		+ sizeof(qint32) * 4
 		+ Serialize::stringSize(_customDeviceModel.current())
 		+ sizeof(qint32) * 4
 		+ (_accountsOrder.size() * sizeof(quint64))
@@ -337,6 +337,7 @@ QByteArray Settings::serialize() const {
 			<< qint32(_hiddenGroupCallTooltips.value())
 			<< qint32(_disableOpenGL ? 1 : 0)
 			<< _photoEditorBrush
+			<< qint32(_photoEditorHighContrastMarker.current() ? 1 : 0)
 			<< qint32(_groupCallNoiseSuppression ? 1 : 0)
 			<< qint32(SerializePlaybackSpeed(_voicePlaybackSpeed))
 			<< qint32(_closeBehavior)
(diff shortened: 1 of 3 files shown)


Telegram/SourceFiles/core/core_settings.cpp
```diff
@@ -214,7 +214,7 @@ QByteArray Settings::serialize() const {
 		+ Serialize::bytearraySize(proxy)
 		+ sizeof(qint32) * 2
 		+ Serialize::bytearraySize(_photoEditorBrush)
-		+ sizeof(qint32) * 3
+		+ sizeof(qint32) * 4
```
ilya-fedin, line 217: Why is this change?

Telegram/SourceFiles/core/core_settings.cpp
```diff
@@ -1273,29 +1276,23 @@ void Settings::incrementRecentEmoji(RecentEmojiId id) {
...
+		for (; i != _recentEmoji.begin(); --i) {
+			if ((i - 1)->rating > i->rating) {
+				break;
 			}
-			break;
+			std::swap(*i, *(i - 1));
 		}
 	}
-	if (i == e) {
+	else {
```
ilya-fedin:
```suggestion
	} else {
```

Telegram/SourceFiles/core/core_settings.h
```diff
@@ -579,6 +579,15 @@ class Settings final {
...
 	}
+	[[nodiscard]] bool photoEditorHighContrastMarker() const {
+		return _photoEditorHighContrastMarker.current();
+	}
+	[[nodiscard]] rpl::producer<bool> photoEditorHighContrastMarkerValue() const {
+		return _photoEditorHighContrastMarker.value();
+	}
+	void setPhotoEditorHighContrastMarker(bool value) {
+		_photoEditorHighContrastMarker = value;
+	}
```
ilya-fedin, line 590: Why these methods are added?

Telegram/SourceFiles/core/core_settings.cpp
```diff
@@ -976,9 +984,9 @@ void Settings::addFromSerialized(const QByteArray &serialized) {
 	}
 	_thirdSectionInfoEnabled = thirdSectionInfoEnabled;
 	_dialogsWithChatWidthRatio = dialogsWithChatWidthRatio;
-	_dialogsNoChatWidthRatio = (dialogsWithChatWidthRatio > 0)
-		? dialogsWithChatWidthRatio
-		: dialogsNoChatWidthRatio;
+	_dialogsNoChatWidthRatio = (dialogsNoChatWidthRatio > 0)
+		? dialogsNoChatWidthRatio
+		: dialogsWithChatWidthRatio;
```
ilya-fedin, line 989: Why this change?

Telegram/SourceFiles/core/core_settings.cpp
omnizs: These changes were made to ensure that the serialized size calculation is accurate, reflecting the actual structure of the data being stored. If this is not required or should be handled differently, please advise and I'll update the code.

Telegram/SourceFiles/core/core_settings.cpp
omnizs: Thank you for the suggestion!  
I'll update this part to use the correct `else { ... }` block style as you proposed.

Telegram/SourceFiles/core/core_settings.h
omnizs: These methods were added to provide access to the high-contrast marker state and its reactive value.  
If this functionality is not needed or should be implemented differently, let me know and I'll adjust accordingly.

Telegram/SourceFiles/core/core_settings.cpp
omnizs: This change ensures that `_dialogsNoChatWidthRatio` is always set to a valid value:  
If `dialogsNoChatWidthRatio` is greater than zero, use it  
Otherwise, fallback to `dialogsWithChatWidthRatio`  
If this logic is incorrect, please clarify the expected behavior.

Telegram/SourceFiles/core/core_settings.h
ilya-fedin: I don't see those methods used anywhere in this PR. This looks like irrelevant change.

ilya-fedin (changes requested):
This PR is titled as refactor but actually adds new photoEditorHighContrastMarker variable

PR #29903, opened 2025-10-15
fix: correct QPainter method call from setPen to setBrush

omnizs:
- Fixed critical bug in `color_picker.cpp` line 321
- Changed `p.setPen(Qt::NoBrush)` to `p.setBrush(Qt::NoBrush)`
- `setPen` and `setBrush` are different `QPainter` properties


ilya-fedin (changes requested):
This PR is titled `fix: correct QPainter method call from setPen to setBrush` but has a lot of unrelated changed adding some "high contrast marker"

PR #29916, opened 2025-10-17
Unification of Send menu

InternalProgramError:
Unification of Send menu (https://bugs.telegram.org/c/55220):

After this fix it will be as on Android:
- Schedule Message
- Send when Online
- Send without sound

Before fixing it was 
- Send without sound
- Schedule Message
- Send when Online

### DIFF
--- a/Telegram/SourceFiles/menu/menu_send.cpp
+++ b/Telegram/SourceFiles/menu/menu_send.cpp
@@ -690,11 +690,6 @@ FillMenuResult FillSendMenu(
 		? *iconsOverride
 		: st::defaultComposeIcons;
 
-	if (sending && type != Type::Reminder) {
-		menu->addAction(
-			tr::lng_send_silent_message(tr::now),
-			[=] { action({ Api::SendOptions{ .silent = true } }, details); },
-			&icons.menuMute);
 	}
 	if (sending && type != Type::SilentOnly) {
 		menu->addAction(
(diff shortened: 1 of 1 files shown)


john-preston:
Thanks.

Neurotoxin001:
This is a terrible change. Thousands of people have gotten used to the old button layout over many years, and now when they try to schedule a post on the channel, they end up sending the message without sound instead.

It's all because of the habit of tapping the exact area of the screen where the “schedule message” button has always been located.

john-preston:
@Neurotoxin001 Hmm, this change didn't land in 6.6.2, right? You think this will happen much, they click through without looking?

Neurotoxin001:
That's right, it hasn't been released yet, but I compiled a tdesktop with new commits and my colleagues are using my version, and yes, they started to frequently miss clicking and send posts past the scheduled messages.

john-preston:
@Neurotoxin001 Ok, fair enough. Lets revert.

InternalProgramError:
Guys, this is the same issue for me. I mostly use the app on my phone, so I’ve gotten used to Schedule being the first button. Then, when I try to schedule something from my computer, I end up sending it as a 'silent message' by mistake. Of course, you’re in a better position to judge how many people this actually affects.

Neurotoxin001:
Well, it's up to the developer to decide.

Issue #27225, opened 2023-12-14
compilation error: no matching function for call to 'make_optional'


ZerdoX-x:
### Steps to reproduce

Trying to emerge `net-im/telegram-desktop-4.12.2` with clang + mold (if this matters)

### Expected behaviour

Successful compilation

### Actual behaviour

Snippet from build log:
```
[...]

PR #29940, opened 2025-10-23
Fix compilation with Clang-21 + libstdc++-16

### LINKED ISSUE #29939: Clang + libstdc++-16 fail: no matching member function for call to 'emplace'
### Steps to reproduce

Hi, there is some new code which fails to compile when compiler is set to clang-21 with newest libstdc++-16 (from git):
```
/var/tmp/portage/net-im/telegram-desktop-6.2.3/work/tdesktop-6.2.3-full/Telegram/SourceFiles/inline_bots/bot_attach_web_view.cpp:1406:23: error: no matching member function for call to 'emplace'
 1406 |                 _botFullWaitingArgs.emplace();
      |                 ~~~~~~~~~~~~~~~~~~~~^~~~~~~
/usr/lib/gcc/x86_64-pc-linux-gnu/16/include/g++-v16/optional:1126:2: note: candidate template ignored: requirement 'is_constructible_v<InlineBots::WebViewInstance::ShowArgs>' was not satisfied [with _Args = <>]
 1126 |         emplace(_Args&&... __args)
      |         ^
/usr/lib/gcc/x86_64-pc-linux-gnu/16/include/g++-v16/optional:1138:2: note: candidate function template not viable: requires at least argument '__il', but no arguments were provided
 1138 |         emplace(initializer_list<_Up> __il, _Args&&... __args)
      |         ^       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Apparently, as `InlineBots::WebViewInstance::ShowArgs` is private nested struct, it is not considered as constructible, see https://stackoverflow.com/questions/60490574/stdis-constructible-returns-inconsistent-value-for-private-constructor

One possible solution is to use `_botFullWaitingArgs = ShowArgs{};`, see attached PR.

Also please check and merge related PR https://github.com/TelegramMessenger/tgcalls/pull/50, if you want to compile it.

### Expected behaviour

It compiles.

### Actual behaviour

It fails to compile.

### Operating system

Gentoo

6.2.3

### Installation source

Other (unofficial) source

### Crash ID

_No response_

### Logs

```shell

```

AngryLoki:
.emplace() can only be used on constructible objects, while private nested struct is not considered as constructible due to access control.

Closes #29939

### DIFF
--- a/Telegram/SourceFiles/inline_bots/bot_attach_web_view.cpp
+++ b/Telegram/SourceFiles/inline_bots/bot_attach_web_view.cpp
@@ -1403,7 +1403,7 @@ void WebViewInstance::showGame() {
 	Expects(v::is<WebViewSourceGame>(_source));
 
 	if (!_bot->isFullLoaded()) {
-		_botFullWaitingArgs.emplace();
+		_botFullWaitingArgs = ShowArgs{};
 		return;
 	}
 	const auto game = v::get<WebViewSourceGame>(_source);


ilya-fedin:
> due to access control

What do you mean by this? Why is it ok with GCC, MSVC and AppleClang?

ilya-fedin:
I believe you're just hitting this bug in Clang/libstdc++ interoperation, it was happening previously (#27225, https://github.com/desktop-app/tg_owt/issues/83) but the fix was different (https://github.com/desktop-app/tg_owt/pull/104):
https://github.com/llvm/llvm-project/issues/50248
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=101227

AngryLoki:
> I believe you're just hitting this bug in Clang/libstdc++ interoperation

Yeah, I think you are right. I tried to build a minimal example here - https://godbolt.org/z/37YhPEjKs - but I can't demonstrate my exact setup:
* clang 21.1.4
* gcc/libstdc++ 16.0.0 20251022

> the fix was different

yes, adding a default constructor helps with `is_constructible_v` as well, but the struct stops from being an aggregate and can not be initialized with designated initializers anymore (which is done in a few places).

So maybe this is indeed a compiler bug, but I don't see signs that it will be fixed soon.

PR #30034, opened 2025-11-21
Remove smooth scrolling in profiles when "Interface Animations" setting is disabled

i-walk-away:
disabling `"Interface Animations"` option in battery settings does not disable the smooth scrolling in profiles. experimental option `"Use legacy scroll processing in profiles"` does not disable it either. smooth scrolling is not present in any other scrollable interface element except for just profiles, which makes it inconsistent across the application

## Expected behavior
disabling "Interface animations" also should disable smooth scrolling in profiles

## Factual behavior
smooth scrolling in profiles is unaffected by any setting

## Implementation
disable smooth scrolling animation in profile pages when `Interface Animations` setting is turned off in battery saving settings. the fix checks `anim::Disabled()` before starting scroll animations in `setupScrollHandling()` and `setupScrollHandlingWithFilter()` methods

i have reviewed how the check for disabled `Interface Animations` is implemented in other parts of the source code to make sure that my changes are consistent with the already existing patterns, and it was, in fact, a simple "if else" check for `anim::Disabled()`, so i did just that

### DIFF
--- a/Telegram/SourceFiles/info/info_flexible_scroll.cpp
+++ b/Telegram/SourceFiles/info/info_flexible_scroll.cpp
@@ -166,7 +166,11 @@ void FlexibleScrollHelper::setupScrollHandling() {
 				: previousValue;
 			if (!_scrollAnimation.animating()) {
 				_scrollTopTo = ((nextStep != -1) ? nextStep : top);
-				_scrollAnimation.start();
+				if (!anim::Disabled()) {
+					_scrollAnimation.start();
+				} else {
+					scrollToY(_scrollTopTo);
+				}
 			} else {
 				if (_scrollTopTo > _scrollTopFrom) {
 					// Down.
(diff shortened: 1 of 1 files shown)


ilya-fedin:
> experimental option `"Use legacy scroll processing in profiles"` does not disable it either

Yeah, looks like it just switches between different smooth scroll impls

23rd:
I would say, if we want to disable the smooth scrolling, we should check `anim::Disabled` before this snippet.
```cpp
	if (AlternativeScrollProcessing.value()) {
		setupScrollHandling();
	} else {
		setupScrollHandlingWithFilter();
	}
```

i-walk-away:
> I would say, if we want to disable the smooth scrolling, we should check `anim::Disabled` before this snippet.
> 
> ```c++
> 	if (AlternativeScrollProcessing.value()) {
> 		setupScrollHandling();
> 	} else {
> 		setupScrollHandlingWithFilter();
> 	}
> ```

sounds valid

ilya-fedin:
Conflicts

PR #30150, opened 2025-12-27
Update to modern OpenGL library on Linux

ilya-fedin:


gudvinr:
Doesn't it seem like there's now a bit too much libraries linked?

`libGL` supposedly has similar content to `libOpenGL` + `libGLX`. Since you removing "legacy" GL support, `libGL` should be gone now.

And, given that `libEGL` already there and `libepoxy` supports more modern EGL, shouldn't you also remove GLX support altogether?

ilya-fedin:
@gudvinr yeah, you're right, libGL shouldn't be linked anymore with this PR. Regarding GLX, I believe it has to stay.

PR #30185, opened 2026-01-07
Fix invalid minimum birthday year in edit birthday box

roman-olefir:
The client allows selecting the year 1875 in the edit birthday box, which
results in a BIRTHDAY_INVALID error due to the 150-year age limit enforced by
the server.

This change updates kYearMin to 1876 so the client-side validation matches
the server restriction


Aokromes:
imho the min date must be current minus 150, not some static date.

john-preston:
yeah, this will break in a year (?)

PR #30289, opened 2026-02-15
Fix unread badge counting from inactive accounts when notifyFromAll is disabled. Fix #26938

### LINKED ISSUE #26938: Receiving notifications from another account despite the setting being disabled
### Steps to reproduce

1. Use two accounts in Telegram Desktop
2. Turn off this setting not to receive notifications from other account:
image
4. Use one of these accounts
5. Write a message on the second account
6. You will see red badge counter from the second account, even though the option is disabled.
image

### Expected behaviour

I don't want to see any notifications from other account, which is not used right now. 
Maybe there was a logic that I will not get these bubble-notifications from other account, but who even use these bubbles in 2023?
I expect not to get information about notifications from other account even from this tray icon.

### Actual behaviour

I see badge counter for messages received on another account.

### Operating system

Ubuntu 22.04.3 LTS x86_64

4.10.3

### Installation source

Snap

### Crash ID

_No response_

### Logs

_No response_

trollkotze:
…ce, no prefix convention visible in their history) and includes the issue-closing keyword as required.Fixes #26938
Domain::updateUnreadBadge() unconditionally summed unread counts from all accounts. When notifyFromAll is disabled, it now skips inactive accounts, so the badge only reflects the active account's unreads.
Also triggers a badge recalculation when switching accounts (since the set of counted accounts changes), and when toggling the setting itself.
Changes
- main_domain.cpp: In updateUnreadBadge(), skip non-active accounts when notifyFromAll is off.
- main_domain.cpp: In activate(), schedule a badge update on account switch when notifyFromAll is off.
- settings_notifications.cpp: Call notifyUnreadBadgeChanged() when the setting is toggled so the badge updates immediately.
Note on #26938
The issue was closed as intended behavior (https://github.com/telegramdesktop/tdesktop/issues/26938#issuecomment-1768017498), with the reasoning that the "Notify from all accounts" setting was only meant to control desktop notifications, not badge counters.
I've read that and I'm ignoring it, because I think that tthis intention was stupid. A setting called "Notify from all accounts" that still counts unreads from all accounts in the badge is unintuitive — the badge is a notification. If this was intentional, the intention lacked common sense. Over two years have passed since that decision, and intentions can change. I'm opening this PR in the hope the maintainers will reconsider.

### DIFF
--- a/Telegram/SourceFiles/main/main_domain.cpp
+++ b/Telegram/SourceFiles/main/main_domain.cpp
@@ -244,7 +244,11 @@ void Domain::notifyUnreadBadgeChanged() {
 void Domain::updateUnreadBadge() {
 	_unreadBadge = 0;
 	_unreadBadgeMuted = true;
+	const auto notifyFromAll = Core::App().settings().notifyFromAll();
 	for (const auto &[index, account] : _accounts) {
+		if (!notifyFromAll && account.get() != _active.current()) {
+			continue;
+		}
 		if (const auto session = account->maybeSession()) {
 			const auto data = &session->data();
 			_unreadBadge += data->unreadBadge();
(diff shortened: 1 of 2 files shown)


ilya-fedin:
> Fixes #26938

You could notice it's told there that this is intended behavior...

PR #30293, opened 2026-02-16
Calls: fix screenshare audio on mic mute and add incoming volume control

paterkleomenis:
## Summary

- **Fix screenshare audio stopping on mic mute**: When the user mutes their microphone during a 1-to-1 call with screen audio sharing, the system audio no longer stops. Previously `setMuteMicrophone()` muted the entire outgoing audio channel, silencing both mic and system audio. Now when screen audio is active, the mic is muted at the ADM mixing level while the channel stays unmuted so system audio continues flowing.

- **Add right-click volume control for incoming audio**: Right-clicking on the incoming video in a 1-to-1 call shows a popup menu with a volume slider and mute toggle, allowing local control of the remote user's audio volume. Reuses the existing `MenuVolumeItem` from group calls.

- **Always use MixingAudioDeviceModule wrapper**: The ADM is now always wrapped with `MixingAudioDeviceModule` so playback volume control is available regardless of loopback capture support.

Depends on https://github.com/desktop-app/lib_webrtc/pull/24

## Testing

```
cmake --build out --config Debug --target Telegram
```

- Start a 1-to-1 call with screen sharing + audio enabled
- Mute microphone → system audio should continue playing for the remote
- Right-click on the incoming video → volume slider should appear
- Adjust slider / mute → incoming audio volume changes locally

### DIFF
--- a/Telegram/SourceFiles/calls/calls.style
+++ b/Telegram/SourceFiles/calls/calls.style
@@ -1008,6 +1008,7 @@ groupCallScreenShareSmall: CallButton(groupCallSettingsSmall) {
 		rippleAreaPosition: point(8px, 12px);
 	}
 }
+
 groupCallMenuToggleSmall: CallButton(groupCallSettingsSmall) {
 	button: IconButton(groupCallSettingsInner) {
 		icon: icon {{ "calls/calls_more", groupCallIconFg }};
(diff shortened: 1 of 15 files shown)


ilya-fedin:
The PR has conflicts

paterkleomenis:
Superseded by #31263.

PR #30302, opened 2026-02-18
Speed up app startup (use faster containers for emoji maps)

futpib:
This shaved off about a second of startup time for me.
The thing that slowed down the startup was [`EmojiListWidget::refreshCustom`](https://github.com/telegramdesktop/tdesktop/blob/2f80db329bfb44639376396ea4cdc46c8d189bca/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L2294) inserting ~8k of custom emojis where each subsequent insert was slower than previous


john-preston:
Thanks!

PR #30303, opened 2026-02-18
Add screen reader support for country select box

rezabakhshilaktasaraei:
## Summary

Exposes painted country rows in the country select box as virtual accessibility children, making the list navigable by screen readers.

### Changes (`country_select_box.cpp`)
- Overrides `accessibilityChildCount()`, `accessibilityChildName()`, `accessibilityChildRect()`, `accessibilityChildState()`, and `accessibilityChildRole()` on the inner widget.
- Each country row becomes a `ListItem` accessible child.
- In phone mode, the name combines country name + dial code (e.g. "Iran, +98").
- In non-phone mode, only the country name is used.
- Fires `accessibilityChildFocused()` when selection changes via keyboard.
- Announces selected row on `focusInEvent`.

### Dependencies
- Requires [lib_ui#278](https://github.com/desktop-app/lib_ui/pull/278) for `Accessible::Item` infrastructure.

### No new lang keys
Uses only existing `lng_country_*` keys — no new strings added.

## Test plan
- [ ] Open country selector (phone registration or settings) with a screen reader
- [ ] Navigate rows with arrow keys — screen reader announces country name (and dial code in phone mode)
- [ ] Verify hit-testing: clicking a row with screen reader cursor selects the correct country

### DIFF
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -7904,4 +7904,6 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 
 "lng_mac_hold_to_quit" = "Hold {text} to Quit";
 
+"lng_sr_country_column_name" = "Country name";
+
 // Keys finished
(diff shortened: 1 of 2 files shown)


john-preston:
Thanks!

PR #30308, opened 2026-02-18
Speed up app startup (defer sticker reads via crl::on_main chain)

futpib:
Chain sticker set reads via separate `crl::on_main` calls instead of running them all synchronously in a single `crl::on_main` callback. Since `crl::on_main` dispatches through the Qt event loop, paint events can be processed between each step, allowing the UI to render before sticker loading completes.

This moves first paint from ~2807ms to ~1294ms in my tests.

### DIFF
--- a/Telegram/SourceFiles/main/main_session.cpp
+++ b/Telegram/SourceFiles/main/main_session.cpp
@@ -181,6 +181,55 @@ Session::Session(
 		_selfUserpicView = view.cloud;
 	}, lifetime());
 
+	// Storage::Account uses Main::Account::session() in those methods.
+	// So they can't be called during Main::Session construction.
+	// They are deferred via crl::on_main which fires after the
+	// constructor returns and _session is set.
+	//
+	// Steps are chained via crl::on_main so that paint events
+	// can be processed between heavy file reads.
+	const auto steps = std::make_shared<std::vector<Fn<void()>>>(
+		std::initializer_list<Fn<void()>>{
+		[=] {
+			local().readInstalledStickers();
+		}, [=] {
+			local().readInstalledMasks();
+		}, [=] {
+			local().readInstalledCustomEmoji();
+		}, [=] {
+			data().stickers().notifyUpdated(Data::StickersType::Stickers);
+			data().stickers().notifyUpdated(Data::StickersType::Masks);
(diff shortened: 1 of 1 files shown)


Telegram/SourceFiles/main/main_session.cpp
```diff
@@ -181,6 +181,62 @@ Session::Session(
...
+		local().readFavedStickers();
+		local().readSavedGifs();
+	});
+	steps->push_back([=] {
+		data().stickers().notifyUpdated(Data::StickersType::Stickers);
+		data().stickers().notifyUpdated(Data::StickersType::Masks);
+		data().stickers().notifyUpdated(Data::StickersType::Emoji);
+		data().stickers().notifySavedGifsUpdated();
+		DEBUG_LOG(("Init: Account stored data load finished."));
+	});
```
ilya-fedin: Couldn't you supply them in the steps variable constructor and use std::initializer_list/std::array?

futpib: Fixed, also replaced InvokeQueued with crl::on_main, works the same

23rd:
Can it be moved to some scope? x_x
A private method for example...

ilya-fedin:
> Can it be moved to some scope? x_x

fwiw he is working with preston on a lib_crl wrapper afaik, so in the end it will likely be way less code and moving may be unneeded

Telegram/SourceFiles/main/main_session.cpp
```diff
@@ -181,6 +181,55 @@ Session::Session(
...
 
+	// Storage::Account uses Main::Account::session() in those methods.
+	// So they can't be called during Main::Session construction.
+	// They are deferred via crl::on_main which fires after the
+	// constructor returns and _session is set.
+	//
+	// Steps are chained via crl::on_main so that paint events
+	// can be processed between heavy file reads.
+	const auto steps = std::make_shared<std::vector<Fn<void()>>>(
+		std::initializer_list<Fn<void()>>{
```
ilya-fedin, line 192: What I mean is that you can replace std::vector with std::initializer_list or std::array. No mutation = no need in vector.

john-preston:
Thanks!

PR #30310, opened 2026-02-18
Speed up app startup (remove redundant cache and pre-warming)

futpib:
Iv::ShowButton() and LocationPicker::Available() each cache the result of Webview::Availability() in a static local.
These caches will be redundant if/when desktop-app/lib_webview#136 is merged.

TLDR: look at https://github.com/desktop-app/lib_webview/pull/136 first

### A better idea: https://github.com/desktop-app/lib_webview/pull/136#issuecomment-3921071476


Telegram/SourceFiles/core/application.cpp
```diff
@@ -325,11 +324,6 @@ void Application::run() {
 	// Create mime database, so it won't be slow later.
 	QMimeDatabase().mimeTypeForName(u"text/plain"_q);
 
-	// Check now to avoid re-entrance later.
```
ilya-fedin, line 328: You're returning re-entrance. That's critical, the app is going to crash.

PR #30313, opened 2026-02-18
Speed up app startup (build emoji replaces in bg thread)

futpib:
Move InstantReplaces::Default() trie construction off the main thread by dispatching it via crl::async.

This is a ~29ms startup speedup in my setup which takes ~600ms in total on futpib:faster-startup branch

### DIFF
--- a/Telegram/SourceFiles/core/application.cpp
+++ b/Telegram/SourceFiles/core/application.cpp
@@ -92,6 +92,8 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 #include "styles/style_window.h"
 
 #include <QtCore/QStandardPaths>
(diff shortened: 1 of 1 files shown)


john-preston:
@futpib For me its 1ms even in unoptimized Debug build, but OK. It feels like its better to move it inside Ui::Emoji::Init? That way it begins earlier and has more time before app asks for that value. Also this will move it to emoji init, which feels semantically correct.

futpib:
I'll double check it then, I may have misattributed the delay or maybe it's caused by something specific to my system or my telegram accounts

john-preston:
No, it's a good idea anyway, I have a powerful (most powerful?) CPU.

futpib:
Moved to https://github.com/desktop-app/lib_ui/pull/290

PR #30314, opened 2026-02-18
Speed up app startup (cache Webview::Availability)

futpib:
Iv::ShowButton() and LocationPicker::Available() each called Webview::Availability() separately (~200ms each on Linux). Replace the per-caller static caches with a single shared cache in Core::CachedWebviewAvailability(), reducing startup from two ~200ms calls to one.

Previous failed attempt at this: https://github.com/desktop-app/lib_webview/pull/136 + https://github.com/telegramdesktop/tdesktop/pull/30310

### DIFF
--- a/Telegram/CMakeLists.txt
+++ b/Telegram/CMakeLists.txt
@@ -482,6 +482,7 @@ PRIVATE
     chat_helpers/ttl_media_layer_widget.h
     core/application.cpp
     core/application.h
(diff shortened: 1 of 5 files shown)


ilya-fedin (approved):
> ~200ms each on Linux

the price of converting async api into a sync one with a nested event loop :sob:

john-preston:
Thanks!

PR #30317, opened 2026-02-19
Freeze chat list order while mouse hovers over it

### LINKED ISSUE #1504: Order of chats change too quickly
When I'm subscribed to several very active chats, the chat order changes too quickly and it's too confusing and cumbersome to switch. I suggest the following change:

When new activity comes to a chat, raise it if one of the following conditions is true:
- Mouse is not over the chat list.
- Position of chat > number of visible chats in the list.

And I would add an option to ignore mouse altogether (I've only mentioned it to have a behavior that people expect).

futpib:
When subscribed to many active chats, the list reorders too quickly, causing misclicks. Freeze the visual order while the mouse is over the chat list and replay deferred reorderings when the mouse leaves or after 5 seconds of inactivity.

Fixes #1504.

  ```mermaid
  stateDiagram-v2                                                                                                                  
      [*] --> Unfrozen                                                                                                               
      Unfrozen --> Frozen: enterEvent / mouseMove                                                                                    
      Frozen --> Frozen: mouseMove to new position (restart 5s timer)
      Frozen --> Unfrozen: leaveEvent / dragLeft                                                                                     
      Frozen --> Unfrozen: timer fires (5s no movement)                                                                              
      Frozen --> Unfrozen: list switch (refreshShownList) 
  ```

### DIFF
--- a/Telegram/SourceFiles/dialogs/dialogs_indexed_list.cpp
+++ b/Telegram/SourceFiles/dialogs/dialogs_indexed_list.cpp
@@ -194,6 +194,20 @@ void IndexedList::adjustNames(
 	}
 }
 
+void IndexedList::freeze() {
+	_list.freeze();
+	for (auto &[ch, list] : _index) {
+		list.freeze();
+	}
+}
+
+void IndexedList::unfreeze() {
+	_list.unfreeze();
+	for (auto &[ch, list] : _index) {
+		list.unfreeze();
+	}
+}
+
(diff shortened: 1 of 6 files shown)


john-preston:
That's nice! Does it work as you would expect?

futpib:
Now it does:


https://github.com/user-attachments/assets/8f593766-5e0b-4a70-9862-49da2599abfc

john-preston:
Thanks!

PR #30324, opened 2026-02-21
typo: Increase the `size` limit on compressed images...

Nusab19:
A tiny typo that I've been seeing for quite a long time. Nobody probably noticed it, so here's the pr.

Just one character change. Won't be much of a thing to review.

**_Peace!_**


Aokromes:
not a typo. (more people noticed that before you)

PR #30325, opened 2026-02-21
Convert https://t.me/ URLs to tg:// scheme in start URL handling

futpib:
Before this fix `Telegram https://t.me/durov/6` did nothing, after this fix it opens the post


john-preston:
Thanks!

PR #30340, opened 2026-02-26
Add accessible name to search button in narrow layout

rezabakhshilaktasaraei:
## Summary

- The search icon button shown when the filters sidebar is active had no accessible name, making it appear as an unnamed button to screen readers.
- Sets `lng_dlg_filter` ("Search") as the accessible name for `_searchForNarrowLayout`.

## Test plan
- [x] Enable filters sidebar (create a folder)
- [x] Navigate with Tab using a screen reader — the search button should announce "Search" instead of just "button"


john-preston:
Thanks!

PR #30350, opened 2026-02-28
Rename accessibility lang keys to lng_sr_ prefix

rezabakhshilaktasaraei:
## Summary

- Rename all accessibility-related lang keys from `lng_` to `lng_sr_` prefix for consistency with the screen reader naming convention established in recent accessibility PRs.

**Renamed keys:**
- `lng_main_menu` → `lng_sr_main_menu`
- `lng_filter_unread_chats` → `lng_sr_filter_unread_chats`
- `lng_minimize_window` → `lng_sr_minimize_window`
- `lng_maximize_window` → `lng_sr_maximize_window`
- `lng_restore_window` → `lng_sr_restore_window`
- `lng_go_back` → `lng_sr_go_back`
- `lng_phone_number` → `lng_sr_phone_number`
- `lng_settings_scale` → `lng_sr_settings_scale`

## Test plan
- [x] Verify screen reader announcements still work correctly for all renamed keys
- [x] Verify no broken references to old key names remain

### DIFF
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -22,9 +22,9 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 "lng_menu_my_stories" = "My Stories";
 "lng_menu_my_groups" = "My Groups";
 "lng_menu_my_channels" = "My Channels";
-"lng_main_menu" = "Main menu";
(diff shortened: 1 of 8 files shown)


john-preston:
Renaming is not a good idea after phrases land on the cloud translation platform, this will just cancel existing translations usage and double the added keys on the platform till the old ones will be manually deleted, which shouldn't happen fast, because they're used in deployed versions.

Issue #30375, opened 2026-03-02
"Rename file" button overlaps "Hide with Spoiler" in media send menu


Neurotoxin001:
### Steps to reproduce

When trying to apply a spoiler to a media file before sending, the context menu is displayed incorrectly — the "Rename file" button pops up and overlaps the "Hide with Spoiler" button.

Furthermore, the "Rename file" button should not be present for compressed images at all. Unlike videos, which keep their given name when downloaded (e.g., "given_name.mp4"), images do not have a displayed name and are simply saved as "download_date.jpg".

1. Open any chat and select a media file to send;

2. Right-click on the attached file in the preview window;

3. Observe the context menu – the rename button covers the spoiler button.

### Expected behaviour

Menu items do not overlap. The "Rename file" option should be absent for images, because pictures do not have a displayed name and are saved with the download date as their name (unlike videos, which keep their given name).

### Actual behaviour

The "Rename file" button pops up right on top of the "Hide with Spoiler" button — overlapping it.

https://github.com/user-attachments/assets/6453f786-3588-471e-9691-0a1b4170b908

### Operating system

Windows 11

### Version of Telegram Desktop

6.6.1

### Installation source

Static binary from official website

### Logs

```shell

```

dajotim937:
??????
Wtf.
3 weeks, no fix for this?

And what's the point of naming/renaming when it still uses "photo_yyyy-mm-dd_hh-mm-ss" naming template when you download pic?

dajotim937:
<img width="556" height="156" alt="Image" src="https://github.com/user-attachments/assets/e8fb59d5-14b5-4c15-a2eb-96c2772ec12c" />

So. Are you fucking serious? 4 weeks of "fixing" and the best you could do it's now 2 context menu with duplicated "hide with spoiler"?
Is there no testing at all? Or there is AI testing? jfc

dajotim937:
Btw you should keep hide with spoiler as first option. There is no "replace" when you send image first time. If it's wrong image you just cancel and drap&drop or ctrl+v correct one. So it makes no sense to make "replace" as first option. And you can open edit with just clicking on image. But there is no alternative to hide with spoiler. Especially when you need to hide only one image of many.

dajotim937:
So, after 1 month of "fixing" now right click open both context menu and image editing.
<details>

[...]

PR #30385, opened 2026-03-03
Fix send-preview rename/spoiler behavior for media and document attachments before sending

Neurotoxin001:
This commit fixes the send-preview rename/spoiler behavior for media and document attachments before sending.

### What was fixed

* Fixed the overlap issue (https://github.com/telegramdesktop/tdesktop/issues/30375): for media previews, actions now appear in a proper menu flow (instead of conflicting behavior);
* For **videos in media mode** (not “send as document”), the menu now supports both: **Rename file** and **Hide with Spoiler**;
* For **photos in media mode**, rename is not shown (as intended, because photos do not have a displayed name and are simply saved as "download_date.jpg".);
* For **photos/videos in document mode**, rename is available and works for both single items and albums;
* Improved default filename logic when source has no normal path/name (for example clipboard images), so rename dialog starts with a sensible name and keeps extension behavior.

### DIFF
--- a/Telegram/SourceFiles/boxes/edit_caption_box.cpp
+++ b/Telegram/SourceFiles/boxes/edit_caption_box.cpp
@@ -489,7 +489,9 @@ void EditCaptionBox::rebuildPreview() {
 			gifPaused,
 			file,
 			[=](Ui::AttachActionType type) {
-				return (type != Ui::AttachActionType::EditCover) || _isVideo;
+				return (type == Ui::AttachActionType::Rename)
+					? false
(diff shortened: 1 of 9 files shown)


23rd:
One feature/fix per commit.

ilya-fedin:
image

ilya-fedin:
Please address 23rd's concern and use rebase instead of merge

Neurotoxin001:
Removed the fix “Improved default logic for file names when the source does not have a normal path/name (e.g., images from the clipboard), so the rename dialog box launches with a reasonable name and preserves the extension behavior,” so now there is essentially only one fix related to the rename button.

PR #30387, opened 2026-03-04
Do not exit fullscreen mode when applying video quality

comdev1337:
Fixes a bug where the video player exits the fullscreen mode upon changing the video quality.

`applyVideoQuality()` calls `clearStreaming()` which sets `_fullScreenVideo = false` every time it's invoked

Steps to reproduce:
1) Open pretty much any video, i.e. https://t.me/ClashReport/73603
2) Go fullscreen
3) Try changing video quality
Immediate result: video stays "fullscreen", black borders disappear, now another double click is required to get back to the fullscreen mode.


john-preston:
@comdev1337 I think we need to restore _fullScreenVideo only if we could re-init streaming, but if we failed for some reason and we call re-display-content, then we shouldn't.

So I'd place restoring in an `else` branch to the !init calls.

comdev1337:
@john-preston updated

PR #30397, opened 2026-03-04
Fix duplicate subtitle widget in group call panel

paterkleomenis:
The `_subtitle.create()` rpl chain calls `setupMembers()` synchronously during the FlatLabel constructor. `setupMembers()` triggers geometry updates that re-enter `refreshTitle()` while `object_ptr::create()` has not yet assigned the new widget to `_subtitle`, so a second subtitle is created and left as an orphaned visible child.

Moving `setupMembers()` before `_subtitle.create()` ensures `_members` exists before the rpl chain fires, preventing the re-entrancy.


ilya-fedin:
pr conflicts...

paterkleomenis:
> pr conflicts...

i am sorry, now it should be good

PR #30446, opened 2026-03-14
Fix startchannel/startgroup deeplinks not working for bots with Threaded Mode enabled.

rdmcd:
When a bot has Threaded Mode (forum) enabled, the isForum() check in showPeerByLinkResolved() intercepts the flow before reaching the AddToGroup/AddToChannel handler, causing the app to open the bot's chat instead of showing the "add bot to channel/group" popup.

### Problem

When opening a ?startchannel (or ?startgroup) deeplink for a bot that has Threaded Mode (forum topics) enabled, the app opens the bot's chat instead of showing the "Add bot to channel/group" popup.

### Steps to reproduce:
1. Find a bot with Threaded Mode enabled (set via BotFather)
2. Open t.me/botusername?startchannel
3. Expected: "Choose a channel" popup appears
4. Actual: Bot's chat (forum view) opens instead

### Cause

In showPeerByLinkResolved(), the peer->isForum() branch (line 678) is evaluated before the AddToGroup/AddToChannel branch (line 741). For bots with Threaded Mode, isForum() returns true, so the forum branch intercepts the flow and the   add-to-channel popup is never reached.

The Boost resolve type was already excluded from this check, but AddToGroup and AddToChannel were not.

### Fix

Add AddToGroup and AddToChannel to the exclusion list alongside Boost, so these deeplinks correctly fall through to AddBotToGroupBoxController::Start().

### DIFF
--- a/Telegram/SourceFiles/window/window_session_controller.cpp
+++ b/Telegram/SourceFiles/window/window_session_controller.cpp
@@ -675,22 +675,6 @@ void SessionNavigation::showPeerByLinkResolved(
 		showPeerInfo(peer, params);
 	} else if (resolveType == ResolveType::HashtagSearch) {
 		searchMessages(info.text, peer->owner().history(peer));
-	} else if (peer->isForum() && resolveType != ResolveType::Boost) {
-		if (!msgId || !useRequestedMessageId) {
-			applyBotStartToken();
-			parentController()->showForum(peer->forum(), params, msgId);
-		} else if (const auto item = peer->owner().message(peer, msgId)) {
-			showMessageByLinkResolved(item, info);
-		} else {
-			const auto callback = crl::guard(this, [=] {
-				if (const auto item = peer->owner().message(peer, msgId)) {
-					showMessageByLinkResolved(item, info);
-				} else {
-					showPeerHistory(peer, params, msgId);
-				}
-			});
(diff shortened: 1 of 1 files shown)


john-preston:
What about fixing the order of options instead of complicating the condition? Would it work?

rdmcd:
Good point! 

Also while investigating this, I found the same bug also affects ?startapp (BotApp) and ShareGame deeplinks — the isForum() check intercepts all bot-specific resolve types, not just AddToGroup/AddToChannel.

Updated the fix to move all three bot-specific handlers (BotApp, ShareGame, AddToGroup/AddToChannel) above the isForum() check.

john-preston:
Ok, but this is a bit more changes than required and also we could simplify a bit:

Just take the original code, take the `} else if (peer->isForum() && resolveType != ResolveType::Boost) {` branch and move it down, we can put it right before the `} else if (resolveType == ResolveType::Mention) {` branch.

If after that we move the `} else if (resolveType == ResolveType::Boost && peer->isChannel()) {` branch above those two, we could even remove the `&& resolveType != ResolveType::Boost` for the `isForum()` case.

rdmcd:
Applied.

john-preston:
Thank you!

PR #30470, opened 2026-03-23
Fix personal photo hint shown when setting group/channel photo

### LINKED ISSUE #30467: Setting profile picture for a group or a channel shows hint for setting a private user photo
### Steps to reproduce

1. Open a group or a channel info page
2. Right click on the photo and select the option to set a new photo
3. Choose a photo from the file picker

### Expected behaviour

No unexpected hints are shown

### Actual behaviour

A hint about setting a private user profile photo is shown, which makes no sense in this context:

Image

### Operating system

Fedora Workstation 43 GNOME

6.6.2

### Installation source

Flatpak

### Crash ID

_No response_

### Logs

```shell

```

arelove:
Fixes #30467

When you set a photo for a group or channel, a hint saying 
"Only you will see this photo..." was shown — which makes no 
sense for groups or channels, only for personal user photos.

The issue is in `editorData` lambda inside `setupUserpicButton` 
in `info_profile_top_bar.cpp`, where the hint phrase was always 
used unconditionally. Fixed by adding a null-check on the peer 
type, so the hint is only shown for non-self users — same as the 
existing logic in `userpic_button.cpp`.

**Before:** when setting a photo for a group or channel, the editor 
showed a hint intended only for personal user photos ("Only you will 
see this photo and it will replace any photo...").

**After:**
image

### DIFF
--- a/Telegram/SourceFiles/info/profile/info_profile_top_bar.cpp
+++ b/Telegram/SourceFiles/info/profile/info_profile_top_bar.cpp
@@ -1136,13 +1136,17 @@ void TopBar::setupUserpicButton(
 			: _peer->name();
 		const auto phrase = (type == ChosenType::Suggest)
 			? &tr::lng_profile_suggest_sure
-			: &tr::lng_profile_set_personal_sure;
+			: (user && !user->isSelf())
+			? &tr::lng_profile_set_personal_sure
+			: nullptr;
 		return Editor::EditorData{
-			.about = (*phrase)(
-				tr::now,
-				lt_user,
-				tr::bold(name),
-				tr::marked),
+			.about = (phrase
+				? (*phrase)(
+					tr::now,
+					lt_user,
+					tr::bold(name),
+					tr::marked)
+				: TextWithEntities()),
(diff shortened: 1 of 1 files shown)


Telegram/SourceFiles/info/profile/info_profile_top_bar.cpp
```diff
@@ -1136,13 +1136,17 @@ void TopBar::setupUserpicButton(
 			: _peer->name();
 		const auto phrase = (type == ChosenType::Suggest)
 			? &tr::lng_profile_suggest_sure
-			: &tr::lng_profile_set_personal_sure;
+			: (user && !user->isSelf())
```
23rd:
```suggestion
			: (user && !user->isSelf() && !_peer->isBot())
```

23rd:
It would be great to squash my commit with yours.

23rd:
> It would be great to squash my commit with yours.

Oh, @Aokromes suggested me I can do it by myself with Github UI. Waiting for some success build.

PR #30491, opened 2026-03-30
Fix cursor staying hidden after closing media viewer on macOS

### LINKED ISSUE #30490: Mouse pointer remains hidden after exiting the video or media display
### Steps to reproduce

1. Open any video in the media viewer.
2. Move the mouse at least once inside the viewer.
3. Leave the mouse still for ~3 seconds until the controls fade out and the cursor disappears.
4. Press the left or right arrow key to navigate to another item (without moving the mouse).
5. Close the media viewer by pressing Escape (without moving the mouse).

### Expected behaviour

The mouse cursor reappears after closing the media viewer.

### Actual behaviour

The mouse cursor remains invisible after the viewer closes. It does not reappear on mouse movement. Switching to another application temporarily restores it, but switching back to Telegram Desktop hides it again. The issue persists for the entire session — quitting and relaunching Telegram Desktop resolves it, as does opening another media viewer and exiting (though this requires clicking blindly on a video with an invisible pointer). This is a macOS [NSCursor hide]/[NSCursor unhide] reference count imbalance caused by the overlay window hiding while Qt::BlankCursor is still active.

### Operating system

macOS 26.4

12.5.278815 Stable

### Installation source

### Crash ID

_No response_

### Logs

```shell

```

SnoElement:
Fixes #30490

## Root cause

When video playback controls auto-hide in the media viewer, the overlay widget's cursor is set to `Qt::BlankCursor`. On macOS this calls `[NSCursor hide]`, which increments a global hide counter.

If the user then navigates to another item using the arrow keys without moving the mouse, the cursor is never restored before closing:

- The arrow key handler only called `activateControls()` if the controls hide timer was still active — but once controls have fully hidden, the timer has already fired, so `activateControls()` was skipped entirely.
- Closing the viewer called `clearBeforeHide()`, which reset `_controlsState` but never reset the widget cursor — so the window was hidden while `Qt::BlankCursor` was still active.
- This left `[NSCursor hide]` unbalanced at the application level. The cursor remains invisible for the entire session: switching to another app temporarily restores it, but switching back to Telegram Desktop hides it again. Only quitting resolves it.

## Fix

Three changes in `media_view_overlay_widget.cpp`:

1. **Arrow key navigation always calls `activateControls()`** — removed the `_controlsHideTimer.isActive()` guard so controls and cursor are always restored when navigating.
2. **`activateControls()` calls `updateCursor()` immediately** — cursor is restored at once when transitioning to `ControlsShowing`, not only after the animation completes.
3. **`clearBeforeHide()` resets the cursor** — added `setCursor(style::cur_default)` before the window hides, ensuring `[NSCursor hide]`/`[NSCursor unhide]` is always balanced.

---

> This fix was developed with the assistance of GitHub Copilot (AI).

### DIFF
--- a/Telegram/SourceFiles/media/view/media_view_overlay_widget.cpp
+++ b/Telegram/SourceFiles/media/view/media_view_overlay_widget.cpp
@@ -5919,9 +5919,7 @@ void OverlayWidget::handleKeyPress(not_null<QKeyEvent*> e) {
 			handleDocumentClick();
 		}
 	} else if (key == Qt::Key_Left) {
-		if (_controlsHideTimer.isActive()) {
-			activateControls();
-		}
+		activateControls();
 		moveToNext(-1);
 	} else if (key == Qt::Key_H && !_stories) {
 		if (_flip & Qt::Horizontal) {
@@ -5944,9 +5942,7 @@ void OverlayWidget::handleKeyPress(not_null<QKeyEvent*> e) {
 			redisplayContent();
 		}
 	} else if (key == Qt::Key_Right) {
-		if (_controlsHideTimer.isActive()) {
-			activateControls();
-		}
+		activateControls();
 		if (!moveToNext(1) && _stories) {
 			storiesClose();
 		}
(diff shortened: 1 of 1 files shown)


Aokromes:
but... your fix is for telegram messenger not telegram desktop?

SnoElement:
Yup, my bad. Not replicable in Telegram Desktop.

Issue #1121, opened 2015-09-23
DPI scaling in Windows


Snnlk:
Would it be possible to allow the app to automatically change the DPI scaling at runtime depending on which screen it is in?

This is an issue in multi-screen environments with different DPI scaling per screen.

Is this some limitation of Qt?

GLiberman:
+1! I have same problem

xcover:
+1. Having 2x Scaling on the Surface makes Telegram only usable on the external (main) screen.

maik-s:
+1! I have same problem

Matze177:
+1. Same problem here.

karlobencic:
Also facing this issue, can't use telegram on multiple monitors with different scale set in Windows 10 screen settings.

PaoloFalomo:
#### I can confirm this on my *DELL XPS 15* (monitor 3K + external ultrawide monitor)

**XPS Display**
![image](https://cloud.githubusercontent.com/assets/10535089/16306403/deb1f2ee-395d-11e6-87cd-99cf76bb0ab7.png)

**External Monitor** (beside there's a normal explorer window.
![image](https://cloud.githubusercontent.com/assets/10535089/16306650/ab261a80-395e-11e6-887a-d91fbb79055e.png)

PaoloFalomo:
I still have this issue with last version (even reinstall didn't worked). There's no auto DPI scaling.

PaoloFalomo:
@auchri this is still not working for me (even with the last update)

ssorokin:
+1 still an issue, making telegram unusable on a high-dpi laptop screen. What's the status of this?

PaoloFalomo:
@ssorokin I confirm.
No response from @telegramdesktop or @auchri or even from @john-preston ... this is a bug and need fix. Maybe there's something more "urgent" right now  😟 .

I had configured the repo for working on the code and may spot the bug but i have to admit that i haven't had much time to work on it 🚶

john-preston:
@PaoloFalomo @telegramdesktop can't respond, because it is an organization account now 😌

This is not something that can be changed any easy way: all the styles are configured when the app is launched according to the current interface scale level chosen in Settings (which is chosen by the current dpi), so at this time there is no way to change the interface scale factor in runtime.

PaoloFalomo:
@john-preston Oh! Ok 😸 for the telegramdesktop user, thx for the info eheh.

Btw i had a similar problem about this developing other win apps.

[...]

PR #30492, opened 2026-03-30
POC: Wayland fractional scale support

zvova7890:
Since Telegram moved to Qt6, it has broken fractional scaling support on both X11 and Wayland. With Qt5, fractional scaling works as usual on X11. So I rebuilt Telegram myself with patches that force XWayland and built it with Qt5, thus fixing recurring compilation issues with new code.

Lately, I was wondering if current AI capabilities would be enough to fix the infamous "unfixable" Telegram problem? Apparently, yes! Just throw in an idea and see the magic. If nobody wants to fix it, let AI do it, right?

Anyway, after a ~day of iterations and fixing crashes/glitches/etc., I ended up with a patch that allows Telegram to run with a crystal-clear picture under a KDE Wayland session with 1.5 and 2.5 scale factors. Even enabling experimental fractional scaling works as intended.

There are still crashes/glitches/hangs because many places are not obviously dependent on the DevicePixelRatio. All usages of DevicePixelRatio should be reviewed, but there is a lot of code, and this can be done gradually, as ideally these changes should not break integer scaling. For example, I spent an hour figuring out why the user profile image history was trying to load images infinitely with no success. It turned out there was a check like `if (target.size() == requested) cached = true;` and computation just don't match because of fractional scale.

So basically, I don’t think anyone would be interested in merging this. Moreover, I don’t think I want to make it production-ready. I’m just sharing this as a proof of concept that says "Hey, it’s possible".

About 98% of the refactoring/bug fixing was done by GPT 5.4. I’m familiar with Qt, but not with the Telegram codebase at all. AI just makes it 9999x faster. God bless RAM.

There are also parts in `lib_ui` and `lib_lottie`; I’ve included the diff here if anyone wants to check it out.

Feel free to close this PR at any time!

[tg-fraction-scale.diff.tar.gz](https://github.com/user-attachments/files/26446534/tg-fraction-scale.diff.tar.gz) diff for v6.6.4

image
image

### DIFF
--- a/Telegram/SourceFiles/core/sandbox.cpp
+++ b/Telegram/SourceFiles/core/sandbox.cpp
@@ -180,7 +180,7 @@ void Sandbox::launchApplication() {
 }
 
 void Sandbox::setupScreenScale() {
-	const auto ratio = devicePixelRatio();
+	auto ratio = devicePixelRatio();
 	LOG(("Global devicePixelRatio: %1").arg(ratio));
 	const auto logEnv = [](const char *name) {
 		const auto value = qEnvironmentVariable(name);
@@ -198,7 +198,7 @@ void Sandbox::setupScreenScale() {
 	logEnv("QT_USE_PHYSICAL_DPI");
 	logEnv("QT_FONT_DPI");
 
-	const auto useRatio = std::clamp(qCeil(ratio), 1, 3);
+	const auto useRatio = std::clamp(ratio, 1., 3.);
 	style::SetDevicePixelRatio(useRatio);
 
 	const auto screen = Sandbox::primaryScreen();
@@ -218,6 +218,7 @@ void Sandbox::setupScreenScale() {
 			style::kScaleMin,
 			style::MaxScaleForRatio(useRatio)));
 	}
+	LOG(("DevicePixelRatio: %1").arg(style::DevicePixelRatio()));
 	LOG(("DevicePixelRatio: %1").arg(useRatio));
 	LOG(("ScreenScale: %1").arg(cScreenScale()));
 }
--- a/.gitmodules
+++ b/.gitmodules
@@ -24,10 +24,10 @@
 	url = https://github.com/desktop-app/codegen.git
 [submodule "Telegram/lib_ui"]
 	path = Telegram/lib_ui
-	url = https://github.com/desktop-app/lib_ui.git
+	url = git@github.com:zvova7890/lib_ui.git
 [submodule "Telegram/lib_lottie"]
 	path = Telegram/lib_lottie
-	url = https://github.com/desktop-app/lib_lottie.git
+	url = git@github.com:zvova7890/lib_lottie.git
 [submodule "Telegram/lib_tl"]
 	path = Telegram/lib_tl
 	url = https://github.com/desktop-app/lib_tl.git
--- a/Telegram/SourceFiles/api/api_chat_invite.cpp
+++ b/Telegram/SourceFiles/api/api_chat_invite.cpp
@@ -154,7 +154,7 @@ void ConfirmSubscriptionBox(
 			creditsIconSize,
 			1.5);
 	state->frame = QImage(
-		Size(photoSize * style::DevicePixelRatio()),
+		Size(style::DevicePixels(photoSize)),
 		QImage::Format_ARGB32_Premultiplied);
 	state->frame.setDevicePixelRatio(style::DevicePixelRatio());
 	const auto options = Images::Option::RoundCircle;
--- a/Telegram/SourceFiles/boxes/background_box.cpp
+++ b/Telegram/SourceFiles/boxes/background_box.cpp
@@ -657,7 +657,7 @@ void BackgroundBox::Inner::validatePaperThumbnail(
 		} else if (!paper.data.backgroundColors().empty()) {
 			paper.thumbnail = Ui::PixmapFromImage(
 				Ui::GenerateBackgroundImage(
-					st::backgroundSize * style::DevicePixelRatio(),
+					style::DevicePixels(st::backgroundSize),
 					paper.data.backgroundColors(),
 					paper.data.gradientRotation()));
 			paper.thumbnail.setDevicePixelRatio(style::DevicePixelRatio());
--- a/Telegram/SourceFiles/boxes/background_preview_box.cpp
+++ b/Telegram/SourceFiles/boxes/background_preview_box.cpp
@@ -121,8 +121,9 @@ constexpr auto kMaxWallPaperSlugLength = 255;
 	const auto takeHeight = (width > height)
 		? size
 		: (height * size / width);
-	const auto ratio = style::DevicePixelRatio();
-	return Images::Prepare(image, QSize(takeWidth, takeHeight) * ratio, {
+	return Images::Prepare(image, QSize(
+		style::DevicePixels(takeWidth),
+		style::DevicePixels(takeHeight)), {
 		.options = Images::Option::TransparentBackground | blur,
 		.outer = { size, size },
 	});
@@ -725,7 +726,7 @@ void BackgroundPreviewBox::applyForPeer() {
 	} else if (_forBothOverlay) {
 		return;
 	}
-	const auto size = this->size() * style::DevicePixelRatio();
+	const auto size = style::DevicePixels(this->size());
 	const auto bg = Images::DitherImage(
 		Images::BlurLargeImage(
 			Ui::GrabWidgetToImage(this).scaled(
--- a/Telegram/SourceFiles/boxes/filters/edit_filter_links.cpp
+++ b/Telegram/SourceFiles/boxes/filters/edit_filter_links.cpp
@@ -359,7 +359,9 @@ PaintRoundImageCallback ChatRow::generatePaintUserpicCallback(
 			int outerWidth,
 			int size) mutable {
 		const auto wide = size + style::ConvertScale(3);
-		const auto full = QSize(wide, wide) * style::DevicePixelRatio();
+		const auto full = QSize(
+			style::DevicePixels(wide),
+			style::DevicePixels(wide));
 		auto repaint = false;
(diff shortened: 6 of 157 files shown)


Telegram/SourceFiles/core/sandbox.cpp
```diff
@@ -211,13 +211,14 @@ void Sandbox::setupScreenScale() {
 	LOG(("Computed screen scale: %1").arg(screenScale));
 	if (Platform::IsMac()) {
 		// 110% for Retina screens by default.
-		cSetScreenScale((useRatio == 2) ? 110 : style::kScaleDefault);
+		cSetScreenScale((useRatio >= 1.99) ? 110 : style::kScaleDefault);
```
ilya-fedin: Why do you change this macOS-specific code?

ilya-fedin:
If I understand right, style::DevicePixelRatio is still set from QGuiApplication::devicePixelRatio? But AFAIK it returns ceiled value on fractional scales, how does it work?

Telegram/SourceFiles/core/sandbox.cpp
zvova7890: Artefact, never mind

zvova7890:
> If I understand right, style::DevicePixelRatio is still set from QGuiApplication::devicePixelRatio? But AFAIK it returns ceiled value on fractional scales, how does it work?

Yes, but not exactly. It is updated later in RefreshDevicePixelRatioFromWindow when QWindow receives a new scale factor from the compositor. Because of that, Telegram should support dynamically changing the scale at runtime. There were several problems with that, which I found during testing -- all known ones were resolved. Actually, it should be connected to QWindow::screenChanged and update the scale factor every time it changes. In that case, if the scale factor changes due to a monitor settings update or the window moving to another screen with a different scale factor, the UI will stay crisp. In this diff it is not yet done, but on my local branch I've already tested -- it seems to be working fine.

ilya-fedin:
> Actually, it should be connected to QWindow::screenChanged and update the scale factor every time it changes.

That's not really right since on Wayland fractional scale factor is per-window rather than per-screen. The right event is QEvent::DevicePixelRatioChange.

> RefreshDevicePixelRatioFromWindow

I don't see it in this PR :thinking:

zvova7890:
> That's not really right since on Wayland fractional scale factor is per-window rather than per-screen. The right event is QEvent::DevicePixelRatioChange.

Yes, you're right. Despite Qt still sending screenChanged events on Wayland for per‑window fractional scale changes, probably because all other platforms do that and Qt wants to keep as much as possible the same behavior across platforms, DevicePixelRatioChange is a better choice.

> I don't see it in this PR 🤔

https://github.com/telegramdesktop/tdesktop/pull/30492/changes#diff-fb20cd304b852c78660103605c2f357ddfd811e2a29a5260731974e38817f923R84

---
BTW, AFAIK Telegram does not support dynamic scale change on all platforms, right? I was interested whether it was possible even theoretically, and it seems yes, I managed to make core UI functionality adapt dynamically to the scale changes (e.g. moving around monitors with different scales or changing the scale for one monitor). Oh, how many caches, so many caches... There is the cache even for rounded rects :)

ilya-fedin:
> https://github.com/telegramdesktop/tdesktop/pull/30492/changes#diff-fb20cd304b852c78660103605c2f357ddfd811e2a29a5260731974e38817f923R84

Oh, thanks

> BTW, AFAIK Telegram does not support dynamic scale change on all platforms, right? I was interested whether it was possible even theoretically, and it seems yes, I managed to make core UI functionality adapt dynamically to the scale changes (e.g. moving around monitors with different scales or changing the scale for one monitor). Oh, how many caches, so many caches... There is the cache even for rounded rects :)

Yeah, the fact that the app has lots of cache for CPU-based animations is the main blocker to support runtime scale changes and implement #1121. I so far believed that it's an architectural choice that couldn't be fixed at that point in time and only new client written from scratch on some other toolkit could fix it (such as https://web.telegram.org/a/get).

zvova7890:
I've pushed an updated version, with initial dynamic DPR support. So far, what I'm using works +- fine. I think it pretty doable 🤔

ilya-fedin:
> I've pushed an updated version, with initial dynamic DPR support. So far, what I'm using works +- fine.

It conflicts though...

zvova7890:
> It conflicts though...

That's because I'm on v6.6.4, not a head

Azq2:
Great thanks, you save my eyes

ilya-fedin:
FWIW, if Qt was a better toolkit and its maintainers had softer acceptance criteria for new features, it could be fixed on Qt level with way less code changes. There was a 180 lines patch in tdesktop Qt patchset previously that was adapting the high dpi rounding policy for per-window scaling so that it works on Wayland with the fractional scaling protocol:
https://github.com/desktop-app/patches/raw/a405719f0963abf7cb93354a390617c0f0d90f17/qtbase_6.9.1/0019-follow-highdpi-rounding-policy-for-platform-dpr.patch
https://github.com/desktop-app/patches/raw/a405719f0963abf7cb93354a390617c0f0d90f17/qtbase_6.9.1/0020-let-platform-backing-store-scale-transparency-filling.patch
https://github.com/desktop-app/patches/raw/a405719f0963abf7cb93354a390617c0f0d90f17/qtwayland_6.9.1/0001-always-fractional-scale.patch
https://github.com/desktop-app/patches/raw/a405719f0963abf7cb93354a390617c0f0d90f17/qtwayland_6.9.1/0002-scale-transparency-filling.patch

ilya-fedin:
> There was a 180 lines patch in tdesktop Qt patchset previously that was adapting the high dpi rounding policy for per-window scaling so that it works on Wayland with the fractional scaling protocol

I'm thinking about resurrecting them. Does anyone really need it?

ilya-fedin:
@zvova7890 do you plan to continue the work or this could be closed?

zvova7890:
> @zvova7890 do you plan to continue the work or this could be closed?

Is anybody interested in having this upstreamed? When I made this PR, I had the impression that it would not be merged for ideological reasons. Is that true? These are quite large changes that probably nobody wants to deal with.

I just keep rebasing and using these changes all the time at a 1.5 fractional scale. There are a few minor problems, and I also haven’t tested it thoroughly with scales other than 1.5 or 2.0. AFAIR, there were some artifacts at a 1.45 scale when I checked it last time. However, I haven’t seen any crashes or other critical bugs so far. It’s pretty usable.

I still think it’s doable. First, we need to upstream the related changes to libui and convince developers to use those helpers instead of `QSize * dpr` (mostly the reason that changes so huge).

ilya-fedin:
> Is anybody interested in having this upstreamed?

I honestly don't know, it's just current version doesn't look clean. Real fractional scale support would need style::DevicePixelRatio replacement with QWidget::devicePixelRatioF. And then lib_ui has `scaling` branch where preston works in a direction against DPR, by watching DPI updates in runtime. It might be better to work in the direction of Qt patches I posted above? 7.2.6/7.2.7 official builds include them again.

zvova7890:
> It might be better to work in the direction of Qt patches I posted above?

You mean trying to upstream the Qt changes again? To be honest, I don’t believe in that either. This is a dead end. God, I love open source :)

Technically, of course, it would be better to fix tdesktop itself. This would bring crystal-clear fractional scaling support (unlike patching Qt - it is still blurry on Wayland, AFAIR), multi-monitor scaling support, and all the other good stuff. However, it is a little harder to maintain and to land new features, because developers need to test new features at fractional scales and stop doing things like `int scale = devicePixelRatio()` and other stuff that this PR fixes in many places.

Adding a fractional `DevicePixelRatio` and propagating the ratio change event is the simplest part of this PR. The hardcore part begins throughout the entire codebase, where the idea of fractional scaling is simply excluded at the ideological level. This PR is just a POC, so everyone can see two things: 1) it is possible in the near future, and 2) the scale of the changes involved. I didn’t make clean, small PRs because I believed that nobody was interested in having this upstreamed, so initially it was more just for fun.

Does the latest Flatpak contain those Qt patches?

ilya-fedin:
> You mean trying to upstream the Qt changes again?

No, I mean use the official build as-is with that patch

> (unlike patching Qt - it is still blurry on Wayland, AFAIR)

Is it? It should draw 1:1... as long as you have one screen. But even if you have multiple screens only graphics will look a little off, text rendering should be still crystal clear. And then your patch currently doesn't seem to handle multi-screen either given that you rely on global style::DevicePixelRatio.

> because developers need to test new features at fractional scales and stop doing things like `int scale = devicePixelRatio()` and other stuff that this PR fixes in many places

It would be hard to convince preston on it I guess since he wants the UI to be pixel perfect and pixel perfect UI is impossible with fractional DPRs since you can't position a widget on a part of a logical pixel. But if you're ready to continue the work on that, I will try to talk with preston about this and see what he thinks (if he won't just ignore my messages :smile:)

> Does the latest Flatpak contain those Qt patches?

Only the build on flathub-beta. Actually there's also a stable version now but I currently struggle to update it in flatpak due to introduction of rust deps.

zvova7890:
> Is it? It should draw 1:1... as long as you have one screen. But even if you have multiple screens only graphics will look a little off, text rendering should be still crystal clear.

Let me check it again. Maybe recent builds will look better.

> And then your patch currently doesn't seem to handle multi-screen either given that you rely on global style::DevicePixelRatio.

The global `DevicePixelRatio` is updated every time the window moves to another screen. However, multi-window cases are not handled (when main and second windows are on different monitors), such as the media viewer, calls, and so on.

It’s interesting why distro packagers don’t bother embedding the patched Qt into Telegram package. Using the stock one is more convenient, of course, but come on, today, almost all laptops use fractional scaling. A lot people are also using 2K/4K desktop monitors, which quite often require fractional scaling as well. Nobody designs monitors for exact integer scaling (well, except Apple)

zvova7890:
I downloaded the official binary, and it looks pretty solid. The fonts are a little smaller than with native fractional scaling, but it’s still better than the one from the repositories now. So, I think we can close this PR

PR #30514, opened 2026-04-02
Show full title tooltip when chat name is truncated

### LINKED ISSUE #1502: Hovering chat title should show the full title in tooltip
For now we have to resize the window to see... but sometimes even that isn't enough.  Same for the chat info panel, the title is just ellipsised and that's all.

re-void:
## Problem

When a chat title is too long, it gets truncated with ellipsis in the top bar. There is no way to view the full name without resizing the window — and even that doesn't always help.

## Solution

Added `QEvent::ToolTip` handling in `TopBarWidget`. When hovering over the title area, a tooltip with the full chat name is shown **only if the text is actually truncated** (`maxWidth() > availableWidth`).

Works for both regular chats and forum topics.

## Changes

- `history_view_top_bar_widget.h` — added `bool event(QEvent *e) override`
- `history_view_top_bar_widget.cpp` — tooltip logic in `event()`, added `<QtWidgets/QToolTip>` include

Fixes #1502

### DIFF
--- a/Telegram/SourceFiles/history/view/history_view_top_bar_widget.cpp
+++ b/Telegram/SourceFiles/history/view/history_view_top_bar_widget.cpp
@@ -801,6 +801,57 @@ void TopBarWidget::mousePressEvent(QMouseEvent *e) {
 	}
 }
 
+bool TopBarWidget::event(QEvent *e) {
+	if (e->type() == QEvent::ToolTip) {
+		const auto pos = static_cast<QHelpEvent*>(e)->pos();
+		const auto nametop = st::topBarArrowPadding.top();
+		const auto nameheight = st::msgNameStyle.font->height;
+		const auto nameleft = _leftTaken;
+		const auto namewidth = width()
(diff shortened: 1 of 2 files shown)


Telegram/SourceFiles/history/view/history_view_top_bar_widget.cpp
```diff
@@ -801,6 +802,52 @@ void TopBarWidget::mousePressEvent(QMouseEvent *e) {
...
+			nametop,
+			namewidth,
+			nameheight);
+		if (nameRect.contains(pos)) {
+			const auto topic = _activeChat.key.topic();
+			if (topic
+				&& _activeChat.section == Section::Replies) {
+				const auto &topicName = topic->chatListNameText();
+				if (topicName.maxWidth() > namewidth) {
+					QToolTip::showText(
```
ilya-fedin: tdesktop uses AbstractTooltipShower

PR #30532, opened 2026-04-05
Change proxy share link from t.me to tg:// deeplink

roman901:
Currently, when the client tries to share a proxy link (including via QR code), it generates it via `https://t.me/proxy?...`
This is inconvenient - because if I want to share a link, it means Telegram is blocked in my country and it won't open. The same applies to QR scanners on mobile devices.

TelegramSwift on macOS already shares links via deeplink, I suggest unifying the client behavior.


dimonstrik (approved):

john-preston:
This makes sense.

PR #30535, opened 2026-04-06
Fix stale macOS tray icon colors after appearance switches

### LINKED ISSUE #30527: Tray Icon Always Show Black Icon
### Steps to reproduce

After just updated to 6.7.2 tray icon suddenly become black, I have double checked telegram is using light mode instead of dark mode.

Image
Image
Image

### Expected behaviour

Show normal telegram icon in tray

### Actual behaviour

Show dark icon in tray

### Operating system

macOS 26.4

6.7.2

### Installation source

### Crash ID

_No response_

### Logs

```shell

```

KevinHuge:
Fixes #30527 

Summary:
Fix the macOS tray icon refresh path so the menu bar icon updates immediately after system light/dark appearance changes, even while Telegram stays unfocused in the background.

Changes:
The tray icon code was observing status button appearance changes but rendering against NSApp effectiveAppearance on macOS 10.14 and newer. Those two appearance sources do not update in lockstep while Telegram is unfocused, which caused the tray icon to redraw with a stale color until the app was focused. This change makes the renderer use status.button.effectiveAppearance directly so the icon color is derived from the same Cocoa object that emits the refresh signal.

### DIFF
--- a/Telegram/SourceFiles/platform/mac/tray_mac.mm
+++ b/Telegram/SourceFiles/platform/mac/tray_mac.mm
@@ -174,12 +174,7 @@ void UpdateIcon(const NSStatusItem *status) {
 		return;
 	}
 
-	const auto appearance = [&] {
-		if (@available(macOS 10.14, *)) {
-			return [NSApp effectiveAppearance];
-		}
-		return status.button.effectiveAppearance;
-	}();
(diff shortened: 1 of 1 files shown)


john-preston:
Thanks!

PR #30543, opened 2026-04-08
Fix startapp/startattach deeplinks for bots with Threaded Mode.

rdmcd:
Follow-up to #30446 — in that PR I fixed only named bot app links (t.me/bot/appname) by moving isForum() below the BotApp resolve type handler, but missed that ?startapp (default mini app), ?startattach and ?attach deeplinks were handled inside the final else block and still intercepted by isForum().

  - Extracted the three attach bot handlers (attachBotUsername, attachBotMainOpen, attachBotToggleCommand) from the else block into standalone branches above the isForum() check
  - Same reorder approach as #30446 — no condition changes to isForum()

### DIFF
--- a/Telegram/SourceFiles/window/window_session_controller.cpp
+++ b/Telegram/SourceFiles/window/window_session_controller.cpp
@@ -651,9 +651,12 @@ void SessionNavigation::showPeerByLinkResolved(
 
 	// t.me/username/012345 - we thought it was a channel post link, but
 	// after resolving the username we found out it is a bot.
+	// For bots with Threaded Mode the numeric segment is not an app
+	// short name, so don't upgrade to BotApp in that case.
 	const auto resolveType = (bot
 		&& !info.botAppName.isEmpty()
-		&& info.resolveType == ResolveType::Default)
+		&& info.resolveType == ResolveType::Default
+		&& !peer->isForum())
 		? ResolveType::BotApp
 		: info.resolveType;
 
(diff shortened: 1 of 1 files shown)


rdmcd:
Update: rebased on dev (resolved trivial conflict with the recent maySkipConfirmation change) and extended the scope to cover one more case in the same area.

`t.me/bot/<numeric>` for a bot with Threaded Mode was failing with "Bot application not found". The URL parser correctly routes /<digits> to &post=, but `showPeerByLinkResolved` upgrades resolveType from Default to BotApp whenever botAppName is non-empty — and botAppName falls back to the &post= string in local_url_handlers.cpp. So the client looked up an app named "142521" and showed the not-found toast.

Two extra changes in `window_session_controller.cpp`:

  - Skip the Default → BotApp upgrade when peer->isForum() — for forum bots a numeric segment is a topic id, not an app short name.
  - Treat info.messageId as authoritative for forum bots in useRequestedMessageId, so the topic opens via showMessageByLinkResolved instead of falling back to the forum view.

Coverage:

  - t.me/bot/myapp (named app) — opens app
  - t.me/bot?startapp=... (default mini app) — opens app
  - t.me/bot?startgroup=... / ?startchannel=... — opens add-bot popup
  - t.me/forumbot/142521 — opens topic
  - t.me/bot/12345 for non-forum bots — unchanged

john-preston:
@rdmcd I'm skeptical about t.me/bot/<numeric> for topic links, what does that even mean, who can create such links (the ids are local for current user and I'm not sure even bot knows what is the topic id for the user, maybe he has his own ids), this is completely different case than topics in supergroups.

john-preston:
Do official mobile apps support those t.me/botname/topic_id links? This is not something I want to add if not used and standardized already for bots with topics.

rdmcd:
You're right — I tested on iOS and Android and `t.me/bot/<numeric>` opens the bot chat (forum view) without a toast, doesn't open a specific topic. So the "open topic via showMessageByLinkResolved" change was overreaching.

Reverted that part. Kept only the `&& !peer->isForum()` guard in the `Default` → `BotApp` upgrade so the desktop client matches iOS/Android: falls back to forum view instead of resolving a non-existent app named after the post id and showing "Bot application not found". This is just preventing the regression introduced by #30446 — not adding the topic-deeplink feature.

john-preston:
@rdmcd Why would you add the change for those links then? This:

// For bots with Threaded Mode the numeric segment is not an app
// short name, so don't upgrade to BotApp in that case.

Or did I misunderstood you?

ilya-fedin:
No answer to review in half a year

PR #30552, opened 2026-04-11
CMakeLists: enable MSVC security flags

chipitsine:
Add /guard:cf and /sdl to compile options, and /guard:cf, /CETCOMPAT, /DYNAMICBASE to link options for Telegram target when building with MSVC. These flags improve control flow guarding, dynamic base relocation, and overall security hardening of the Windows build.


ilya-fedin:
Such options should be in https://github.com/desktop-app/cmake_helpers/blob/master/options_win.cmake

chipitsine:
if you prefer, I'll move them there

PR #30555, opened 2026-04-11
Update compress_certificate extension to match Chrome 119+ (add zlib support)

chipitsine:
Improve DPI resistance by mimic chrome behaviour.

FakeTLS ClientHello was fingerprint-detectable because compress_certificate (0x001b) only advertised brotli, while Chrome 119+ advertises both zlib and brotli (in that order).


john-preston:
> Chrome 119+ advertises both zlib and brotli

No, it doesn't.

PR #30393, opened 2026-03-04
[Regression] Bring back original video quality option, handle attributes correctly

comdev1337:
Fixes #29780
Fixes #30407 
Fixes: a bug (?) where the video quality text (SD/HD/4K) is not displayed on the gear button if a video has no `alt_documents` (only speed options are available in the dropdown).

### Summary

This PR aims to help **the minor subset of users with a functioning eyesight** by restoring the original video quality option in the media player. The original stream is included as the second option (consistent with the ordering in the Android client).

[Example video](https://t.me/vkpelmen/3636). Observe: wheels turn into slop when 1080p is selected.

https://github.com/user-attachments/assets/1d32b841-17a8-4f24-9eaa-9a9ba94d0d80

[Example video](https://t.me/cursedvideofiles/9). Handling wrong attributes correctly. We now have a 1080p stream!

https://github.com/user-attachments/assets/25d91513-ed8e-4078-ab73-b404ac26bf82

  ### Implementation
   - Upon resolving video qualities, the original quality is added to the array with a 1000000 integer offset (`Media::kVideoQualityOriginalOffset`). This eliminates resolution ambiguity between an original 1080p stream and a sloppy 1080p.
   - The settings deserializer has been updated to accept values up to 1004320 to ensure the "Original" quality preference persists across restarts.
   - Tie-Breaker Fix: Fixed an edge case in `chooseQuality()` where selecting the transcoded 1080p option selects the original 1080p, if the transcoded stream size is over the original file size, due to
```cpp
if (abs == closestAbs && quality->size < closestSize)
```
   - Get video dimensions from ffmpeg directly (`realsize`), do not rely on attributes whenever possible.
     
  

---
Tried throwing a bunch of different videos at it. Works on my machine.

Issue #30407, opened 2026-03-06
Video player follows incorrect attributes for scaling, duration


comdev1337:
### Steps to reproduce

Open a video with
1. quality options
2. Incorrect w, h or/and duration attributes

Examples:
- https://t.me/Ateobreaking/165411 (1280x720 uploaded as fake 848x464, iOS client)
- https://t.me/cursedvideofiles/6 (same video id as above, just added some useful context)
- https://t.me/cursedvideofiles/8 (video from above reuploaded as fake 100x100, fake duration of 1 sec)
- https://t.me/cursedvideofiles/7 (video from above reuploaded as fake 1488x1488, fake duration of 1488 sec)
- https://t.me/warhistoryalconafter/261089 (1280x720 uploaded as fake 848x464, iOS client)
- https://t.me/strikedronescompany/418 (1920x1080 uploaded as fake 848x464, iOS client)

0:08 Note how the dimensions of the video player change from **1280x720** => **incorrect 848x464**, even thought the selected **480p** stream is supposed to have the resolution of **852x480**

https://github.com/user-attachments/assets/e260abde-e927-475d-91ef-621402097026

### Expected behaviour

The video player loads attributes for the selected stream and doesn't follow them blindly. This results in correct video player dimensions, duration being displayed.

### Actual behaviour

The video player blindly relies on wrong **original file's attributes** instead of the attributes of the **steam that's currently selected** in the dropdown menu. This results in wrong video player dimensions, wrong duration being shown.

For dimensions `_document->dimensions` in
https://github.com/telegramdesktop/tdesktop/blob/5f1121123c6cccbb9ed9efa456c211f1c32d7299/Telegram/SourceFiles/media/view/media_view_overlay_widget.cpp#L1180-L1187
For duration `_document->duration()` in
https://github.com/telegramdesktop/tdesktop/blob/5f1121123c6cccbb9ed9efa456c211f1c32d7299/Telegram/SourceFiles/media/view/media_view_overlay_widget.cpp#L4660-L4671

### Operating system

Windows 11

### Version of Telegram Desktop

6.6.2

### Installation source

Static binary from official website

### Logs

```shell

```

john-preston:
I strongly feel that we should demand correct attributes, not work around incorrect attributes. Why would you send incorrect attributes anyway.

comdev1337:
@john-preston 
> I strongly feel that we should demand correct attributes

[...]

PR #30622, opened 2026-04-28
fix: use original stream's duration attribute for transcodes

### LINKED ISSUE #30616: 480p streams playback broken
### Steps to reproduce

1. Select 480p in the video quality.
2. Press play.
3. Watch.

### Expected behaviour

The video should play fully.

### Actual behaviour

Introduced in 6.7.7, still reproducible in 6.7.8.
The original videos are 720p. Their 480 streams are fine. The playback of 480p streams is broken:
- The UI shows an incorrect duration (e.g. 10 seconds for a 16-second or 18-second video, 20 seconds for a 26-second video).
- The video playback freezes at that incorrect duration while the audio keeps playing normally until the end.

### Operating system

Windows 10 64-bit

Telegram Desktop portable 64-bit 6.7.8, 6.7.7

### Installation source

### Crash ID

_No response_

### Logs

```shell

```

comdev1337:
Fixes #30616 by reverting the broken duration document selection behavior as it was before #30393 - transcodes rely on the original's duration attribute instead of their own

Core issue: the backend provides incorrect duration for slop streams both in attributes and in moov, so there's no fallback possible (unless you do something like packet-based duration, which would require actual effort)

This results in the correct duration in https://t.me/cursedvideofiles/11 but incorrect duration for slop steams in https://t.me/cursedvideofiles/8

### DIFF
--- a/Telegram/SourceFiles/media/view/media_view_overlay_widget.cpp
+++ b/Telegram/SourceFiles/media/view/media_view_overlay_widget.cpp
@@ -5136,16 +5136,12 @@ void OverlayWidget::restartAtSeekPosition(crl::time position) {
 	}
 	const auto overrideDuration = _stories
 		|| (_chosenQuality && _chosenQuality != _document);
-	const auto durationDocument = (_chosenQuality && _chosenQuality != _document)
-		? _chosenQuality
-		: _document;
-
 	auto options = Streaming::PlaybackOptions{
 		.position = position,
 		.durationOverride = ((overrideDuration
-			&& durationDocument
-			&& durationDocument->hasDuration())
-			? durationDocument->duration()
+			&& _document
+			&& _document->hasDuration())
+			? _document->duration()
 			: crl::time(0)),
 		.hwAllowed = Core::App().settings().hardwareAcceleratedVideo(),
 		.seekable = !_stories,


john-preston:
Thanks!

zebop117:
Make this behavior controlled by a user preference so people can flip it either way and still watch a video with broken metadata. Don't hide this preference in hidden/debug preferences.
Since there's no universal solution for the root cause (incorrect metadata) on the client side, playback will always be broken for someone (unless a miracle happens and people start thinking and testing before vomiting out their shit in production). I mean, the client could fetch and decode the entire stream in order to get the actual duration, but this isn't a reasonable solution and it defeats the purpose of having **streams** that don't require downloading the entire file in order to start watching.
Switching to HLS playlists isn't a solution either because nothing can stop the backend from fucking up fragment duration in metadata. In fact, it just gives them an additional way to fuck things up.

john-preston:
Why a transcoded version could have different real duration than the original? I missed that in the original introduction of this change..

comdev1337:
@zebop117 

>Make this behavior controlled by a user preference

They should just make videos streamable with an external player, any decent player can do that.

>Since there's no universal solution for the root cause (incorrect metadata) on the client side

I mentioned a better way to do this above -- "unless you do something like packet-based duration which would require actual effort". I [tried to do that](https://github.com/telegramdesktop/tdesktop/commit/41275681ed6d08b80d746ce9d426c84cc1b93d13) but I couldn't get it to 100% work properly, and I couldn't be bothered to waste any more tokens on it.

>I mean, the client could fetch and decode the entire stream in order to get the actual duration

Theoretically it should be enough to look at pts/dts, you don't need to decode the entire stream

comdev1337:
@john-preston 
>Why a transcoded version could have different real duration than the original?

Because the original's duration attribute is **unchecked user input**, as I explained [here](https://github.com/telegramdesktop/tdesktop/issues/30616#issuecomment-4334335009) and #30407.

Even your own backend [doesn't provide](https://github.com/telegramdesktop/tdesktop/issues/30616#issuecomment-4326538737) correct duration metadata for streams it produces (something I wasn't aware about, but it explains https://github.com/telegramdesktop/tdesktop/commit/fabdd89c4ae386c4af633b1e65fb730bbc900bcf, I suppose)

Not to mention transcoding can change the duration in scenarios involving VFR => CFR, timecode errors, corrupted frames, etc.

john-preston:
@comdev1337 well, if someone deliberately sends a video with incorrect duration in the attributes, and someone can't play their video.. well, they achieved their goal? they could send some broken unplayable file from the start..

comdev1337:
@john-preston 
>if someone deliberately sends a video with incorrect duration

there's a more likely hypothetical scenario: a bug in a client (similar to [this](https://github.com/TelegramMessenger/Telegram-iOS/issues/2038))

Remu-rin:
> Even your own backend [doesn't provide](https://github.com/telegramdesktop/tdesktop/issues/30616#issuecomment-4326538737) correct duration metadata for streams it produces (something I wasn't aware about, but it explains [fabdd89](https://github.com/telegramdesktop/tdesktop/commit/fabdd89c4ae386c4af633b1e65fb730bbc900bcf), I suppose)

Is this another example? `https://t.me/vikonzone/6556`
Original, 720p and 480p are fine.
But 1080p (non-original) seems to have wrong duration (both tdesktop 6.7.8 and Android 12.6.4, but Android plays after 21s and lags at starting the loop).

comdev1337:
Yes, and it has fake dimensions
```
--- ATTRIBUTES (Original) ---
  Resolution: 464x848 (848p)
  Duration:   00:26.05 (26.05s)
  Codec:      none/original
  Mime:       video/mp4
  File Size:  20.84 MB

--- ALT DOCUMENTS (14 entries) ---

--- Alt #1 ---
  Resolution: 1072x1920 (1920p)
  Duration:   00:22.07 (22.07s)
  Codec:      av01
  Mime:       video/mp4
  File Size:  9.82 MB

--- Alt #3 ---
  Resolution: 1072x1920 (1920p)
  Duration:   00:21.12 (21.12s)
  Codec:      h264
  Mime:       video/mp4
  File Size:  13.14 MB

--- Alt #5 ---
  Resolution: 712x1280 (1280p)
  Duration:   00:25.95 (25.95s)
  Codec:      av01
  Mime:       video/mp4
  File Size:  4.44 MB

--- Alt #7 ---
  Resolution: 712x1280 (1280p)
  Duration:   00:25.95 (25.95s)
  Codec:      h264
  Mime:       video/mp4
  File Size:  5.68 MB

--- Alt #9 ---
  Resolution: 476x856 (856p)
  Duration:   00:25.95 (25.95s)
  Codec:      av01
  Mime:       video/mp4
  File Size:  2.09 MB

--- Alt #11 ---
  Resolution: 476x856 (856p)
  Duration:   00:25.95 (25.95s)
  Codec:      h264
  Mime:       video/mp4
  File Size:  2.66 MB
```

PR #30669, opened 2026-05-08
Allow timestamp links inside text formatting entities

evtn:
Currently, timecode links work only when not in entities, except, for some reason, spoilers

This means you can't, for example, place a description with timecodes in an expandable blockquote or format timecodes with italics. 

This PR addresses this. I also didn't include links and code blocks to allowed entities because that would be confusing and probably those are the reason for that condition in the first place.


ilya-fedin:
Conflicts

PR #30672, opened 2026-05-09
Storage: show specific error for insufficient disk space

### LINKED ISSUE #30668: Telegram Desktop does not show “Not enough disk space” error during download failure
### Steps to reproduce

1. Fill the PC storage almost completely.
2. Open Telegram Desktop.
3. Start downloading a large file.
4. Wait for the download to fail.

### Expected behaviour

Telegram Desktop should show a clear error message such as:
“Not enough storage space on your PC.”

### Actual behaviour

Telegram Desktop only shows “Failed” and “Try Again” without explaining that the issue is caused by insufficient disk space.

### Operating system

Windows 11

6.7.6 x64

### Installation source

### Crash ID

N/A (No crash occurred)

### Logs

```shell
No logs available / N/A
```

vaezim:
**Summary**

Previously, all file write failures sent a `FailureReason::FileWriteFailure`. This PR adds a specific `FailureReason::NoDiskSpaceFailure` failure reason for out-of-disk-space scenarios and displays a dedicated error message box to the user.

**Related Issues:**

Fix [#30668](https://github.com/telegramdesktop/tdesktop/issues/30668)

Before:
download_error

After:
no_disk_error

### DIFF
--- a/Telegram/SourceFiles/data/data_document.cpp
+++ b/Telegram/SourceFiles/data/data_document.cpp
@@ -1286,10 +1286,17 @@ void DocumentData::handleLoaderUpdates() {
 				Ui::hideLayer();
 				save(origin, failedFileName);
 			};
-			Ui::show(Ui::MakeConfirmBox({
-				tr::lng_download_finish_failed(),
-				crl::guard(&session(), retry)
-			}));
+			if (error.failureReason == FailureReason::NoDiskSpaceFailure) {
+				Ui::show(Ui::MakeConfirmBox({
+					tr::lng_download_no_disk_space_failed(),
+					crl::guard(&session(), retry)
+				}));
+			} else {
+				Ui::show(Ui::MakeConfirmBox({
+					tr::lng_download_finish_failed(),
+					crl::guard(&session(), retry)
+				}));
+			}
 		} else if (error.failureReason == FailureReason::FileWriteFailure) {
 			if (!Core::App().settings().downloadPath().isEmpty()) {
 				Core::App().settings().setDownloadPathBookmark(QByteArray());
--- a/Telegram/SourceFiles/storage/file_download.cpp
+++ b/Telegram/SourceFiles/storage/file_download.cpp
@@ -376,8 +376,23 @@ bool FileLoader::writeResultPart(int64 offset, bytes::const_span buffer) {
 			_skippedBytes += offset - fsize;
 		}
 		_file.seek(offset);
-		if (_file.write(reinterpret_cast<const char*>(buffer.data()), buffer.size()) != qint64(buffer.size())) {
-			cancel(FailureReason::FileWriteFailure);
+		auto result = _file.write(reinterpret_cast<const char*>(buffer.data()), buffer.size());
+		auto writtenBytes = result;
+		while (writtenBytes != qint64(buffer.size()) && result > 0) {
+			if (_file.error() == QFile::ResourceError) {
+				cancel(FailureReason::NoDiskSpaceFailure);
+				return false;
+			} else if (_file.error() != QFile::NoError) {
+				cancel(FailureReason::FileWriteFailure);
+				return false;
+			}
+			result = _file.write(reinterpret_cast<const char*>(buffer.data()) + writtenBytes, buffer.size() - writtenBytes);
+			writtenBytes += result;
+		}
+		if (writtenBytes != qint64(buffer.size())) {
+			cancel(_file.error() == QFile::ResourceError
+				? FailureReason::NoDiskSpaceFailure
+				: FailureReason::FileWriteFailure);
 			return false;
 		}
 		return true;
--- a/Telegram/SourceFiles/storage/file_download.h
+++ b/Telegram/SourceFiles/storage/file_download.h
@@ -55,6 +55,7 @@ public:
 	enum class FailureReason {
 		NoFailure,
 		FileWriteFailure,
+		NoDiskSpaceFailure,
 		OtherFailure,
 	};
 
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -1125,6 +1125,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 "lng_download_path_choose" = "Choose download path";
 "lng_download_path_failed" = "File download could not be started.\n\nThis might be because your selected download location is invalid. Try changing it in Settings > Advanced > Download Path.";
 "lng_download_finish_failed" = "File download could not be completed.\n\nWould you like to try again?";
+"lng_download_no_disk_space_failed" = "Not enough disk space to download this file.\n\nWould you like to try again?";
 
 "lng_settings_section_privacy" = "Privacy and Security";
 


Telegram/SourceFiles/storage/file_download.cpp
```diff
@@ -377,6 +377,10 @@ bool FileLoader::writeResultPart(int64 offset, bytes::const_span buffer) {
 		}
 		_file.seek(offset);
 		if (_file.write(reinterpret_cast<const char*>(buffer.data()), buffer.size()) != qint64(buffer.size())) {
+			if (errno == ENOSPC || _file.error() == QFileDevice::ResourceError) {
+				cancel(FailureReason::NoDiskSpaceFailure);
+				return false;
+			}
 			cancel(FailureReason::FileWriteFailure);
```
ilya-fedin:
```suggestion
			if (errno == ENOSPC || _file.error() == QFileDevice::ResourceError) {
				cancel(FailureReason::NoDiskSpaceFailure);
				return false;
			}
cancel(_file.error() == QFileDevice::ResourceError
	? FailureReason::NoDiskSpaceFailure
	: FailureReason::FileWriteFailure);
```

vaezim: The problem is that when there is no disk space, `_file.error()` returns 0 (no error). That's why I'm checking `errno` against `ENOSPC`.

Also, `QFileError::ResourceError` does not always mean insufficient disk space.

That's why I think in line 384, the general `FailureReason::FileWriteFailure` would be better for now. Although in the future we can handle each failure reason separately.

ilya-fedin: > The problem is that when there is no disk space, _file.error() returns 0 (no error). 

Why do you check for QFileError::ResourceError then?

> That's why I'm checking errno against ENOSPC.

I'm not sure that is a good thing to do, as far as I understand there's no guarantee Qt doesn't do some other system call after the main operation making errno returning error for the wrong system call.

ilya-fedin: I'm also not sure whether this compiles on Windows at all?

ilya-fedin: I assume the only 100% safe and cross-platform way is to use the message straight from https://doc.qt.io/archives/qt-5.15/qiodevice.html#errorString?

vaezim: You're right Qt may be calling other system calls after it and errno is not recognized in windows.

I checked `QIODevice::errorString()`. It returns `Unknown error`.

vaezim: I was also inclined towards getting the error from Qt but the only option `_file.error()` isn't returning any error. I guess the only options are directly from the system itself. `errno` on Mac and linux and `GetLastError()` in windows.

ilya-fedin: This most likely means you're just checking error on the wrong QFile object

vaezim: In line 379 size of written bytes is checked from the output of `_file.write()` against `buffer.size()`. So the error is coming from `_file` no?

ilya-fedin: You have no error from both error() and errorString() which means no

vaezim: Apparantly Qt does not gurantee to determine the real reason of write failure

ilya-fedin: I guess they mean that error() is a custom enum? errorString is filled directly from strerror(errno) (and the Windows equivalent)

Telegram/SourceFiles/storage/file_download.cpp
```diff
@@ -377,6 +377,18 @@ bool FileLoader::writeResultPart(int64 offset, bytes::const_span buffer) {
 		}
 		_file.seek(offset);
 		if (_file.write(reinterpret_cast<const char*>(buffer.data()), buffer.size()) != qint64(buffer.size())) {
+			bool diskFullError = (_file.error() == QFileDevice::ResourceError);
+			if (!diskFullError) {
+#ifdef Q_OS_WIN
+				diskFullError = (GetLastError() == ERROR_DISK_FULL || GetLastError() == ERROR_HANDLE_DISK_FULL);
+#elif defined Q_OS_UNIX
+				diskFullError = (errno == ENOSPC);
+#endif // Q_OS_WIN || Q_OS_UNIX
```
ilya-fedin: Sorry, but I still believe we shouldn't use platform code here. Looking at Qt code, either ResourceError or WriteError should work and errorString should always provide the system error string.

https://github.com/qt/qtbase/blob/44551815ca3ec8cae06f5f394c5e259e936eca7a/src/corelib/io/qfsfileengine.cpp#L778-L779

https://github.com/qt/qtbase/blob/44551815ca3ec8cae06f5f394c5e259e936eca7a/src/corelib/io/qfsfileengine_win.cpp#L352

vaezim: If `writtenBytes` is nonzero, `ENOSPC` error is not set

ilya-fedin: I assume you need to repeat the write (with written bytes offset in the buffer of course) until either there's a error or we have written everything

vaezim:
You were right. One more round of `_file.write()` was needed until `_file.error()` returned `QFile::ResourceError`

Telegram/SourceFiles/storage/file_download.cpp
```diff
@@ -376,7 +376,24 @@ bool FileLoader::writeResultPart(int64 offset, bytes::const_span buffer) {
...
+			}
+			result = _file.write(reinterpret_cast<const char*>(buffer.data()) + writtenBytes, buffer.size() - writtenBytes);
+			writtenBytes += result;
+		}
+		if (writtenBytes != qint64(buffer.size())) {
+			if (_file.error() == QFile::ResourceError) {
+				cancel(FailureReason::NoDiskSpaceFailure);
+				return false;
+			}
 			cancel(FailureReason::FileWriteFailure);
```
ilya-fedin:
```suggestion
			cancel(_file.error() == QFile::ResourceError
				? FailureReason::NoDiskSpaceFailure
				: FailureReason::FileWriteFailure);
```

ilya-fedin:
What's the system you tested this on? The Qt code looks like ResourceError will work on Windows only if the file is open via QFile::open overload that accepts FILE* (which makes it use the same code path as used on Unix-like systems). So either file opening has to be reworked to call `_file.open(fopen(_filename.toUtf8(), ...), ..., QFileDevice::AutoCloseHandle)` or the error string has to be passed through.

ilya-fedin:
FWIW, I discussed this with @john-preston and he said `The most common problem is that you ran out of disk space, please check` could just be added to the existing error message instead

vaezim:
> FWIW, I discussed this with @john-preston and he said `The most common problem is that you ran out of disk space, please check` could just be added to the existing error message instead

You mean to remove all `*.cpp` files changes and only edit the error message in `Telegram/Resources/langs/lang.strings` ?

Telegram/SourceFiles/data/data_document.cpp
```diff
@@ -1286,10 +1286,17 @@ void DocumentData::handleLoaderUpdates() {
...
+				Ui::show(Ui::MakeConfirmBox({
+					tr::lng_download_no_disk_space_failed(),
+					crl::guard(&session(), retry)
+				}));
+			} else {
+				Ui::show(Ui::MakeConfirmBox({
+					tr::lng_download_finish_failed(),
+					crl::guard(&session(), retry)
+				}));
+			}
```
ilya-fedin, line 1299: Oh, just noticed
```suggestion
			Ui::show(Ui::MakeConfirmBox({
				error.failureReason == FailureReason::NoDiskSpaceFailure
					? tr::lng_download_no_disk_space_failed()
					: tr::lng_download_finish_failed(),
				crl::guard(&session(), retry)
			}));
```

ilya-fedin:
> You mean to remove all `*.cpp` files changes and only edit the error message in `Telegram/Resources/langs/lang.strings` ?

Not sure... AFAIK it has to be actually changed on translation platform. I asked John whether you can do that or whether he will do it.

Telegram/SourceFiles/storage/file_download.h
```diff
@@ -55,6 +55,7 @@ class FileLoader : public base::has_weak_ptr {
 	enum class FailureReason {
 		NoFailure,
 		FileWriteFailure,
+		NoDiskSpaceFailure,
```
ilya-fedin, line 58: I believe either No or Failure should be removed (and from the lang string name too)

john-preston:
Yeah, the easiest way is to add "\n\nThe most common problem is that you ran out of disk space, please check." to the end of the error message on the error problem and not change anything at all in this repository. Sounds fine?

Otherwise it should be a complete implementation, tested on all three platforms in both cases of that reason or other reason, in case disk was full on download start and became full in the middle of the download, not just some change that may work or may not in some cases.

vaezim:
Yeah on windows if write is done with WinAPI, it gets tricky because all errors are mapped to `QFile::WriteError` on Qt and we have to rely on error string which is not reliable.

ilya-fedin:
> Yeah on windows if write is done with WinAPI, it gets tricky because all errors are mapped to `QFile::WriteError` on Qt

Well, you can use the QFile::open overload accepting FILE* to force it through libc which is the same code path as on nixes

> we have to rely on error string which is not reliable.

Is it really that bad?

vaezim:
> Well, you can use the QFile::open overload accepting FILE* to force it through libc which is the same code path as on nixes

Force it to use stdlib instead of winApi? not sure, I read winApi write better handles large files.

> Is it really that bad?

Parsing certain keywords out from the error string is a bit too complex for this matter. Also error messages might get changed in the future.

ilya-fedin:
> Parsing certain keywords out from the error string is a bit too complex for this matter. Also error messages might get changed in the future.

I mean present it to the user, not parse it

vaezim:
isn't translate file a static file? can we insert something into it at runtime? Also posix error messages might not be easily understandable for average user.

ilya-fedin:
Yeah, you can add like `{error}` to the string and then use `lt_error` as an argument to the string function, an example is lng_channels_leave

vaezim:
Only to get the error, I have to keep the while loop in file_download.cpp line 381 since no error is reported the first time. John said not to touch any source files.

john-preston:
@vaezim My proposed solution is to simply add a phrase for all write failures to the error text on translations platform. I doubt runtimr detection is really needed.

vaezim:
As in our discussion with @ilya-fedin Qt only sets error when `write()` function has written 0 bytes. That's why the first time when it has written some bytes and failed (due to disl space), `_file.errorString()` is `Unknown error`. The next time we call `write()` it can't write any more bytes and returns 0 and now Qt sets NoResource error and error string.

so minor change in file_download.cpp was necessary.

Just changing translate file error message would show `Unknown Error` to the user

https://github.com/telegramdesktop/tdesktop/pull/30672#discussion_r3215550855

ilya-fedin:
> Just changing translate file error message would show `Unknown Error` to the user

Maybe you didn't get it. What @john-preston meant is the error message to be hardcoded like:
```
File download could not be completed.

The most common problem is that you ran out of disk space, please check.

Would you like to try again?
```
Independent of real failure reason.

ilya-fedin:
> Only to get the error, I have to keep the while loop in file_download.cpp line 381 since no error is reported the first time. John said not to touch any source files.

In case you're ok with John's solution, just say that and close the PR, no code changes needed - John will just update the string on the translation platform. In case you want to know that it was no disk space for sure then you can change the pre-existing string to have `{error}` (`lt_error` won't appear without that, as it's a codegen) and pass it from your current code with while loop. Or, well, you can go the fopen route... I'm not sure why say WinAPI better handles large files, I couldn't find anything about that with fast googling at least. From what I found, the only difference seem to be that fwrite has its own buffer and actual write happens either when it fills or when fflush is called but there should be no real difference for us I assume since we write everything at once and IIUC the file will be immediately closed anyway (which means flush).

vaezim:
Sure, let's go with John's solution. I'll close this PR.

PR #30674, opened 2026-05-09
Update tgcalls, gcc build fix.

joecool1029:
Just adds ```<cstdint>``` so GCC 16 builds stop failing.


john-preston:
Yeah, I've updated the submodule in latest `dev`, I guess it's the same.

PR #30683, opened 2026-05-11
Fix embedding/playback of portrait 4K videos, bump max video resolution to 8K

### LINKED ISSUE #30678: Weird behavior of 4K video uploads
### Steps to reproduce

1. click on any 4K (2160x3840, vert or hor) video/GIF
=> media player stuck forever at blurred 1st frame
=> a few can play normally, but other video/GIF with same dimension/bitrate doesn't always plays, can't find a pattern that what special kinds are bugging

2. try to upload a new/previously uploaded 4K video
=> always regarded as a FILE attachment

TG web is alright, only TG desktop start bugging TODAY

### Expected behaviour

1. plays/streams normally regardless of faststart/web-optimized
2. for 4K mp4 file, should always seen as VIDEO attachment

### Actual behaviour

(refer to step to reproduce)

### Operating system

win 10

Version 6.8.1 x64

### Installation source

### Crash ID

_No response_

### Logs

```shell

```

comdev1337:
Fixes #30678
See: https://github.com/telegramdesktop/tdesktop/issues/30678#issuecomment-4425704934

- FFmpeg may validate decoder-allocation dimensions, not just visible frame dimensions. Some H.264/HEVC frames are internally padded for alignment, so a visible 2160x3840 frame can trip max_pixels even though 2160 * 3840 == 3840 * 2160. The patch keeps Telegram’s real visible-frame validation intact, but gives FFmpeg max_pixels a small alignment slack.
- The upload/playback frame-area cap is raised from 4K to 8K, 7680x4320 videos can be embedded
- 8K videos must not allocate full-size ARGB thumbnails/cover frames during inspection. A full 7680x4320 BGRA image is ~126 MiB and exceeds the existing 64 MiB frame-storage check, so the patch renders small thumbnail/cover frames while preserving the real video dimensions separately.
- Video document attributes now use the decoded video dimensions, not the generated thumbnail size, so scaled-down thumbnails do not accidentally turn an 8K video into 320x180 metadata.
- Playback startup now reads dimensions from the decoded frame metadata first and waits for the UI’s requested render size before allocating image storage, avoiding full-resolution first-frame allocation.


john-preston:
Thanks for fixing the regression. But other changes I don't think are desired right now, they can be discussed, but separate from regression fix for the files that played before but stopped playing in the latest version.

I don't want to raise sending / playable files dimensions to 8K, they were never sent / played before and this change has nothing to do with recent regression. Also, rendering the initial frame in streaming downscaled to isn't what I want, because I cache that frame separately as a "full first frame", and 320x320 won't be good for most of the videos.

john-preston:
For now I've merged the fixing regression part.

comdev1337:
Fair enough, I should have split these. Removed the 665a2887e6641e10cb76a21ccaf0d9a99b5256f4 part.

>Also, rendering the initial frame in streaming downscaled to isn't what I want, because I cache that frame separately as a "full first frame", and 320x320 won't be good for most of the videos.

Bumped `kMaxFrameStorageBytes` to 128 MiB instead

kalandrack69:
Day

ilya-fedin:
Conflicts

PR #30689, opened 2026-05-12
Fix swipe reply target for grouped media

Katze-942:
Swipe reply now targets the exact grouped media item under the cursor/finger instead of always replying to the first item in the group.


23rd:
Thanks, it will be merged but in manual way.

23rd:
d8dc10876ca19ec3df3bfa7543927fada95e6c14

PR #30695, opened 2026-05-14
Support Shift+click to extend message text selection

dmitry:
When a message has selected text, `Shift+click` in the same message now extends the selection to the click point, just like in browsers. Previously, `Shift+click` simply dropped the old selection and started a new one.

Works in regular chats, channels, topics, replies, pinned, and scheduled messages.

How to try: select a few words in a message, then `Shift+click` somewhere else in the same message - the selection should extend to the click point.

### DIFF
--- a/Telegram/SourceFiles/history/history_inner_widget.cpp
+++ b/Telegram/SourceFiles/history/history_inner_widget.cpp
@@ -1916,10 +1916,13 @@ void HistoryInner::mousePressEvent(QMouseEvent *e) {
 	}
 	_mouseActive = true;
 	registerReadMetricsActivity();
-	mouseActionStart(e->globalPos(), e->button());
+	mouseActionStart(e->globalPos(), e->button(), e->modifiers());
 }
 
-void HistoryInner::mouseActionStart(const QPoint &screenPos, Qt::MouseButton button) {
+void HistoryInner::mouseActionStart(
+		const QPoint &screenPos,
+		Qt::MouseButton button,
+		Qt::KeyboardModifiers modifiers) {
 	mouseActionUpdate(screenPos);
 	if (button != Qt::LeftButton) return;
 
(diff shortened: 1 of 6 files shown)


Telegram/SourceFiles/history/history_inner_widget.cpp
```diff
@@ -1998,7 +2001,25 @@ void HistoryInner::mouseActionStart(const QPoint &screenPos, Qt::MouseButton but
...
-				if (uponSelected) {
+				const auto canShiftExtend = (modifiers & Qt::ShiftModifier)
+					&& (_selected.size() == 1)
+					&& (_selected.cbegin()->first == _mouseActionItem)
+					&& (_selected.cbegin()->second != FullSelection)
+					&& (_selected.cbegin()->second.from
+						!= _selected.cbegin()->second.to);
+				if (canShiftExtend) {
+					auto symbol = _mouseTextSymbol;
+					if (dragState.afterSymbol) ++symbol;
```
23rd: Can be replaced with `?:`?

Telegram/SourceFiles/history/history_inner_widget.cpp
```diff
@@ -1998,7 +2001,25 @@ void HistoryInner::mouseActionStart(const QPoint &screenPos, Qt::MouseButton but
...
+					const auto extended = HistoryView::ExtendTextSelectionTo(
+						_selected.cbegin()->second,
+						symbol);
+					_mouseTextSymbol = (extended.from == symbol)
+						? extended.to
+						: extended.from;
+					_selected.begin()->second = extended;
+					_mouseAction = MouseAction::Selecting;
+					repaintItem(_mouseActionItem);
```
23rd, line 2022: Do we need `repaintItem`?

dmitry: We mutate `_selected.begin()->second` directly without a setter, so nothing auto-repaints.

Telegram/SourceFiles/history/history_inner_widget.cpp
dmitry: Applied in both `HistoryInner` and `ListWidget`.

john-preston:
This is nice! Sadly, we don't track really where the anchor is, so the feel still won't be native (did we select from symbol 20 to symbol 10 or from symbol 10 to symbol 20 and then shift click on symbol 15 - the results should be different, like in browsers, but here we don't save that information). But this a start!

ilya-fedin:
Conflicts

PR #30698, opened 2026-05-15
fix: preserve PNG transparency in chat wallpapers

### LINKED ISSUE #29773: White background in chat, when it's supposed to be transparent
### Steps to reproduce

1. Apply this theme

[tdesktop-palette.txt](https://github.com/user-attachments/files/22295004/tdesktop-palette.txt)

2. Use this image as a bg in chats (it's a transparent PNG)

Image

3. See that it's not transparent

Image

### Expected behaviour

It should be transparent, like here
https://github.com/weskerty/TelegramGlassTheme

### Actual behaviour

It's white background

### Operating system

NixOS unstable, Hyprland 0.50.0

From nixpkgs: Telegram 6.1.1 and AyuGram 5.16.3, couldn't run your "static" binary, weird that "static" binary requires a ton of other libraries to be installed on system

### Installation source

Other (unofficial) source

### Crash ID

_No response_

### Logs

```shell

```

pratikpatel8982:
Respect the user's chosen image as-is without forcing opacity, which naturally enables transparent chat background support.

fixes #29773

### DIFF
--- a/Telegram/SourceFiles/boxes/background_box.cpp
+++ b/Telegram/SourceFiles/boxes/background_box.cpp
@@ -250,7 +250,7 @@ void BackgroundBox::chooseFromFile() {
 		auto image = Images::Read({
 			.path = result.paths.isEmpty() ? QString() : result.paths.front(),
 			.content = result.remoteContent,
-			.forceOpaque = true,
+			.forceOpaque = false,
 		}).image;
 		if (image.isNull() || image.width() <= 0 || image.height() <= 0) {
 			return;
--- a/Telegram/SourceFiles/settings/sections/settings_chat.cpp
+++ b/Telegram/SourceFiles/settings/sections/settings_chat.cpp
@@ -707,7 +707,7 @@ void ChooseFromFile(
 		auto image = Images::Read({
 			.path = result.paths.isEmpty() ? QString() : result.paths.front(),
 			.content = result.remoteContent,
-			.forceOpaque = true,
+			.forceOpaque = false,
 		}).image;
 		if (image.isNull() || image.width() <= 0 || image.height() <= 0) {
 			return;


ilya-fedin:
Did you actually test it? The server accepts only jpegs so it will lose the transparency anyway, all you likely did is that the transparency will turn black instead of white.

pratikpatel8982:
> Did you actually test it? The server accepts only jpegs so it will lose the transparency anyway, all you likely did is that the transparency will turn black instead of white.

yes i tested it, the chat backround wallpaper is chosen through the chat settings,
and by setting `forceOpaque=false` the pipeline preserve transparency of the chat background.
image
1. The option `Choose from File` has its own file-chooser method inside `settings_chat.cpp` and this option was tested by me.
2. However `Choose from Gallery` option further expands to gallery and another `Choose from file` option which is taken from `background_box.cpp`. This was added later and is untested. However it should work the same cause they both call the same method and later go through the same pipeline.
image

Here is the image, a fully transparent png which was used for testing.
Transparente

Here is the result: half screenshot of chat window.
image

here is the compiled app (linux)+ the transparent png used: https://gofile.io/d/30Ludp
It also supports semi-transparent images.

pratikpatel8982:
any feedback??

ilya-fedin:
> any feedback??

I couldn't speak for sure (I believed it's not possible due to jpegs after all) but my knowledge of platform differences make me think it's not intended to work. Generally, tdesktop's window is opaque and only on Linux, due to no system APIs allowing to use system shadows with custom titlebar, the window is transparent. And if such API ever appear or even an API to set opaque region (allowing only client-side shadows being actually transparent), making use of it will make the window opaque, just like on other platforms.

pratikpatel8982:
> I couldn't speak for sure (I believed it's not possible due to jpegs after all) but my knowledge of platform differences make me think it's not intended to work. Generally, tdesktop's window is opaque and only on Linux, due to no system APIs allowing to use system shadows with custom titlebar, the window is transparent. And if such API ever appear or even an API to set opaque region (allowing only client-side shadows being actually transparent), making use of it will make the window opaque, just like on other platforms.

That makes total sense, and I completely agree regarding the future of Linux window APIs. If an opaque region API ever gets adopted and tdesktop switches to it, this will naturally break and revert to behaving exactly like Windows/macOS.

However, since this PR just flips `forceOpaque` to false and doesn't add any new rendering logic or UI toggles to maintain, it essentially acts as a zero-cost Easter Egg for current Linux/Wayland users with transparent PNGs. Standard JPEGs still render perfectly as opaque backgrounds.

Also can test on windows, if someone is willing to share compiled application. And the application preserves image transparency at every step, just forces opaque at last probably due to server restrictions.

Happy to leave it up to you if you think it's worth merging for the time being, or if you'd prefer to keep the forced opacity!

ilya-fedin:
> If an opaque region API ever gets adopted and tdesktop switches to it, this will naturally break and revert to behaving exactly like Windows/macOS.

AFAIK opaque region API just tells the compositor that it doesn't have to repaint on transparent regions, so we essentially will get into a situation that people will start complaining about buggy rendering if this is merged.

pratikpatel8982:
That makes perfect sense. If tdesktop adopts an opaque region API in the future, the compositor won't know to repaint the desktop behind the transparent PNG pixels. That would definitely result in un-cleared buffer artifacts and smearing.

Since relying on the current lack of that API creates technical debt and a risk for future bug reports, I completely agree we shouldn't merge this upstream as default behavior. Thanks for taking the time to explain the compositor interaction!

If we still want to try supporting this, I can wrap this `forceOpaque` logic behind an experimental toggle in the settings. That way, it remains strictly opt-in with a warning, and standard users are completely unaffected.

Let me know if you'd be open to an experimental toggle. Otherwise, feel free to close this PR to save yourself the upstream maintenance headaches! I'll just keep it as a local patch for my own setup and propose it to forks like AyuGram instead.

ilya-fedin:
> Let me know if you'd be open to an experimental toggle.

I can't decide that, you have to ask preston

user123456000:
Should we ping him here?

user123456000:
@ilya-fedin

ilya-fedin:
> @ilya-fedin

What?

user123456000:
> > @ilya-fedin
> 
> What?

You said we needed to ask preston. Should we ping him here?

ilya-fedin:
You can try but I'm not sure it will help, he rarely reads github.

user123456000:
@john-preston , could you take a look at this?

john-preston:
The chat backgrounds are synced with the server, they're cloud synced for account between devices and they can't be semitransparent when uploaded to the server, this also should break rendering on macos / windows, where non-opaque background won't fully fill the painted area. I didn't check the changes thoroughly on different systems, but initially this isn't supposed to work. I'm not strictly opposed the idea of an experimental option, but I don't want to add one which is completely untested and supposedly breaks things on other platforms.

assembledev:
There is a narrower case here: PNG backgrounds embedded in `.tdesktop-theme` files.

This path does not enter the custom wallpaper upload flow: `LoadTheme()` stores the image as `Data::ThemeWallPaper()`, while `checkUploadWallPaper()` only uploads `Data::IsCustomWallPaper(_paper)`. Therefore the server-side JPEG conversion discussed above does not apply here.

I tested changing only `forceOpaque` to `false` in `window_theme.cpp` on Telegram Desktop 7.0.2 under Wayland/Hyprland. Semi-transparent theme backgrounds work correctly, while opaque themes remain unchanged.

I don't think this needs an experimental toggle or platform-specific logic: embedding alpha in the PNG and importing that theme is already explicit opt-in. Users can choose such themes on platforms where they work.

Would a separate one-line PR scoped only to `.tdesktop-theme` backgrounds be acceptable?

pratikpatel8982:
@assembledev does it allow user to pick transparent/semi-transparent theme backgrounds or it only accepts png files embedded inside `.tdesktop-theme` ??

assembledev:
Effectively it provides the same visual feature, but only through a `.tdesktop-theme`: the transparent or semi-transparent PNG must be packaged as `background.png` or `tiled.png`, and then the user imports that theme.
For example, my theme contains a 1×1 `tiled.png` whose single pixel is RGBA `#1a1b26db` (about 86% opacity). Telegram tiles it across the entire chat canvas, producing a uniform semi-transparent background on Wayland.
It intentionally does not affect standalone wallpaper PNGs or the upload/sync path.

For reference, this is the entire change I tested:

```diff
diff --git a/Telegram/SourceFiles/window/themes/window_theme.cpp b/Telegram/SourceFiles/window/themes/window_theme.cpp
@@ -333,7 +333,7 @@ bool LoadTheme(
 			auto background = Images::Read({
 				.content = backgroundContent,
-				.forceOpaque = true,
+				.forceOpaque = false,
 			}).image;
```

ilya-fedin:
> I don't think this needs an experimental toggle or platform-specific logic: embedding alpha in the PNG and importing that theme is already explicit opt-in. Users can choose such themes on platforms where they work.

A user shares .tdesktop-theme to a group -> a user on other platform installs it -> sees artifacts -> reports a bug

PR #30701, opened 2026-05-16
Fix possible packet loss 

giorka:
This fixes a possible outbound E2E packet loss issue in `Call::checkForOutboundMessages()`.

Previously, only the last message returned by `call_pull_outbound_messages()` was forwarded:

```cpp
result.value().back()
```

However, the API returns a collection of outbound messages, so multiple pending packets could be dropped if more than one message was queued between polls.

The updated implementation forwards all outbound messages in order instead of only the last one.


23rd:
Indentations are incorrect.

giorka:
Oops, I missed the project's formatting style. Fixed now

PR #30709, opened 2026-05-20
Fixed a warning by removing unused function.

osokin:
/wrkdirs/usr/ports/net-im/telegram-desktop/work/tdesktop-6.8.3-full/Telegram/SourceFiles/iv/markdown/iv_markdown_view.cpp:64:23: warning: unused function 'PrepareFailureReasonText' [-Wunused-function]
   64 | [[nodiscard]] QString PrepareFailureReasonText(
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.

### DIFF
--- a/Telegram/SourceFiles/iv/markdown/iv_markdown_view.cpp
+++ b/Telegram/SourceFiles/iv/markdown/iv_markdown_view.cpp
@@ -44,6 +44,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 namespace Iv::Markdown {
 namespace {
 
+#ifndef NDEBUG
 [[nodiscard]] QString PrepareTerminalFailureName(
 		PrepareTerminalFailure failure) {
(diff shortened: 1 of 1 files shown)


23rd:
Well, if we want to use this again for debugging purposes, will we have to rewrite the function in the future?

23rd:
By the way.
```
iv_markdown_view.cpp:826:22: error: use of undeclared identifier 'PrepareFailureReasonText' const auto reason = PrepareFailureReasonText(failure);
```

john-preston:
@osokin It should be wrapped in #ifndef NDEBUG, like the code, that uses it.

john-preston:
@osokin I suspect the previous function, PrepareTerminalFailureName, should be in the same `#ifndef NDEBUG` block, because now it is not used anymore in Release build.

john-preston:
Thanks!

Issue #30715, opened 2026-05-22
Support Mermaid rendering in Markdown (.md) previews on Desktop


dankarization:
## Feature request

Please support rendering fenced `mermaid` blocks inside Markdown (`.md`) files and Markdown previews in Telegram Desktop.

Example:

````md
```mermaid
flowchart LR
    URL[Markdown file] --> TG[Telegram preview]
    TG --> Diagram[Rendered Mermaid diagram]
```
````

## Why this matters

- Mermaid is widely used for diagrams in technical documentation.
- Telegram is increasingly used to exchange Markdown files between people, teams, and AI tools.
- Without Mermaid rendering, `.md` files with diagrams lose a large part of their value inside Telegram.
- Desktop is an especially strong place for this because users often inspect docs, specs, and architecture notes there.

## Suggested behavior

- Detect fenced code blocks with language `mermaid`
- Render them in Markdown/file preview
- If rendering is unavailable or fails, show the raw code block
- A manual "Render diagram" action would also be a good first version

## Nice-to-have

- Keep behavior aligned with Telegram Android/iOS when possible
- Export/share rendered preview as image if feasible

Thanks.

PR #30716, opened 2026-05-22
Draft: prototype Mermaid embeds for Markdown preview

dankarization:
## Summary

Prototype support for fenced `mermaid` blocks in Markdown preview.

This patch converts code fences with language `mermaid` into a clickable placeholder that opens an embedded preview overlay and attempts to render the diagram there.

Related issue: #30715

## What this does

- detects fenced code blocks with info string `mermaid`
- prepares them as a markdown placeholder instead of a plain code block
- attaches an embedded HTML resource to the placeholder
- renders the diagram inside the existing markdown embed overlay
- reports preferred size back to the overlay so the preview can resize

## Why draft

This is a prototype path, not a polished final implementation.

Known limitations:
- it loads Mermaid from an external CDN instead of a bundled local runtime
- it has not gone through the full Telegram Desktop build/test pipeline
- it does not yet include screenshot/golden tests for rendered output
- fallback behavior is intentionally minimal

## Intent

The point of this PR is to show a concrete integration path using the existing markdown placeholder + embed overlay infrastructure, so the discussion can move from abstract feature request to a real patch shape.

If this direction makes sense, the next obvious steps would be:
- replace CDN loading with a vendored/runtime-approved Mermaid integration
- add proper rendering tests
- refine placeholder text and failure UX

### DIFF
--- a/Telegram/SourceFiles/iv/markdown/iv_markdown_prepare_blocks.cpp
+++ b/Telegram/SourceFiles/iv/markdown/iv_markdown_prepare_blocks.cpp
@@ -55,6 +55,161 @@ constexpr auto kMaxVisualQuoteDepth = 3;
 	return text;
 }
 
+[[nodiscard]] PreparedPlaceholderBlockId GeneratePreparedPlaceholderBlockId(
+		PrepareState *state) {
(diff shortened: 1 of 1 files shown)


ilya-fedin:
No updates in half a year

PR #30723, opened 2026-05-23
Fix sticker filtering in chat export

Katze-942:
I ran into a case where stickers and custom emoji were still included in chat export even when sticker export was disabled.

This change treats custom emoji as stickers for export settings and applies media type filtering consistently before writing exported files.

I'm not sure this is the best place to enforce this, but it fixes the issue in my testing.


ilya-fedin:
Conflicts

PR #30755, opened 2026-06-01
Fix accessibility for PeerList (NVDA Screen Reader)

HassanAlBarshoumy:
This PR fixes a critical accessibility issue where NVDA and other screen readers fail to read the names of contacts, blocked users, and group members in the application.

It implements the missing QAccessible overrides for PeerListContent, allowing it to announce its ListItem children's name and status, and emit focus events when the selection changes via keyboard arrows or mouse.

Tested locally with NVDA and Windows accessibility features.


ilya-fedin:
Conflicts

PR #30757, opened 2026-06-01
Fix rotated video rendering in calls

### LINKED ISSUE #30712: Telegram Desktop Linux renders iPhone video feed with incorrect orientation and tiled/mirrored output
### Steps to reproduce

Make or receive video calls between iphone and Linux Desktop

### Expected behaviour

Normal video call

### Actual behaviour

Image appears as follows:: 
Image

### Operating system

Fedora 44 Plasma 6.6.5

6.8.3 Beta

### Installation source

### Crash ID

_No response_

### Logs

```shell

```

valad47:
In new QRhi renderer, every video call with a non standart video rotation(such as video call from a mobile device)  was not correctly rotated, so it was producing strange effects as mirroring and tilting

This commit fixes #30712

### DIFF
--- a/Telegram/SourceFiles/calls/calls_video_incoming.cpp
+++ b/Telegram/SourceFiles/calls/calls_video_incoming.cpp
@@ -836,8 +836,8 @@ public:
 			_uniformBuffer, 0, sizeof(uniforms), &uniforms);
 
 		std::array<std::array<float, 2>, 4> texCoords = { {
-			{ { 0.f, 0.f } }, { { 1.f, 0.f } },
 			{ { 0.f, 1.f } }, { { 1.f, 1.f } },
+			{ { 1.f, 0.f } }, { { 0.f, 0.f } },
 		} };
 		if (const auto shift = (rotation / 90); shift != 0) {
 			std::rotate(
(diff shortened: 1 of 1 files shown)


Telegram/SourceFiles/calls/calls_video_incoming.cpp
```diff
@@ -836,8 +836,8 @@ class Panel::Incoming::RendererRhi final
 			_uniformBuffer, 0, sizeof(uniforms), &uniforms);
 
 		std::array<std::array<float, 2>, 4> texCoords = { {
-			{ { 0.f, 0.f } }, { { 1.f, 0.f } },
 			{ { 0.f, 1.f } }, { { 1.f, 1.f } },
+			{ { 1.f, 0.f } }, { { 0.f, 0.f } },
 		} };
```
valad47, line 841: In new renderer order of coords was changed to match QRhi vertex order, but because of this rotation of video breaking correct coords, so we have to leave this buffer the same as in old renderer

Telegram/SourceFiles/calls/calls_video_incoming.cpp
```diff
@@ -846,10 +846,10 @@ class Panel::Incoming::RendererRhi final
...
 		const float coords[] = {
-			0.f, ph, texCoords[0][0], texCoords[0][1],
-			pw,  ph, texCoords[1][0], texCoords[1][1],
-			0.f, 0.f, texCoords[2][0], texCoords[2][1],
-			pw,  0.f, texCoords[3][0], texCoords[3][1],
+			0.f, ph, texCoords[3][0], texCoords[3][1],
+			pw,  ph, texCoords[2][0], texCoords[2][1],
+			0.f, 0.f, texCoords[0][0], texCoords[0][1],
+			pw,  0.f, texCoords[1][0], texCoords[1][1],
 		};
```
valad47, line 853: So to properly rotate video and match QRhi vertex coords layout, we have to rotate texture coords and then use indexes in order that is correct for vertex coords

john-preston:
Thank you for the fix!

PR #30763, opened 2026-06-04
Add Chats accessibility automation id to the chat list

rezabakhshilaktasaraei:
﻿## Summary

Sets the object name of the dialogs chat list (`Dialogs::InnerWidget`, accessibility role `List`) to `Chats`, so it is exposed with a stable UI Automation **AutomationId** of `Chats`.

Without an explicit object name the chat list has no AutomationId at all (Qt builds it from the object-name chain, and the list own object name is unset), so UI automation and accessibility tooling have no stable handle for it. A fixed AutomationId gives them a reliable, language-independent way to locate the chat list. (To be precise: the AutomationId is not derived from the localized accessible name - this is about giving the list a stable id, not replacing its name.)

### Why this is useful

Screen-reader users rely on add-ons that bind a hotkey to jump straight to a region of the app - e.g. the UnigramPlus NVDA add-on uses ALT+1 for the chat list, ALT+2 for the message list, etc. Because Unigram exposes no stable ids, that add-on matches localized UI names, and its README warns it "may not work in some localizations" and asks the user to select their interface language. A fixed `Chats` AutomationId lets an equivalent Telegram Desktop add-on locate the chat list regardless of language.

## Implementation

- One line in the `InnerWidget` constructor: `setObjectName(u"Chats"_q);`, alongside the existing `setAccessibleName(...)`.

On the default Windows build (Qt 5.15.19), Qt UIA provider builds the AutomationId by walking the object from itself up through its parents and returning the accumulated name; since the list own object is named `Chats` and its ancestors are unnamed, the resulting AutomationId is exactly `Chats`. On Qt 6 the id is computed differently (an explicit identifier, else a dotted object-name/class-name path), so it would be a path ending in `Chats`. No behavior change for end users.

### DIFF
--- a/Telegram/SourceFiles/dialogs/dialogs_inner_widget.cpp
+++ b/Telegram/SourceFiles/dialogs/dialogs_inner_widget.cpp
@@ -300,6 +300,7 @@ InnerWidget::InnerWidget(
 , _freezeTimer([=] { _shownList->unfreeze(); update(); }) {
 	setAttribute(Qt::WA_OpaquePaintEvent, true);
 	setAccessibleName(tr::lng_recent_chats(tr::now));
+	setObjectName(u"Chats"_q);
 
 	style::PaletteChanged(
 	) | rpl::on_next([=] {


ilya-fedin:
> This gives accessibility tooling and UI automation tests a reliable, language-independent way to locate the chat list, independent of its (localized) accessible name.

I fail to understand the rationale. The default automation ID isn't language-dependent and this line doesn't provide any other explanation why it's unreliable. So it's not clear why this change is needed.

rezabakhshilaktasaraei:
﻿Good points - let me clarify the rationale (the original wording was misleading) and address the Qt-version concern.

**Rationale (the wording was imprecise).** You are right that the AutomationId is not derived from the localized accessible name, so framing it as "independent of the localized name" was confusing. The real situation is simpler: `Dialogs::InnerWidget` has no objectName, so its AutomationId is empty - there is nothing for UI automation / accessibility tooling to bind to. Setting `objectName("Chats")` gives it a stable, language-independent AutomationId (`Chats`). The concrete use case is a screen-reader add-on that binds a hotkey to jump to the chat list: the UnigramPlus NVDA add-on does exactly this (ALT+1 -> chat list, ALT+2 -> message list, ...), but since Unigram exposes no stable ids it has to match localized UI names, and its README warns it "may not work in some localizations".

**Qt 5.15 (the default Windows build).** I checked the qtbase 5.15.19 source we compile (`src/plugins/platforms/windows/uiautomation/qwindowsuiamainprovider.cpp`):

```cpp
QString QWindowsUiaMainProvider::automationIdForAccessible(const QAccessibleInterface *accessible)
{
    QString result;
    if (accessible) {
        QObject *obj = accessible->object();
        while (obj) {
            QString name = obj->objectName();
            if (name.isEmpty())
                return result;          // accumulated so far, not ""
            if (!result.isEmpty())
                result.prepend(u'.');
            result.prepend(name);
            obj = obj->parent();
        }
    }
    return result;
}
```

It walks from the element own object upward and, on an empty objectName, returns the accumulated `result` (not an empty string). The list object is named `Chats`, so it is captured on the first iteration; the next (unnamed) parent then returns `result`, which is already `Chats`. So on the default Qt 5.15.19 build the chat list AutomationId is exactly `Chats` - and without the objectName it is empty, which is the gap this closes.

**Qt 6.** There it is computed differently (`QAccessibleBridgeUtils::accessibleId` - an explicit identifier, else a dotted objectName/class-name path), so the result would be a path ending in `Chats` rather than exactly `Chats`. If identical ids across Qt 5/6 are wanted, I can additionally set the accessible identifier on Qt 6.

I will update the PR description to reflect this.

ilya-fedin:
I remember you were complaining about scripting in the past and I came up with https://github.com/desktop-app/patches/blob/master/qtbase_5.15.19/0032-accessible-class-name-from-rtti.patch / https://github.com/desktop-app/patches/blob/master/qtbase_6.11.1/0028-accessible-class-name-from-rtti.patch. I.e. this problem should be already fixed and this PR be redundant.

rezabakhshilaktasaraei:
﻿Fair point - that patch does give a stable, language-independent ClassName, so the "can a tool find it at all" need is already covered. Thanks for the reminder.

One thing I'd like to confirm before closing this. `typeid(*o).name()` isn't consistent across compilers - on MSVC it's like `class Dialogs::InnerWidget`, on GCC/Clang it's the mangled form (e.g. `N7Dialogs11InnerWidgetE`) - and it changes if the class is ever renamed. `AutomationId` is the UIA property meant for selecting a specific control, and a fixed string would be identical across platforms and rename-proof.

So which would you recommend add-on authors target for these accessibility entry points - the RTTI ClassName, or a dedicated AutomationId? If ClassName is the intended convention I'm happy to close this as redundant.

ilya-fedin:
> AutomationId is the UIA property meant for selecting a specific control, and a fixed string would be identical across platforms and rename-proof.

But UIA is a Windows-specific thing, it couldn't be cross-platform?

rezabakhshilaktasaraei:
You're right - UIA/AutomationId is Windows-only, so scratch the cross-platform wording.

The practical difference that's left: for the NVDA add-on I'm writing, `AutomationId = "Chats"` is simpler to target than `ClassName = "class Dialogs::InnerWidget"` - no compiler `class ` prefix, and it won't break if the class is renamed.

Here's the same chat-list element from NVDA's developer info, English vs Russian UI:

| property | English UI | Russian UI |
| --- | --- | --- |
| `name` (accessible name) | `Chats` | `Чаты` |
| `UIA automationID` | `Chats` | `Chats` |
| `UIA className` | `class Dialogs::InnerWidget` | `class Dialogs::InnerWidget` |

The accessible name localizes, while `automationID` (and `className`) stay put - so an add-on must not key off `name`, and `automationID = Chats` is the cleanest stable anchor.

That's a minor convenience on top of what your ClassName patch already covers, though. If you'd rather add-on authors rely on ClassName, I'm happy to close this as redundant - your call.

ilya-fedin:
> The practical difference that's left: for the NVDA add-on I'm writing, `AutomationId = "Chats"` is simpler to target than `ClassName = "class Dialogs::InnerWidget"` - no compiler `class ` prefix, and it won't break if the class is renamed.

If you insist that automation id should be human-readable then a question arises: are you going to label EVERY class by hand? This doesn't sound good to me. We need to automate that, e.g. rework the API that sets accessible name to automatically also set accessible identifier but as in english locale. And then it sounds like it should be accessible identifier exactly, not object name. But accessible identifiers are Qt 6.9+ so Windows users aren't going to get them unless Qt is patched to backport it.

If it shouldn't be human-readable then class name should be ok I believe.

rezabakhshilaktasaraei:
﻿Good idea - auto-deriving the identifier from the English name (instead of hand-labeling) is cleaner and scales. And agreed it should be the accessible *identifier*, not objectName - I actually already switched to that: lib_ui sets `QAccessible::Identifier` on Qt 6 and falls back to `QAccessible::UserText` read by a Qt 5.15 backport patch (identifiers are 6.8+), so Windows users get it too.

So the plan: rework the accessible-name setter so it also sets the identifier to the English value of the same lang string (we have the default English values), rather than per-widget ids. Add-on authors then target the stable English names. I would keep a manual `setAccessibleId` override for the handful of entry points where we want a short, unique, guaranteed-stable id. Sound good?

ilya-fedin:
> Add-on authors then target the stable English names. 

I would like to correct you: the chance that class name is changed is way less likely than English translation, since English translation could be changed at any time via the translation platform, with no code changes.

john-preston:
@ilya-fedin If I understood correctly the idea is to set literal english names to some objects here and there, manually, not connected to translations platform in any way.

ilya-fedin:
> If I understood correctly the idea is to set literal english names to some objects here and there, manually, not connected to translations platform in any way

Yeah, his conclusion is weird, class name should be enough

rezabakhshilaktasaraei:
Closing this - after discussion we're going with Qt's automationId class-name-chain fallback (to be backported to Qt 5.15) instead of hand-set object names/identifiers, so widgets get a stable id from their class hierarchy automatically with no per-widget code. Thanks for the review!

PR #30821, opened 2026-06-13
Fix inverted touchpad swipe gestures on Linux

### LINKED ISSUE #29573: Touchpad gestures on Linux inconsistent with other software
### Steps to reproduce

Swipe sideways in chat view or chat list

### Expected behaviour

For example, moving two fingers left on a touchpad smoothly "moves" a message left for a reply

### Actual behaviour

Swipe left triggers a back action, and swipe right works as a reply. This behavior is inconsistent with visualization and with other software (like browsers). Likely this is made with traditional scrolling in mind, when most desktops (like Gnome, and i think last versions of KDE) use natural scrolling.

### Operating system

Fedora Silverblue 42

5.16.4

### Installation source

### Crash ID

_No response_

### Logs

```shell

```

erik-balfe:
Linux touchpad swipes are left/right reversed vs macOS.

`swipe_handler.cpp` flips wheel delta only when `QWheelEvent::inverted()` is true. Qt on Linux often leaves it false for touchpads even with natural scrolling (default on GNOME/KDE).

Fix: invert `pixelDelta` wheel events on Linux.

Fixes #29573

### DIFF
--- a/Telegram/SourceFiles/ui/controls/swipe_handler.cpp
+++ b/Telegram/SourceFiles/ui/controls/swipe_handler.cpp
@@ -359,7 +359,11 @@ void SetupSwipeHandler(SwipeHandlerArgs &&args) {
 			if (cancel) {
 				processEnd();
 			} else {
-				const auto invert = (w->inverted() ? -1 : 1);
+				const auto isTouchpad = !w->pixelDelta().isNull();
+				const auto invert = (w->inverted()
+					|| (base::Platform::IsLinux() && isTouchpad))
+					? -1
+					: 1;
 				const auto delta = Ui::ScrollDeltaF(w) * invert;
 				updateWith({
 					.globalCursor = w->globalPosition().toPoint(),


Telegram/SourceFiles/ui/controls/swipe_handler.cpp
```diff
@@ -359,7 +359,11 @@ void SetupSwipeHandler(SwipeHandlerArgs &&args) {
 			if (cancel) {
 				processEnd();
 			} else {
-				const auto invert = (w->inverted() ? -1 : 1);
+				const auto isTouchpad = !w->pixelDelta().isNull();
+				const auto invert = (w->inverted()
+					|| (base::Platform::IsLinux() && isTouchpad))
```
ilya-fedin, line 364: That's absolutely not right, you're going to invert this for people using normal scrolling (I do). The current code following the Qt hint is right. If it doesn't work for you, you should report to Qt so that they fix it.

PR #30841, opened 2026-06-16
Update TooManyCooks to latest to support LoongArch

tzcnt:
The latest commit of TooManyCooks supports LoongArch 64-bit.


pfeatherstone:
Wow TMC has made it into a big project !

PR #30845, opened 2026-06-17
Add an accessibility option to disable animated avatars

### LINKED ISSUE #24658: [Feature Request] Add reduce motion option

Animated avatars are extremely distracting and not accessible. These avatars can show anything, including flashing colors, which is dangerous for people sensitive to flashing lights (epileptics).

Add a reduce motion option which would show animated avatars in places other than user profile (contacts, chat list and next to chat bubbles) only on hover.
Discord enables avatar animations only on hover for everyone and it works well. Users can still see the animation without going to another user's profile, but only when they intend to click on the profile or react to messages.

Completely disable animated avatars in places other than user profile when the reduce motion option is set. I like the previous solution more though as hovering over the user or their messages doesn't happen all the time.

### Additional context

This option is already available in the iOS client.

C0rn3j:
Fixes https://github.com/telegramdesktop/tdesktop/issues/24658

Please note the comments in the linked thread this fixes, and comments [such as these](https://bugs.telegram.org/c/18322/4) on older issues for other platforms where this autoplay is no longer the case, as for the reasoning why this accessibility toggle is necessary.

---

This PR adds an `Animated avatars` accessibility toggle into the `Battery and Animations` view:

image

Result of which can be seen here (left side is unmodified TG, right side is this PR):
[Screencast_20260617_123747.webm](https://github.com/user-attachments/assets/cd46b3b7-085b-46ae-983b-916b059f81f1)

*Disclaimer: This PR was assisted by an LLM*

---

I have read the [CONTRIBUTING.md](https://github.com/telegramdesktop/tdesktop/blob/61caec4b30e4ea5bce9477117ef82cb337c55e2c/.github/CONTRIBUTING.md), so I am aware that new UI elements are forbidden from being in PRs.

If you consider adding this option as such, let me know, and I can rework it to be under the existing `Interface animations` toggle (which I do actually otherwise want to be on), and after the PR is merged I shall open a new issue about separating it into two.

### DIFF
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -963,6 +963,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 "lng_settings_power_chat_effects" = "Effects in messages";
 "lng_settings_power_calls" = "Animations in Calls";
 "lng_settings_power_ui" = "Interface animations";
+"lng_settings_power_animated_avatars" = "Animated avatars";
 "lng_settings_power_auto" = "Save Power on Low Battery";
 "lng_settings_power_auto_about" = "Automatically disable all animations when your laptop is in a battery saving mode.";
 "lng_settings_power_turn_off" = "Please turn off Save Power on Low Battery to change these settings.";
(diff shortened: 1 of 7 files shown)


ilya-fedin:
I'm doubt this will be merged since animated avatars is a paid feature and people basically pay to ALWAYS annoy you with them. I guess you're better to send this to some fork.

f242:
Yeah, all those accessibility options to reduce flashing are for wimps! Once your epileptic seizure is over, just get up off the floor, dust yourself off, and keep using it like nothing happened😆

PR #30893, opened 2026-06-28
Fix use-after-free when deleting multiple folders.

gearonixx:
Telegram crashes when the user deletes more than 2 folders 
in the Settings -> Folders widget.

fixed

### DIFF
--- a/Telegram/SourceFiles/dialogs/dialogs_pinned_list.cpp
+++ b/Telegram/SourceFiles/dialogs/dialogs_pinned_list.cpp
@@ -66,7 +66,7 @@ void PinnedList::setPinned(Key key, bool pinned) {
 		const auto index = int(it - begin(_data));
 		_data.erase(it);
 		key.entry()->cachePinnedIndex(_filterId, 0);
-		for (auto i = index, count = int(size(_data)); i != count; ++i) {
+		for (auto i = index; i < int(size(_data)); ++i) {
 			_data[i].entry()->cachePinnedIndex(_filterId, i + 1);
 		}
 	}


john-preston (changes requested):
Thanks for tracking this down. This is the right lifetime axis, but the current patch does not actually force the content tree to be destroyed before `_state`.

`_content` is constructed with `this` as the parent, and `object_ptr` intentionally does not delete parented objects from its destructor. That means the widget tree still survives until Qt child cleanup in the base `QWidget`/`QObject` destructor, after `Folders` members such as `_state` have already been destroyed.

Please explicitly destroy the content at the end of `Folders::~Folders()`, after the pending save while `_state` is still alive:

```cpp
Folders::~Folders() {
	if (!Core::Quitting() && _state->save) {
		_state->save(nullptr, nullptr);
	}
	_content.destroy();
}
```

This keeps the save path's access to the row widgets intact, then tears down the child lifetimes/subscriptions while the captured `FoldersState` is still valid.

gearonixx:
Done.

john-preston:
Do you have a crash report (Crash ID) from 6.9.3 crashing in this state? I can't reproduce right now. On any system, but in the version from https://desktop.telegram.org - doesn't crash with me. Can you reproduce? If you have "Install beta versions" toggled, it will suggest sending a crash report on next relaunch and there will be "Crash ID". Please submit the report and copy ID here.

gearonixx:
6.9.3 (6009003) Linux x86_64

Crash ID (6.9.3): `87036d61-c230-41a7-4f029cae-747acddc`

(same crash on older versions - 6.5.1: `d22873f4-6d6a-495a-3a649d90-509b6a2e`)

---

Crash ID (6.9.3 - https://desktop.telegram.org): `a06eec6e-9b78-4e21-c5771686-c98c8792`

john-preston:
Well, look. I'll be fixing the crash from the dump you sent, but it is not fixed in this PR, I'm not sure if this PR fixes something really happening or not:

image

john-preston:
(both first and last crash-dumps happen not in ~Folders, but later).

gearonixx:
### Steps to reproduce:

1. create a folder
2. pin 4 chats
2. right-click -> remove

gearonixx:
The crash was triggered by pinned chats. Applied the fix, no longer crashes.

gearonixx:
You can reproduce, compile with the patch, and see for yourself.

gearonixx:
> * create a folder
> 
> * pin 4 chats
> 
> * right-click -> remove


If it doesn't crash, **change the pin order** of the chats and delete the folder again.

The crash depends on the **position of the removed chat** in the pinned list.

---

When a folder is deleted, the pinned chats inside it are removed from the filter's chat list one by one. Each removal of a pinned chat enters `PinnedList::setPinned`, which erases the entry and then runs a loop to reindex the remaining pinned entries.

The problem is that this loop captured the list's length once before starting, so when the reindexing recursively triggered another removal, the list shrank underneath it while the cached length stayed the same.

As a result, the loop kept iterating up to the stale length and eventually indexed `_data` past the vector's real end, tripping the libstdc bounds assertion and aborting the app.

john-preston:
Yes, now the patch does fix the crash, thanks. I'll also add a fix to address the re-entry that caused this code to not work initially, but this one can be merged too. Thanks.

PR #30912, opened 2026-06-30
[img-editor] Fixed std::clamp() crash in photo editor

### LINKED ISSUE #30921: Crash when trying to draw in the built-in photo editor
### Steps to reproduce

1. Open any image from the chat or gallery
2. Press the edit button (pencil) to switch to drawing mode
3. Select the "line" tool (or any other)
4. Start drawing a line on the image
5. The app closes instantly (crash)

### Expected behaviour

A drawn line appears on the image, the editor works without drops.
Related closed issues: #30644, #30659. Both were closed as completed for version 6.7.8, but the crash persists in 6.9.3 with a different assertion (`std::clamp`).

### Actual behaviour

Instant crash

### Operating system

Fedora Linux 43 (Workstation Edition)

6.9.3

### Installation source

Flatpak

### Crash ID

_No response_

### Logs

```shell
logs: qt.gui.imageio.jpeg: Corrupt JPEG data: premature end of data segment  
qt.gui.imageio.jpeg: Corrupt JPEG data: premature end of data segment  
/usr/include/c++/15.2.0/bits/stl_algo.h:3638: constexpr const _Tp& std::clamp(const _Tp&, const _Tp&, const _Tp&) [with _Tp = double]: Assertion '!(__hi < __lo)' failed.
```

mmistika:
There is a bug in the photo editor. When starting drawing near the bottom or right border of the image, the brush padding extends outside the frame boundaries. This causes a crash due to a violation in `std::clamp` where the lower bound ( lo ) becomes greater than the upper bound ( hi ).

This PR resolves the crash by using `std::min` to ensure the lower limit passed to `std::clamp` is never greater than the upper limit:
- Clamping the right boundary using `std::min(contentRight, sceneSize.width())` as the lower bound.
- Clamping the bottom boundary using `std::min(contentBottom, sceneSize.height())` as the lower bound.

Potentially closes #30921

### DIFF
--- a/Telegram/SourceFiles/editor/scene/scene_item_canvas.cpp
+++ b/Telegram/SourceFiles/editor/scene/scene_item_canvas.cpp
@@ -78,6 +78,8 @@ void ItemCanvas::computeContentRect(const QPointF &p) {
 	const auto sceneSize = scene()->sceneRect().size();
 	const auto contentLeft = std::max(0., _contentRect.x());
 	const auto contentTop = std::max(0., _contentRect.y());
+	const auto contentRight = contentLeft + _contentRect.width();
+	const auto contentBottom = contentTop + _contentRect.height();
 	_contentRect = QRectF(
 		QPointF(
 			std::clamp(p.x() - _brushMargins.left(), 0., contentLeft),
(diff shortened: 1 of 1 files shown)


john-preston:
Thanks!

PR #30932, opened 2026-07-05
Migrate webpage timestamp links to wrapped URL format

### LINKED ISSUE #30888: Regression: can no longer copy or drag YouTube timestamp links
### Steps to reproduce

1. Send a YouTube video link as a message. Make sure the link preview doesn't include Telegram's own player for the video (with a black, circular play button, shown below).

   Image

   It should be purely a link (with a red, rounded-rectangle play button, shown below).

   Image

2. Reply to the message with the a timestamp, like 1:00.

   Image

3. Right-click on the timestamp within the message you just sent, or try to click and drag it into the message field or into a browser window.

### Expected behaviour

Upon right-clicking, you should see the option "Copy link", allowing you to copy the YouTube timestamp link (for example, https://www.youtube.com/watch?v=yBwD4iYcWC4&t=60).

Image

Alternatively, you should be able to drag the timestamp into the message field (to paste the link there), showing a cursor similar to this:

Image

or into a browser window (to open it in that window, next to or in place of an existing tab).

### Actual behaviour

There is no "Copy link" option, and you're unable to drag the YouTube hyperlink anywhere.

This means the only way you can interact with it is by clicking it, which forcibly opens the video in a new tab at the end of your tabs list in your default browser.

This is highly inconvenient, especially when you have a lot of tabs.

This was working as of v6.8.2 and was not working in v6.9.1, but I don't know which version it broke in, as I skipped a few versions when updating.

It may be linked to the "Fix clickable timestamps in blockquote" bullet point in [the release notes of v6.8.4 beta in the changelog](https://telegramdesktop.github.io/tdesktop/changelog/#v6.8.4).

### Operating system

Windows 7

6.9.3

### Installation source

### Crash ID

_No response_

### Logs

```shell

```

mmistika:
This migrates webpage timestamp links to the wrapped URL format (`internal:wrapped?url=...`). This fixes a regression introduced in version 6.9.0 where YouTube timestamp links lost their drag-and-drop capabilities and "Copy Link" option in the context menu.

Closes #30888.

### DIFF
--- a/Telegram/SourceFiles/history/view/media/history_view_media.cpp
+++ b/Telegram/SourceFiles/history/view/media/history_view_media.cpp
@@ -6,6 +6,7 @@ For license and copyright information please follow this link:
 https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 */
 #include "history/view/media/history_view_media.h"
+#include "ui/basic_click_handlers.h"
 
 #include "boxes/send_credits_box.h" // CreditsEmoji.
 #include "history/history.h"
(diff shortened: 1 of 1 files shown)


john-preston:
Thanks!

Issue #30847, opened 2026-06-18
Crash (out-of-bounds / _GLIBCXX_ASSERTIONS abort) when seeking/playing a voice message


siberianwm:
### Steps to reproduce
1. Open a chat containing a voice message.
2. Click a position on the voice message's timeline/waveform to seek. (main trigger; the crash also happens without any click — see note)
3. Telegram aborts (SIGABRT).

### Expected behaviour
Playback seeks to the clicked position / continues normally.

### Actual behaviour
The app aborts via a libstdc++ hardened assertion (`__glibcxx_assert_fail` → `abort`) — an out-of-bounds index in the voice-message seek / position handling.

Across 6 collected systemd coredumps the abort signature is identical (`__glibcxx_assert_fail`). 3 of them contain a mouse-event frame (`QApplicationPrivate::sendMouseEvent` → `QWidget::event`) — i.e. clicking the timeline to seek. The other 3 carry the same assertion but no mouse frame, so the OOB is also reached without any click — most likely from a playback position update (e.g. the voice message reaching its end).

Note: this only *crashes* on builds compiled with `_GLIBCXX_ASSERTIONS` (the default hardening on Arch / CachyOS). On the official static build the same out-of-bounds is silent undefined behaviour and does not abort — so it will not surface in the in-app / beta crash reporter, but the OOB access is present in the code regardless. It looks like the same family as the already-fixed MPRIS seek-offset index bug (#24194 / desktop-app/lib_base#116), but in the voice-message UI / playback seek path.

### Stack trace (systemd coredump, click variant)
```
#2  abort                       (libc.so.6)
#3  __glibcxx_assert_fail       (libstdc++.so.6)
#4  Telegram + 0x88e1f4
#5  Telegram + 0x3da2123
#6  Telegram + 0x3cacd77
#7  Telegram + 0x31188a7
#8  Telegram + 0x338b39e
#9  Telegram + 0x4ce2dc7
#10 Telegram + 0x353b372
#11 Telegram + 0x353be69
#12 QWidget::event              (libQt6Widgets.so.6)
#13 QApplicationPrivate::notify_helper
#14 QApplication::notify
#17 QApplicationPrivate::sendMouseEvent
#23 QGuiApplicationPrivate::processMouseEvent
```
(Telegram frames are address-only — the distro build is stripped.)

### Operating system
Arch Linux / CachyOS, KDE Plasma 6 (Wayland)

### Version of Telegram Desktop
6.9.3

[...]

Issue #26104, opened 2023-03-29
Telegram crashes with code 134 when trying to watch the video (normal view, not autoplaying preview)


Farxial:
### Steps to reproduce

Upload to Telegram and try to watch the video /\*RU: Загрузить и попытаться посмотреть видео\*/ https://drive.google.com/file/d/1h8hSs0Rkh1mxmE6Ko9GhdeMw8bsLfkAt/view?usp=sharing
This is a video fragment obtained from (renamed) /\*RU: Это фрагмент видео, полученный из (переименованного)\*/
[Magical_Pokahn_TV_[09_of_12]\_[ru_jp]\_[Cuba77_&_AnimeReactor_Ru].avi](https://drive.google.com/file/d/1Buz_cvrQ0-TsKpHE7KlieHpdR0OYrzY3/view?usp=sharing) (Size: 253736960; Hashes: MD5: 7167E387495A886189F4C5EDCE27B7C1, SHA1: 3ED0D17A99F087D1C0C407D9D0CD493974CDF122, SHA256: 687C63B3CA30FE55AF59F75F9780624F2B62FD00E05FD63A19AC6CCCDF54CC96, MD4: 1B431046357CEFAAA484F8E4AE5CCFCE, ED2K-V2: 31FFB6DF0EBF567E020C90AAD016D0E6)
using script /\*RU: скриптом\*/ (ffmpeg-tpart):
```
#!/bin/sh

export NAME="${1}"
export EXT=${2}
export SUB_NAME="${3}"
export START=${4}
export END=${5}
export EXT2=${6}

#ffmpeg -ss ${START} -i "${NAME}.${EXT}" -to ${END} -c copy -copyts "${NAME} ${SUB_NAME}.${EXT}"
ffmpeg -ss ${START} -i "${NAME}.${EXT}" -to ${END} -vcodec libx264 -acodec mp3 -copyts "${NAME} ${SUB_NAME}.${EXT2}"
```
`ffmpeg-tpart "09. 絶交の呪文は色男さんいらしゃーい ̸ 白の呪文はクリスマス" avi Batata 01:29.98 01:59.52 mp4`
The commented line didn't work either. /\*RU: С закомментированной строкой тоже не работало.\*/

### Expected behaviour

Do not crash (and even better play the video: yes, ffmpeg converted it crookedly, but Mozilla Firefox (Linux) and The KMPlayer 4.2.1.4 KMP Plus (Windows) play it).
[...]

PR #30970, opened 2026-07-13
Fix media viewer crash when seeking without duration

Mellow-Sky:
# Fix media viewer crash when seeking without duration

## Problem

The media viewer can handle keyboard seek actions before streamed track
information contains a usable duration. At that point
`video.state.duration` is `kTimeUnknown`, while the current position may
already be valid.

`seekRelativeTime()` passes that state to
`std::clamp(position + delta, 0, duration)`. This violates the clamp
precondition because the upper bound is below zero and aborts hardened
libstdc++ builds.

Two Telegram Desktop 6.9.3 core dumps reproduced the same path:

- Right Arrow: `position = 78533`, `delta = 5000`,
  `duration = kTimeUnknown`.
- Left Arrow: `position = 371933`, `delta = -5000`,
  `duration = kTimeUnknown`.

## Fix

- Ignore relative seeks until both position and duration are known and the
  duration is available.
- Ignore progress-based seeks while duration is unknown or unavailable.
- Keep `restartAtSeekPosition()` unchanged because it is also used to start
  playback before track information is ready.

Normal seeks retain the existing clamping behavior.

## Verification

- Verified both core-dump call sites against the v6.9.3 source and the
  Arch Linux binary disassembly.
- Verified the unknown-duration and unavailable-duration paths return before
  position arithmetic or playback restart.
- Verified normal lower- and upper-bound clamping is unchanged.
- `git diff --check` passes.
- A full local build was not run because no configured TDesktop build
  environment is available on this machine.

Possibly related: #26104 and #30847.

### DIFF
--- a/Telegram/SourceFiles/media/view/media_view_overlay_widget.cpp
+++ b/Telegram/SourceFiles/media/view/media_view_overlay_widget.cpp
@@ -5348,18 +5348,31 @@ void OverlayWidget::flushPendingFrameStep() {
 void OverlayWidget::seekRelativeTime(crl::time time) {
 	Expects(_streamed != nullptr);
 
+	const auto &state = _streamed->instance.info().video.state;
+	const auto position = state.position;
+	const auto duration = state.duration;
+	if (position == kTimeUnknown
+		|| duration == kTimeUnknown
+		|| duration == kDurationUnavailable) {
+		return;
+	}
 	const auto newTime = std::clamp(
-		_streamed->instance.info().video.state.position + time,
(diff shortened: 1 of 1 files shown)


ilya-fedin:
> * A full local build was not run because no configured TDesktop build
>       environment is available on this machine.

I assume this means you didn't test your own changes? I'm not sure what's the point to upload a change if you aren't even sure it fixes the problem...

john-preston:
Thanks, looks correct.

PR #30986, opened 2026-07-15
Paste Excel selections as article tables

comdev1337:
I aislopped Excel selection pasting as article tables, which I find rather useful. It works on my machine, but it'd definitely benefit from adding text entities parsing.

image

### DIFF
--- a/Telegram/SourceFiles/iv/editor/iv_editor_clipboard.cpp
+++ b/Telegram/SourceFiles/iv/editor/iv_editor_clipboard.cpp
@@ -11,6 +11,9 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 
 #include <QtCore/QMimeData>
 #include <QtCore/QPointer>
+#include <QtCore/QStringList>
+
+#include <array>
 
 namespace Iv::Editor {
 namespace {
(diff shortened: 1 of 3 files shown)


john-preston:
@comdev1337 Findings
1. [P1] Bound and defer table construction
Location: iv_editor_clipboard.cpp:178–207, called from iv_editor_widget.cpp:8677
ExcelTableBlockFromMimeData() eagerly creates three representations of the selection: row strings, QStringLists for every row, and finally a RichPage::TableCell for every field. The inline handler performs all this work before checking whether the operation is only MimeAction::Check. InputField invokes the hook independently for capability checking and actual insertion, so large selections can be fully materialized during both paths. The conversion also happens synchronously before the code posts anything through crl::on_main.   
Most of that work can be guaranteed to fail later. The default article limits are 500 blocks and 20 table columns, and every table row increments blockCount in addition to the table block itself. Validation only occurs after constructing a candidate state. Thus, a 100,000-row range can allocate a huge object graph on the UI path and then be rejected because the table alone represents 100,001 blocks.   
Reproduction: Copy a whole Excel column, or a large rectangular range containing short or empty values, then paste or trigger a paste-capability check in the article editor. The exact stall threshold is machine-dependent, but the work is currently unbounded.
Recommended fix: Separate cheap Excel recognition from conversion. Check should inspect only the available formats. Insert should scan once with a budget derived from the current article’s remaining block, column, and text limits, stopping before allocating all rows and cells. Avoid retaining rows, cells, and the completed RichPage table simultaneously.
2. [P2] Do not consume clipboard data that the article state will reject
Location: iv_editor_widget.cpp:4063 and iv_editor_widget.cpp:8677
Both new dispatch paths consume the paste as soon as ExcelTableBlockFromMimeData() returns a block. However, that function only checks that the flattened text is rectangular; it does not check article limits. If insertion later fails, pasteStructuredClipboardData() merely shows the limit toast. The inline hook has already returned true, causing InputField to skip its ordinary HTML/plain-text paste path, while the hidden-field path has already accepted the keyboard event.    
Under the default configuration, copying A1:U2 produces a rectangular 21-column table that the parser accepts but state validation rejects because maxTableCols is 20. The result is that standard paste inserts neither a table nor the existing text/media fallback. The same occurs for row counts that push the complete article above its block limit.  
Reproduction: Copy a two-row, 21-column Excel range and press Ctrl+V in an article paragraph using the default limits. Only the limit error is produced; no clipboard content is inserted.
Recommended fix: Preflight the candidate against the current resolved article limits before claiming the MIME data. When table insertion is not valid, return false so the existing text/media path runs, or explicitly retry the paste as plain text.
3. [P2] Preserve text-paste behavior in raw math fields
Location: iv_editor_widget.cpp:8677
The existing native structured-paste branch requires ClipboardPasteInsertContext(activeTextInsertContext()). activeTextInsertContext() deliberately returns nullopt for FieldMode::Raw, which includes the formula editor. The new Excel branch does not use that guard and therefore intercepts Excel clipboard data while a raw field is active.  
With no insertion context, pasteStructuredClipboardData() commits the visible inline field and performs block-level insertion after the active block. By contrast, when the MIME hook declines the data, InputField falls through to its normal text insertion logic.   
Reproduction: Edit an existing math formula, copy a single Excel cell containing x^2, and press Ctrl+V. The formula edit is committed and a one-cell article table is inserted after the math block, instead of x^2 being pasted into the formula field.
Recommended fix: Apply the same rich insertion-context guard to the Excel branch. The separate hidden-field keyboard path can continue providing structural table insertion when no inline field is being edited.
4. [P2] Do not treat every embedded line break as a new table row
Location: iv_editor_clipboard.cpp:117–137
Excel supports explicit line breaks inside an individual cell. The parser nevertheless uses flattened mimeData->text() and treats every CR or LF as a row separator, followed by every tab as a column separator. There is no quoted-field handling and no use of the actual HTML/BIFF cell boundaries. [Microsoft Support](https://support.microsoft.com/en-us/excel/start-a-new-line-of-text-inside-a-cell-in-excel)  
A one-column selection containing a cell such as first<Alt+Enter>second is therefore interpreted as two table rows when the embedded line break is exposed through QMimeData::text(). In multi-column cases, it may instead make row widths inconsistent and unexpectedly disable table conversion.
Recommended fix: Parse the Excel HTML table—or a native table representation—so cell boundaries are known before interpreting newlines and tabs. A quoted-TSV parser is a weaker fallback. Parsing the HTML would also provide a path to preserving hyperlinks and other text entities, which the current plain-string assignment drops.

PR #31018, opened 2026-07-20
Include muted unread dialogs in unread-on-top

rodion-m:
## Summary

The existing experimental unread-on-top option only promotes unread state that is not muted. Muted unread chats can therefore remain mixed with read dialogs and still be missed while the option is enabled.

This change makes the existing option cover all unread dialog state:

- unread messages and manual unread marks, including muted ones
- unread reactions, mentions, and polls

It keeps the current sort-key infrastructure and ordering unchanged: fixed/system entries first, then per-filter pinned dialogs, then unread dialogs ordered by date, then ordinary dialogs. The option remains opt-in and defaults to off.

## Validation

- Built the `Telegram` target in Debug on macOS for arm64 and x86_64 with compiler warnings treated as errors.
- Manually verified that a muted unread dialog is placed above a newer read dialog.
- Verified pinned dialogs remain above the unread block and unread dialogs retain date order.
- `git diff --check` passes.

### DIFF
--- a/Telegram/SourceFiles/dialogs/dialogs_entry.cpp
+++ b/Telegram/SourceFiles/dialogs/dialogs_entry.cpp
@@ -55,8 +55,8 @@ uint64 UnreadOnTopDialogPos(uint64 sortKeyByDate) {
 
 base::options::toggle OptionUnreadOnTop({
 	.id = kOptionDialogsUnreadOnTop,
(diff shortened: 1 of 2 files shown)


rodion-m:
Closing because the existing upstream option already covers the intended use case, and extending it to muted unread chats is not needed.

PR #31022, opened 2026-07-20
Match kinetic scroll timer interval with display refresh rate

poli0iq:
Qt animates QScroller kinetic scrolling at fixed 16 ms / ~60 fps, so the inertia feels laggy on high refresh rate displays. Override the interval to match the highest refresh rate.


ilya-fedin:
I believe it should be in lib_ui where the QScroller machinery lives

PR #31023, opened 2026-07-20
smoother scrolling on macos

teidesu:
an attempt to make scrolling the chat history smoother on macos. two parts: render scroll positions with sub-pixel precision (needs desktop-app/lib_ui#324), and stop decoding images on the main thread, which caused visible stutter while scrolling media-heavy chats. it's not perfect but noticeably smoother.

- **skip thumbnail bytes generation for local stickers** — local stickers were re-encoding a WEBP thumbnail that was never used; skip it via a new `generateThumbnailBytes` flag.
- **decode sticker images without blocking the main thread** — `checkStickerLarge` constructed the `Image` (synchronous decode) on the main thread; move the decode to `crl::async` with a loading guard.
- **decode downloaded and partial cached images asynchronously** — `FileLoader` now also decodes `partial:`-prefixed cache entries off-thread when the partial data covers the requested size, and decodes freshly downloaded images on `crl::async` before firing updates instead of leaving it to the consumer on the main thread.
- **apply sub-pixel scroll offset in history view** — `HistoryInner` subscribes to the scroll view's sub-pixel offset, translates painting by the fraction (keeping the floating date badge and forum bar pinned) and compensates in all hit-testing paths.

note: the code in this pr was written partially by claude fable 5 and partially by gpt-5.6-sol.

### DIFF
--- a/Telegram/SourceFiles/chat_helpers/stickers_lottie.cpp
+++ b/Telegram/SourceFiles/chat_helpers/stickers_lottie.cpp
@@ -332,15 +332,15 @@ not_null<DocumentData*> GenerateLocalSticker(
 		.caption = {},
 		.idOverride = LocalStickerId(path),
 	});
-	task.process({ .generateGoodThumbnail = false });
+	task.process({
+		.generateGoodThumbnail = false,
+		.generateThumbnailBytes = false,
+	});
 	const auto result = task.peekResult();
 	Assert(result != nullptr);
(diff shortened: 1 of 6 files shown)


Prana111025 (commented):
Changed

teidesu:
removed subpixel scrolling related changes as per https://github.com/desktop-app/lib_ui/pull/324#issuecomment-5052773857

keeping the async decoding fixes still

PR #31032, opened 2026-07-21
Mark messages read by focus instead of visibility with screen readers

rezabakhshilaktasaraei:
For a screen reader user, on-screen visibility says nothing about what was actually read: just scrolling (or the view moving on its own) marked messages read that were never spoken.

With a screen reader active, everything that inferred reading from visibility is now inert, and the read mark follows the accessibility focus instead, building on the focus-marking already added in aeb3e6be75 and 5f3e4d9a5e:

- deliberate navigation (arrows, PageUp/Down, UIA focus) marks everything up to the focused message - the same semantics as a sighted user paging to the bottom; the automatic pick when the list takes focus never marks anything;
- the unread bar becomes the stable landmark reading starts from. The bar's lifecycle was driven by the same visibility logic (destroyed on close once the bottom was seen, not recreated when the few unread messages fit on one screen, skipped by a saved scroll state or a jump to the end), so a visit after a partial read used to land on an arbitrary row and the first arrow press could sweep dozens of messages read. Now, while something actually stays unread, every visit rebuilds the bar above the first message still unread and lands the focus on it; a bar the reading advanced past is dropped on close so the next one sits at the new reading edge.

Without a screen reader nothing changes.

Tested with NVDA on Windows 10: opening a chat with unreads announces "Unread messages" first and leaves the counter untouched however many times the chat is reopened, arrowing reads and marks messages one by one, PageDown/End marks up to the landing point, and after reading part of the way a reopen lands on the bar again, moved down to exactly the first unheard message. The sighted behavior is unchanged.

### DIFF
--- a/Telegram/SourceFiles/history/history_inner_widget.cpp
+++ b/Telegram/SourceFiles/history/history_inner_widget.cpp
@@ -1478,7 +1478,9 @@ void HistoryInner::paintEvent(QPaintEvent *e) {
 				_controller->emojiInteractions().startAutoplay(view);
 			}
 		}
-		if (readTill && _widget->markingMessagesRead()) {
+		if (readTill
+			&& _widget->markingMessagesRead()
+			&& !Ui::ScreenReaderModeActive()) {
 			session().data().histories().readInboxTill(readTill);
 		}
 		if (markingAsViewed && !readContents.empty()) {
@@ -4413,6 +4415,9 @@ void HistoryInner::checkActivation() {
 		// Side-effect: Also clears all notifications from forum topics.
 		Core::App().notifications().clearFromHistory(_history);
 	}
+	if (Ui::ScreenReaderModeActive()) {
+		return;
+	}
 	if (_curHistory != _history || _history->isEmpty()) {
 		return;
 	}
@@ -6814,15 +6819,29 @@ void HistoryInner::announceAccessibilityFocusedChild() {
 		announceAccessibilityFocus(_accessibilityFocusedIndex);
 		return;
 	}
-	const auto barIndex = accessibilityUnreadBarIndex();
-	const auto index = (barIndex >= 0 && barIndex + 1 < count)
-		? (barIndex + 1)
-		: (count - 1);
 	const auto elements = accessibleElements();
-	const auto item = accessibilityItemAtIndex(
-		index,
-		elements,
-		barIndex);
+	const auto barIndex = accessibilityUnreadBarIndex();
+	auto index = barIndex;
+	if (index < 0) {
+		// The bar is not always there while messages stay unread (it is
+		// not recreated when the few unread messages fit on one screen):
+		// land on the first unread message itself then, so that every
+		// visit starts at the reading edge and the first arrow press
+		// marks one heard message, not everything above the newest row.
+		_history->calculateFirstUnreadMessage();
+		if (const auto first = _history->firstUnreadMessage()) {
+			for (auto i = 0, n = int(elements.size()); i != n; ++i) {
+				if (elements[i] == first) {
+					index = i;
+					break;
+				}
+			}
+		}
+	}
+	if (index < 0) {
+		index = count - 1;
+	}
+	const auto item = accessibilityItemAtIndex(index, elements, barIndex);
 	setAccessibilityFocusedItem(index, item);
 }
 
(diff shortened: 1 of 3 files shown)


john-preston:
**Verdict: CHANGES_REQUIRED.** Complete the reader-mode inbox policy for automatic message-arrival and list-refresh paths. The new gates preserve ordinary and explicitly disabled behavior, but the advertised focus-only behavior still has a common bypass. This is an incomplete feature over inherited automatic-read routes, not a newly introduced mode-off regression.

**P2, confirmed: apply the policy to automatic arrivals and unread-bar refresh too.** With reader mode active, keep an initialized, active chat at the bottom, leave focus in the composer or on an earlier message, and receive a regular incoming unread message. [HistoryWidget::newItemAdded](https://github.com/telegramdesktop/tdesktop/blob/2dede7b22b3e10d570392e82b8a340948e7f01b9/Telegram/SourceFiles/history/history_widget.cpp#L4452) still destroys the unread bar and calls `readInboxOnNewMessage` whenever the ordinary eligibility predicate passes. [That call](https://github.com/telegramdesktop/tdesktop/blob/2dede7b22b3e10d570392e82b8a340948e7f01b9/Telegram/SourceFiles/data/data_histories.cpp#L342) forces `readInboxTill` through the incoming ID. No accessibility-focus change is required, so the arriving message, and any earlier unread messages, can become read despite the new paint/activation gates.

The other implementation has the same bypass: [refreshRows](https://github.com/telegramdesktop/tdesktop/blob/2dede7b22b3e10d570392e82b8a340948e7f01b9/Telegram/SourceFiles/history/view/history_view_list_widget.cpp#L885) dispatches `listItemsAddedToEnd` when already at the newest edge; [ChatWidget's handler](https://github.com/telegramdesktop/tdesktop/blob/2dede7b22b3e10d570392e82b8a340948e7f01b9/Telegram/SourceFiles/history/view/history_view_chat_section.cpp#L5107) reads qualifying incoming messages through the history, replies, or sublist backend. In addition, `markLastAsRead` at line 888 remains based on scroll position and ordinary eligibility. When no bar exists, [listMessagesBar](https://github.com/telegramdesktop/tdesktop/blob/2dede7b22b3e10d570392e82b8a340948e7f01b9/Telegram/SourceFiles/history/view/history_view_chat_section.cpp#L5196) can consume unread rows rather than create their landmark. `_refreshingViewer` is already cleared before `refreshRows`, so it does not close these paths. These routes predate this PR, but leaving them active defeats its central policy in an everyday arrival scenario; this is substantive completeness work, not an unrelated historical omission. Gate automatic inbox advancement and its unread-bar teardown with the same reader-mode policy, while keeping deliberate focus reads eligible.

The existing focused-read path otherwise retains the relevant eligibility conditions. Keyboard navigation is gated by `ScreenReaderModeActive`, computes a concrete non-bar item, and reads only after deliberate movement; [HistoryInner](https://github.com/telegramdesktop/tdesktop/blob/2dede7b22b3e10d570392e82b8a340948e7f01b9/Telegram/SourceFiles/history/history_inner_widget.cpp#L4318) and [ListWidget](https://github.com/telegramdesktop/tdesktop/blob/2dede7b22b3e10d570392e82b8a340948e7f01b9/Telegram/SourceFiles/history/view/history_view_list_widget.cpp#L3762) retain `markingMessagesRead`. Explicit accessibility actions retain that predicate too. Automatic focus selection only assigns/announces; it does not read. The separator maps to a null item, so landing on it cannot send a receipt. Paging, jumping, or explicitly focusing an outgoing row can intentionally read earlier incoming rows cumulatively; that is the stated policy, not a defect.

[HistoryWidget eligibility](https://github.com/telegramdesktop/tdesktop/blob/2dede7b22b3e10d570392e82b8a340948e7f01b9/Telegram/SourceFiles/history/history_widget.cpp#L4763) preserves initialization, load/animation and support-mode restrictions. Both implementations reach [MainWindow::markingAsRead](https://github.com/telegramdesktop/tdesktop/blob/2dede7b22b3e10d570392e82b8a340948e7f01b9/Telegram/SourceFiles/mainwindow.cpp#L595), retaining hidden/minimized/layer/active/idle checks and the existing `AutoScrollInactiveChat` exception. These checks are not lost by suppressing visibility reads.

PR #31081, opened 2026-07-29
fix gap above dialog search results

teidesu:
fixes an empty gap appearing above dialog search results after expanding stories by click.

repro:
1. click the collapsed stories miniatures to expand them
2. focus the search field
3. type a query

pull-down expansion is not affected.

explicit expansion kept its height while stories were hiding for search. reset it before starting the hide animation.

this pr was autherd by gpt-5.6-sol running in codex desktop


john-preston:
Thanks!

PR #31086, opened 2026-07-31
Fix Mini App cookies on Linux

### LINKED ISSUE #30889: Telegram Desktop WebView does not send cookies for same-origin requests after initial page load
### Steps to reproduce

1. Open a HTTPS web page inside Telegram Desktop WebView.
2. The initial HTML response sets a session cookie using `Set-Cookie`.
3. From that loaded page, make a same-origin request to a config/API endpoint, for example `/assets/app-config.json`.
4. Check the request headers received by the backend.

### Expected behaviour

Telegram Desktop WebView should send the cookie that was set by the initial HTML response when the page makes subsequent same-origin requests.

Example:

```text
GET /assets/app-config.json
Cookie: session=...
```

This is the behaviour observed in regular browsers such as Chrome, Firefox and Safari.

### Actual behaviour

The initial page loads correctly and receives `Set-Cookie`, but subsequent same-origin requests from that page never include this cookie in Telegram Desktop WebView.

Example:

```text
GET /assets/app-config.json
Cookie header is missing
```

As a result, endpoints that depend on the session cookie fail. In my case this leads to `502 Bad Gateway` for the follow-up config/assets request inside Telegram Desktop WebView.

The same page works correctly in regular browsers.

As a workaround, I had to stop relying on cookies for this follow-up config request and embed the required config directly into the initial HTML response.

### Operating system

Ubuntu 24.04

6.9.3

### Installation source

### Crash ID

*No response*

### Logs

```shell
No Telegram crash logs. Backend/proxy logs show that the initial HTML request succeeds, but the follow-up same-origin config/assets request is received without the expected `Cookie` header and fails with `502 Bad Gateway`.
```

Suthankan1:
## Summary

- enable third-party cookies for the external Linux Mini App shell
- preserve the default cookie policy for embedded webviews and other platforms
- update `lib_webview` to the opt-in API from desktop-app/lib_webview#146

Fixes #30889.

## Dependency

- https://github.com/desktop-app/lib_webview/pull/146 must merge first

## Testing

- generated and inspected the updated Linux helper D-Bus interface
- compiled, linked, and ran strict cookie API probes against Ubuntu 24.04 WebKitGTK 6.0 and 4.1
- passed XML and whitespace validation

A full Telegram build was not run because this checkout has no configured `out/` tree or local dependency bundle.

### DIFF
--- a/Telegram/SourceFiles/ui/chat/attach/attach_bot_webview.cpp
+++ b/Telegram/SourceFiles/ui/chat/attach/attach_bot_webview.cpp
@@ -2129,6 +2129,7 @@ bool Panel::createWebview(const Webview::ThemeParams &params) {
 		Webview::WindowConfig{
 			.opaqueBg = params.bodyBg,
 			.storageId = _storageId,
+			.allowThirdPartyCookies = _externalShell,
 			.mode = _externalShell
 				? Webview::WindowMode::External
 				: Webview::WindowMode::Embedded,


Suthankan1:
Draft status is intentional and is only for dependency ordering. The code change is complete and the CLA is signed. This PR's gitlink points to commit `2bfe882` from desktop-app/lib_webview#146; until that dependency lands in the official `desktop-app/lib_webview` repository, a clean tdesktop checkout cannot fetch the referenced submodule commit from its configured upstream URL.

I have also reproduced the behavior with a cross-origin HTTPS WebKitGTK 6 fixture: the default policy produced `cookie-present=false`, while the new opt-in produced `cookie-present=true` for the same `HttpOnly; Secure; SameSite=None` cookie. I am preparing the official Linux Docker toolchain for a full Debug compile.

Once lib_webview#146 is merged, I will normalize the gitlink to the upstream merge result if necessary and mark this PR ready for review.

Suthankan1:
Additional Linux Debug compilation validation is complete. Using Telegram Desktop's official `ghcr.io/telegramdesktop/tdesktop/centos_env:latest` x86_64 image, I configured the project as Debug with `CMAKE_COMPILE_WARNING_AS_ERROR=ON` and built the four translation units containing or directly consuming this change:

- `webview/webview_embed.cpp`
- `webview/platform/linux/webview_linux_webkitgtk_library.cpp`
- `webview/platform/linux/webview_linux_webkitgtk.cpp`
- `ui/chat/attach/attach_bot_webview.cpp`

The focused build completed successfully (324/324, exit 0). An immediate repeat reported `ninja: no work to do.` `git diff --check` passes in both repositories. This complements the cross-origin HTTPS behavioral fixture already reported above.

The PR remains intentionally Draft only because its `Telegram/lib_webview` gitlink currently points to commit `2bfe882`, which is not yet reachable from the official `desktop-app/lib_webview` remote. After desktop-app/lib_webview#146 merges, I will update the gitlink if the merge strategy changes the SHA and then mark this PR ready.

john-preston:
This is correct submodule update. Thanks.

Issue #345, opened 2014-12-20
It crash on Mountain Lion (sound issue)


inigosola:
It crash when I received a new message, and I can see it is because play sound.

Telegram Desktop works fine when I disable `Play sound` setting.

telegramdesktop:
Should be fixed in last versions, both from tdesktop.com and from AppStore.

PR #31097, opened 2026-08-02
Expose discrete slider sections as accessible tabs

rezabakhshilaktasaraei:
The tab-like switchers built on Ui::DiscreteSlider / SettingsSlider (search result tabs, stickers box tabs, emoji panel tabs, message options box tabs and so on) were invisible to screen readers: the sections are painted, so there was nothing to focus or read.

The slider now reports itself as a tab control with one virtual tab per section - the UI calls these controls tabs, so the screen reader calls them that too. In screen reader mode the widget becomes a Tab stop: Tab lands on the active tab, and the arrows (mirrored in RTL) and Home/End switch right away to the tab they land on, the way native Windows tab controls behave - the switch is announced through the focus move, so the screen reader reads the tab it landed on. UIA SetFocus / Invoke are dispatched by a stable per-section identity on the main thread.

Not every strip wants that: where switching loads content or can be refused, setAccessibilityActivateOnBrowse(false) keeps a browse-then-commit model - the arrows only move the reading position and announce the tab with its selected state, Enter or Space activates. The search scope tabs (#31184) and the chat folder tabs (#31098) use it. For such subclasses the browse position is exposed as an rpl producer (accessibilitySectionBrowsed), so a scrollable strip can bring the browsed tab into view, activation goes through a protected virtual (activateSectionByAccessibility), so locked premium folders can be intercepted, and a protected accessor (accessibilityBrowsedSection) lets a keyboard-invoked context menu anchor on the browsed tab - a menu asked for from the keyboard arrives with the center of an empty input method rect, so a position based lookup can't find the right one.

Three things outside this repo are needed on Windows, all about what the Qt 5 UIA bridge does and doesn't forward:

- desktop-app/patches#263. Qt 5.15 forwards a focus event to focusChild() only for an accessible with a table interface, so taking focus announced the container and not the tab it points at. Upstream generalized the condition to any element with children in 6.2, and that patch backports it.
- desktop-app/lib_ui#348. The selection interface is what carries the "selected" / "not selected" state a screen reader reads; that PR lets it accept tab items and makes a browsed painted section win over the selected one when the container forwards focus.
- desktop-app/lib_ui#345. The bridge reads only the checked flag out of a state change, so reporting a new selection that way stayed silent. The selection events raised on activation are the ones a screen reader actually hears.

Tested with NVDA on Windows 10 with all three applied: Tab lands on the active tab and it is announced once, and the arrows switch and announce the tab they land on, exactly like the native Windows tab controls. Without a screen reader nothing changes - the widget is not focusable outside screen reader mode.

### DIFF
--- a/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp
+++ b/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp
@@ -91,6 +91,10 @@ rpl::producer<int> DiscreteSlider::accessibilitySectionBrowsed() const {
 	return _accessibilitySectionBrowsed.events();
 }
 
+void DiscreteSlider::setAccessibilityActivateOnBrowse(bool activate) {
+	_accessibilityActivateOnBrowse = activate;
+}
+
 int DiscreteSlider::sectionsCount() const {
 	return int(_sections.size());
 }
@@ -304,12 +308,27 @@ void DiscreteSlider::keyPressEvent(QKeyEvent *e) {
 		browseAndActivate(0);
 	} else if (key == Qt::Key_End && count > 0) {
 		browseAndActivate(count - 1);
+	} else if (!e->isAutoRepeat()
+		&& !_accessibilityActivateOnBrowse
+		&& (key == Qt::Key_Space
+			|| key == Qt::Key_Return
+			|| key == Qt::Key_Enter)
+		&& _accessibilitySelected >= 0
+		&& _accessibilitySelected < count) {
+		// Only an opted-out strip needs the explicit commit step.
+		activateSectionByAccessibility(_accessibilitySelected);
 	} else {
 		RpWidget::keyPressEvent(e);
 	}
 }
 
 void DiscreteSlider::browseAndActivate(int index) {
+	// An opted-out strip (see setAccessibilityActivateOnBrowse) only moves
+	// the browse position, keeping activation on Enter / Space.
+	if (!_accessibilityActivateOnBrowse) {
+		setAccessibilitySelected(index, Announce::Always);
+		return;
+	}
 	// A native Windows tab control switches to the tab the arrows land on
 	// right away, with no separate Enter / Space step, and what the screen
 	// reader announces is the focus moving onto that tab - which by then
--- a/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp
+++ b/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp
@@ -8,6 +8,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 #include "ui/widgets/discrete_sliders.h"
 
 #include "ui/effects/ripple_animation.h"
+#include "ui/screen_reader_mode.h"
 #include "styles/style_widgets.h"
 
 namespace Ui {
@@ -86,6 +87,10 @@ void DiscreteSlider::setAdditionalContentWidthToSection(int index, int w) {
 	}
 }
 
+rpl::producer<int> DiscreteSlider::accessibilitySectionBrowsed() const {
+	return _accessibilitySectionBrowsed.events();
+}
+
 int DiscreteSlider::sectionsCount() const {
 	return int(_sections.size());
 }
@@ -154,6 +159,9 @@ void DiscreteSlider::refresh() {
 	}
 	if (_selected >= _sections.size()) {
 		_selected = 0;
+	}
+	if (_accessibilitySelected >= int(_sections.size())) {
+		_accessibilitySelected = -1;
 	}
 	resizeToWidth(width());
 	update();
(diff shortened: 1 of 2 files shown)


Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp
```diff
@@ -7,10 +7,34 @@ For license and copyright information please follow this link:
...
+// only, like the folders list container, and let focusChild() carry the
+// focused state to the browsed section item.
+class ListAccessible final : public Accessible::Widget {
+public:
+	using Accessible::Widget::Widget;
+
+	QAccessible::State state() const override {
+		auto result = Accessible::Widget::state();
+		if (rp()->accessibilityChildCount() > 0) {
+			result.focused = false;
```
ilya-fedin: That doesn't look right. This means Qt wil report no focused widget when this widget has focus since its internal state will think the widget is in focus thus reporting `focused = true` only for it in the window but then you override it and screen reader will get `focused = false` for all widgets in the window.

Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp
```diff
@@ -7,10 +7,34 @@ For license and copyright information please follow this link:
...
 #include "ui/effects/ripple_animation.h"
+#include "ui/screen_reader_mode.h"
 #include "styles/style_widgets.h"
 
 namespace Ui {
+namespace {
+
+// The slider holds real keyboard focus for its virtual (painted) sections,
+// so Qt would report the container itself as focused. Report it focusable
+// only, like the folders list container, and let focusChild() carry the
```
ilya-fedin: I don't see how it could work? If you don't want the widget to get focus, you either:
1. Set no focus policy
2. Set a focus proxy (QWidget::setFocusProxy)
3. Call setFocus to some other widget on focus event

Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp
rezabakhshilaktasaraei: You were right, and it was worse than it looked: the slider really does hold keyboard focus, so the override was reporting a lie about it. It is gone from the branch.

What it was papering over is a platform gap. On Qt 5.15 the Windows bridge redirects a focus event to focusChild() only for an accessible that exposes a table interface, so focus landed on the container instead of the section focusChild() points at. Upstream generalized that condition to any element with children in 6.2; desktop-app/patches#263 backports it. With that in place the container reports focused, as it should, and the section is what gets announced.

Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp
rezabakhshilaktasaraei: Sorry, the comment above it was misleading. The widget does want the focus: it handles the arrows and is the list's single Tab stop, so none of the three options applies. All I was after was which element the screen reader announces once focus is there - the browsed section rather than the container - and the state override was simply the wrong tool for that. It is gone.

focusChild() already points at the browsed section; what was missing was the platform raising the focus event on it, which desktop-app/patches#263 fixes.

john-preston:
**Verdict: CHANGES_REQUIRED.** Keep the painted-tab approach and manual-activation option, but repair ambiguous action identities, live mode-disable handling, synchronization after mouse activation, and initial AT-SPI child focus. The inherited #345 selection-event defect remains a separate prerequisite correction.

**P2 — Duplicate labels route actions to the wrong section.** [discrete_sliders.cpp:463](https://github.com/telegramdesktop/tdesktop/blob/24a066315c7834ab01c3535e576f753b3b9403b3/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp#L453-L475) derives identity solely from the label hash, and resolution returns its first match. This needs no improbable hash collision: create two chat folders named “Work,” then request SetFocus or Invoke on the second tab. Both carry the same token, so the queued action targets the first folder. The [folder strip passes titles without their filter IDs](https://github.com/telegramdesktop/tdesktop/blob/24a066315c7834ab01c3535e576f753b3b9403b3/Telegram/SourceFiles/ui/widgets/chat_filters_tabs_strip.cpp#L390-L400); the [editor validates length/emptiness and rules, not title uniqueness](https://github.com/telegramdesktop/tdesktop/blob/24a066315c7834ab01c3535e576f753b3b9403b3/Telegram/SourceFiles/boxes/filters/edit_filter_box.cpp#L981-L1008). Plain-text extraction also discards distinctions carried by entities. Item caches are per-index, so do not describe this as two indices necessarily returning the same interface: the confirmed failure is identity-based action routing and stale-identity resolution.

Use unique section tokens, independent of text. The smallest safe generic design can assign fresh nonzero tokens when sections are rebuilt, allowing stale actions to cancel; folders can supply actual filter IDs if preserving identity through reorders is required. Do not use an index alone across rebuilds.

**P2 — Reader-mode deactivation leaves the new key handling active.** [discrete_sliders.cpp:295](https://github.com/telegramdesktop/tdesktop/blob/24a066315c7834ab01c3535e576f753b3b9403b3/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp#L295-L322) has no ScreenReaderModeActive gate. Tab into a slider with a reader active, then stop the reader while that widget retains focus. [FocusManager](https://github.com/desktop-app/lib_ui/blob/8188a76f9ddbc04b4517c6ef0db3f7170062359f/ui/accessible/ui_accessible_widget.cpp#L36-L48) changes its policy to NoFocus, but that prevents future traversal; it does not clear current focus. The configured Qt 5.15.19 `QWidget::setFocusPolicy()` only assigns the policy and propagates it to a focus proxy. Arrows/Home/End therefore continue changing tabs with mode inactive; a manual strip still commits on Enter/Space. This also applies to any disable-preference transition that leaves/restores focus there. It contradicts the claimed isolation outside reader mode.

Delegate these newly introduced keyboard paths when project screen-reader mode is inactive, and handle retained focus consistently. Recheck the same policy in queued focus-taking actions: their plain `setFocus()` can assign focus despite NoFocus. Keep accessibility-tree exposure distinct from permission to introduce screen-reader-only keyboard/focus behavior.

**P2 — Mouse activation leaves arrow navigation on the old tab.** [discrete_sliders.cpp:302](https://github.com/telegramdesktop/tdesktop/blob/24a066315c7834ab01c3535e576f753b3b9403b3/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp#L301-L306) always prefers the stored browse index. On a three-section automatic strip, focus section 0, click section 2, then press Right in LTR. The existing [mouse release/setActiveSection path](https://github.com/telegramdesktop/tdesktop/blob/24a066315c7834ab01c3535e576f753b3b9403b3/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp#L229-L246) updates the active section but never the new browse index. The next Right computes `0 + 1`, selects section 1, and moves backwards from the selected section 2. Until that key, the interface reports section 0 focused and section 2 selected. Synchronize explicit mouse activation with the browse position when this control holds focus, without making an ordinary mouse click acquire focus. Preserve intentionally separate browsing in manual mode.

**P2 — Preserve the child focus event on AT-SPI entry.** [focusInEvent at lines 282-292](https://github.com/telegramdesktop/tdesktop/blob/24a066315c7834ab01c3535e576f753b3b9403b3/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp#L282-L292) selects with Announce::No and relies solely on the widget's native focus event. Tab into the strip, or invoke SetFocus while another widget has focus: the native event names the container. The configured Qt5 bridge and [official Qt6.11.2 AT-SPI bridge](https://github.com/qt/qtbase/blob/v6.11.2/src/gui/accessible/linux/atspiadaptor.cpp) pass that interface to sendFocusChanged, which emits signals for pathForInterface(interface), never focusChild(). QAccessibleEvent resolves only an explicit child index, and neither app nor library installs an update handler to redirect it. No tab-targeted focus signal is sent until a later explicit browse. This event-target omission is source-confirmed, not a runtime speech claim. Use a shared backend-aware focus-entry policy that retains explicit child events where required and suppresses duplicates where forwarding already exists. The #31169 review independently traced the full bridge route; its Windows prerequisite does not fix AT-SPI.

PR #31098, opened 2026-08-03
Make the chat folder tabs and the chat list column accessible

rezabakhshilaktasaraei:
Depends on #31097 and builds on top of it: that PR exposes the painted sections of Ui::DiscreteSlider as accessible tabs, this one completes the chat folder tabs strip specifically. It contains no code from #31097 - it only compiles once that lands, since it overrides a virtual and subscribes to a producer introduced there. With #31097 the strips switch to the tab the arrows land on; the folder tabs keep the browse-then-commit model through the opt-out added there (the last commit here), since switching folders reloads the chat list and the strip may include locked ones - the arrows only browse and announce, Enter or Space commits.

Keyboard or screen reader activation of a locked (premium) folder now behaves like a click on it: the premium promo opens instead of switching to a folder the user cannot use. Folder tab names include the unread count via the existing lng_filter_unread_chats key and locked folders get the lng_sr_folder_locked name and description, mirroring the folders sidebar. Unread count changes fire a name-changed event so a screen reader focused on a tab picks them up, and the strip container is named with lng_filters_title instead of being announced as a bare tab control. Browsing with arrows also scrolls the strip to keep the browsed tab visible, reusing the same scrollToIndex animation the strip already uses when the active filter changes. No new lang keys.

The folder context menu (edit, mark as read, remove) was reachable with a right click only. contextMenuEvent looked up the tab under e->pos(), but a menu asked for from the keyboard comes with the center of an empty input method rect, so the lookup ran past every section and the request was dropped - the menu key did nothing. A keyboard request now takes the tab the browse position is on, or the active one when it is on none, and skips locked folders like the mouse path does. The menu is also anchored on that tab (through Ui::ContextMenuPosition from desktop-app/lib_ui#348) instead of the mouse cursor, which may sit nowhere near the strip; a mouse-invoked menu keeps opening at the cursor.

The third commit orders the chat list column in the Tab chain by position, with setVisualTabOrder. The bars above the list and the folder tabs strip are created only once they are needed, long after the list and the buttons below it, so Tab used to reach the folder tabs after the list instead of before it. It now follows the column: main menu, search, the bars, the folder tabs, the list, jump to top, the buttons at the bottom.

Tested with NVDA on Windows 10 with both PRs applied: browsing the folder tabs reads titles with unread counts and scrolls overflowing tabs into view, Enter activates the browsed folder or opens the premium box on a locked one, the menu key opens the menu of the browsed folder, and Tab walks the column top to bottom.

### DIFF
--- a/Telegram/SourceFiles/dialogs/dialogs_widget.cpp
+++ b/Telegram/SourceFiles/dialogs/dialogs_widget.cpp
@@ -730,6 +730,11 @@ Widget::Widget(
 
 	_search->customUpDown(true);
 
+	// The bars above the list and the folder tabs are created only once they
+	// are needed, long after the list and the buttons below it, so the Tab
+	// chain has to follow the column instead of the creation order.
+	setVisualTabOrder(true);
+
 	updateJumpToDateVisibility(true);
 	updateSearchFromVisibility(true);
 	setupSupportMode();
--- a/Telegram/SourceFiles/ui/widgets/chat_filters_tabs_slider.cpp
+++ b/Telegram/SourceFiles/ui/widgets/chat_filters_tabs_slider.cpp
@@ -7,6 +7,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 */
 #include "ui/widgets/chat_filters_tabs_slider.h"
 
+#include "lang/lang_keys.h"
 #include "ui/effects/ripple_animation.h"
 #include "ui/widgets/side_bar_button.h"
 #include "styles/style_dialogs.h"
@@ -169,15 +170,18 @@ void ChatsFiltersTabs::setUnreadCount(int index, int unreadCount, bool mute) {
 				.muted = mute,
 			});
 			update();
+			accessibilityChildNameChanged(index);
 		}
 	} else if (!unreadCount) {
 		_unreadCounts.erase(it);
 		update();
+		accessibilityChildNameChanged(index);
 	} else if (it->second.count != unreadCount || it->second.muted != mute) {
 		it->second.count = unreadCount;
 		it->second.muted = mute;
 		it->second.cache = cacheUnreadCount(unreadCount, mute);
 		update();
+		accessibilityChildNameChanged(index);
 	}
 	updateSectionsContentWidths();
 }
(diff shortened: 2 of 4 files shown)


john-preston:
**Verdict: CHANGES_REQUIRED.** Keep the implementation, add the small locked-tab selection-state correction below. No target-specific ordinary mouse/keyboard regression was established. Prerequisite defects remain separately tracked in #31097 and #345; this assessment assumes they are corrected.

The strip now opts out of automatic activation: arrows browse and scroll folders, while Enter/Space commits. Locked activation opens the existing Premium prompt. Accessible names include unread counts or the locked-folder explanation, and unread changes emit child-name events. Keyboard context menus use the browsed folder and its rectangle; mouse menus retain their previous cursor position. The chat-list column opts into the existing visual Tab-order helper.

**P2 — Locked folders still claim to support selection.** At [chat_filters_tabs_slider.cpp:495](https://github.com/telegramdesktop/tdesktop/blob/f4bd0dbb9d366b0aa32fbc3dc15f754869cbf499/Telegram/SourceFiles/ui/widgets/chat_filters_tabs_slider.cpp#L491-L499), the new override deliberately opens the Premium prompt instead of activating a locked section. However, the class adds only name/description overrides, leaving [#31097's child state](https://github.com/telegramdesktop/tdesktop/blob/24a066315c7834ab01c3535e576f753b3b9403b3/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp#L409-L424) reporting `selectable = true` for every tab.

Reproduce with a non-Premium account that has a locked folder, then request SelectionItem.Select on that folder through an accessibility client. The [integrated selection interface](https://github.com/desktop-app/lib_ui/blob/8188a76f9ddbc04b4517c6ef0db3f7170062359f/ui/accessible/ui_accessible_widget.cpp#L340-L355) accepts the advertised state, dispatches pressAction and returns true. The [painted item action](https://github.com/desktop-app/lib_ui/blob/8188a76f9ddbc04b4517c6ef0db3f7170062359f/ui/accessible/ui_accessible_item.cpp#L220-L225) reaches this new locked override. The existing [Qt 5 selection backport](https://github.com/desktop-app/patches/blob/f5c456284a0ae92c01ce5e7cb82ea37b15ed0827/qtbase_5.15.19/0036-backport-uia-selection-interface.patch#L140-L144) converts true to S_OK. The client is told selection succeeded, yet the selected folder remains unchanged and a Premium prompt appears. This is a source-confirmed contract mismatch, not a claim that runtime speech was observed.

Add `ChatsFiltersTabs::accessibilityChildState(int)`, start with the base state, and clear `selectable` for a locked index. Preserve focusability, browsing and the press action so users can still hear why the folder is unavailable and intentionally open subscription options. Do not mark the whole tab disabled. The existing sidebar makes the same distinction, but the selection call path is the reason for the correction.

PR #31103, opened 2026-08-03
Fix reply keyboards in bot forum topics

ahamkara-Nh:
## Why

Telegram Desktop kept one reply keyboard per history. Bot forum topics share a history, so the All feed showed a topic keyboard while individual threads showed none at all. Mobile clients correctly showed the matching reply keyboard in each topic.

## Summary

- keep reply keyboard state isolated per bot forum topic, including cold-load reconstruction and live updates
- preserve keyboard state across temporary-to-real topic IDs and edit-mode transitions
- route keyboard and ForceReply sends to the source message
- restore the existing reply-header and cancellation behavior for ForceReply
- hide bot-command controls while reply keyboards or ForceReply are active
- clear topic keyboards when their bot leaves the chat

## Steps to reproduce
Open a private chat with a bot that has threaded mode enabled.
Open a topic where the bot sends ReplyKeyboardMarkup.
Switch between All and that topic.
Have the bot replace the keyboard, or create a new topic whose first reply contains a keyboard.
Before this change, All displayed a topic keyboard of the latest topic, topic views displayed none

## Validation

- Windows x64 Debug build with MSVC 14.44
- fresh-login test in an isolated workdir
- manual coverage: All tab, multiple existing and new topics, live replacement, relaunch reconstruction, keyboard-button send, ForceReply reply/cancel, and edit-mode restore

## Related

- [Custom Keyboard not visible in Topic / Forum chats in Telegram Desktop](https://bugs.telegram.org/c/47099)

## After
1 after
2 after

### DIFF
--- a/Telegram/SourceFiles/api/api_chat_participants.cpp
+++ b/Telegram/SourceFiles/api/api_chat_participants.cpp
@@ -192,7 +192,7 @@ void ApplyBotsList(
 
 	auto needBotsInfos = false;
 	auto botStatus = channel->mgInfo->botStatus;
-	auto keyboardBotFound = !history || !history->lastKeyboardFrom;
+	auto present = base::flat_set<PeerId>();
 	for (const auto &p : list) {
 		const auto participant = channel->owner().peer(p.id());
 		const auto user = participant->asUser();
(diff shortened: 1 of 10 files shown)


23rd:
Need to be rebased on the latest tag/branch.

ahamkara-Nh:
Closing this PR: the issue is already fixed upstream by the modern chat view and bot keyboard rewrite (landed in 7.1.0, e.g. [chat] Migrate bot keyboard and force-reply, d6b543c004), where each forum topic is a per-thread section with its own keyboard state. This PR targets the pre-rewrite code and would be obsolete on the latest branch. Thanks for the work!

Issue #348, opened 2014-12-22
Adding Persian language to Telegram


erajabi:
Hi, 

As you know, there are a lot of Persian users for Telegram. How can I contribute to add Persian language (Farsi) to this nice software?

Thanks,

PR #31110, opened 2026-08-04
Make topic tabs accessible to screen readers

rezabakhshilaktasaraei:
﻿The topic tabs inside a forum chat (and the sender tabs in a channel's Direct Messages inbox) were invisible to screen readers: the tab buttons had no names, no list semantics and no keyboard access, so a blind user could not tell which topics exist, which one is active, or switch between them.

The tab strip is now a tab control - it is named tabs everywhere in the UI, so the screen reader calls it that too. Each tab is a named item (title with unread count, mention state and the pinned mark, updated as the badges change), the active one is reported as selected, and the strip is named "Topics" or "Direct Messages" with the right orientation for each position. In screen reader mode the strip gets one roving Tab stop on the active tab; arrows move focus between tabs (Up/Down when vertical, Left/Right when horizontal, Home/End for the edges) without switching - the browse-then-commit pattern for tabs whose panes load their content - Enter or Space activates, the context menu key opens the topic menu anchored on the tab (a mouse-invoked menu keeps opening at the cursor), and the focused tab is scrolled into view. Keyboard focus survives the frequent rebuilds of the strip on unread updates. The folders sidebar follows in #31176.

The position toggle button got an accessible name, with the current position reported in its description. It is also one persistent widget now, reused across mode switches instead of being recreated with the strip, so pressing it from the keyboard keeps focus on it without the screen reader re-announcing the button on every press. Sighted users are unaffected - the button is only focusable in screen reader mode.

The strip is created only once it is needed, long after the top bar, the list and the composer, so its Tab stops used to land at the end of the chain. The chat sections now keep their Tab chain in visual order (the lib_ui visual Tab order added for the chat list column), which places the strip at its reading-order position wherever it sits - left column, top or bottom - and follows it as the toggle moves it around. The one thing the container cannot see on its own is the roving stop moving between tabs deep inside the strip, so the slider tells it after each move; without that, a Tab entering the strip from outside would follow wiring to the previous stop. The composer inside a topic needed the same treatment: its buttons are created in an order unrelated to the row layout, so switching topics used to scramble their Tab order, and its send-as picker was missing its accessible name.

Needs desktop-app/lib_ui#348 for the PageTab button role.

Tested with NVDA on Windows 10 in a forum group with the tabs layout: browsing, activation, unread counts, context menus and cycling the tab bar through all three positions - after each move the next Tab finds the strip at its new place, and leaving the strip and Tabbing back in lands on the active tab - the arrows browse freely inside, but once focus leaves, the roving stop parks back on the current topic.

### DIFF
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -8581,6 +8581,10 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 "lng_sr_bot_verified_badge" = "Verified Bot";
 "lng_sr_profile_menu" = "Profile menu";
 "lng_sr_close_panel" = "Close panel";
+"lng_sr_tabs_position" = "Change tab bar position";
+"lng_sr_tabs_position_top" = "Tab bar at the top";
+"lng_sr_tabs_position_bottom" = "Tab bar at the bottom";
+"lng_sr_tabs_position_left" = "Tab bar on the left";
 
 "lng_ai_compose_title" = "AI Editor";
 "lng_ai_compose_apply" = "Apply";
--- a/Telegram/SourceFiles/history/history_widget.cpp
+++ b/Telegram/SourceFiles/history/history_widget.cpp
@@ -1182,6 +1182,13 @@ HistoryWidget::HistoryWidget(
 	orderWidgets();
 	setupShortcuts();
 
+	// The topic tabs strip and the bars above the list are created only
+	// once they are needed, long after the top bar, the list and the
+	// composer, so the Tab chain has to follow the layout instead of the
+	// creation order - including the strip's moves between the left
+	// column, the top and the bottom.
+	setVisualTabOrder(true);
+
 	_attachToggle->setAccessibleName(tr::lng_attach(tr::now));
 	_tabbedSelectorToggle->setAccessibleName(tr::lng_emoji_sticker_gif(tr::now));
 	_botKeyboardShow->setAccessibleName(tr::lng_bot_keyboard_show(tr::now));
--- a/Telegram/SourceFiles/history/view/controls/history_view_compose_controls.cpp
+++ b/Telegram/SourceFiles/history/view/controls/history_view_compose_controls.cpp
@@ -1306,6 +1306,10 @@ ComposeControls::ComposeControls(
 , _unavailableEmojiPasted(std::move(descriptor.unavailableEmojiPasted))
 , _saveDraftTimer([=] { saveDraft(); })
 , _saveCloudDraftTimer([=] { saveCloudDraft(); }) {
+	// The buttons are created in an order unrelated to where they sit in
+	// the row (the send button first of all), so the Tab chain has to
+	// follow the layout instead of the creation order.
+	_wrap->setVisualTabOrder(true);
 	if (_st.radius > 0) {
 		_backgroundRect.emplace(_st.radius, _st.bg);
 	}
@@ -5510,6 +5514,7 @@ bool ComposeControls::updateSendAsButton(
 	}
 	const auto &st = _st.chooseSendAs;
 	_sendAs = std::make_unique<Ui::SendAsButton>(_wrap.get(), st.button);
+	_sendAs->setAccessibleName(tr::lng_send_as_title(tr::now));
 	if (videoStream) {
 		Ui::SetupSendAsButton(_sendAs.get(), st, videoStream, _show);
 		_videoStreamAdmin = videoStream->creator();
--- a/Telegram/SourceFiles/history/view/history_view_chat_section.cpp
+++ b/Telegram/SourceFiles/history/view/history_view_chat_section.cpp
@@ -1063,6 +1063,12 @@ ChatWidget::ChatWidget(
 
 	orderWidgets();
 
+	// The topic tabs strip is created only once it is needed, long after
+	// the top bar, the list and the composer, so the Tab chain has to
+	// follow the layout instead of the creation order - including the
+	// strip's moves between the left column, the top and the bottom.
+	setVisualTabOrder(true);
+
 	updateControlsVisibility();
 
 	refreshSuggestPostToggle();
--- a/Telegram/SourceFiles/history/view/history_view_subsection_tabs.cpp
+++ b/Telegram/SourceFiles/history/view/history_view_subsection_tabs.cpp
@@ -342,9 +342,12 @@ void SubsectionTabs::setupSlider(
 		return _reordering > 0;
 	});
 
-	slider->sectionContextMenu() | rpl::on_next([=](int index) {
-		if (index >= 0 && index < _slice.size()) {
-			showThreadContextMenu(_slice[index].thread);
+	slider->sectionContextMenu() | rpl::on_next([=](
+			Ui::SubsectionTabMenuRequest request) {
+		if (request.index >= 0 && request.index < _slice.size()) {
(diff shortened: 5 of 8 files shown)


john-preston:
**Verdict: CHANGES_REQUIRED.** The PageTab design is appropriate, but logical focus is not preserved across normal refreshes, and focus-driven scrolling can synchronously destroy the button handling the focus event. Live reader-mode deactivation also leaves new navigation active, and some backends drop its selection notification. These are source conclusions; no crash or screen-reader behavior was reproduced at runtime.

**P1 — Defer focus-driven scrolling until focus dispatch has completed.** The new [focus handler](https://github.com/telegramdesktop/tdesktop/blob/3b88e138b8be023426990f82c548995e6aec0690/Telegram/SourceFiles/ui/controls/subsection_tabs_slider.cpp#L554-L556) calls the delegate and then the base handler on the same object. Use a vertical forum strip with more already-loaded topics than the rendered slice, focus near its beginning, and press End. Revealing the last rendered tab can delete that very button before the base handler runs. This is an exact synchronous path:

- [buttonFocused](https://github.com/telegramdesktop/tdesktop/blob/3b88e138b8be023426990f82c548995e6aec0690/Telegram/SourceFiles/ui/controls/subsection_tabs_slider.cpp#L898-L907) emits requestShown; its [owner callback](https://github.com/telegramdesktop/tdesktop/blob/3b88e138b8be023426990f82c548995e6aec0690/Telegram/SourceFiles/history/view/history_view_subsection_tabs.cpp#L354-L379) calls scrollToY.
- lib_ui #348's [scrollToY](https://github.com/desktop-app/lib_ui/blob/8188a76f9ddbc04b4517c6ef0db3f7170062359f/ui/widgets/scroll_area.cpp#L1115-L1117) sets the scrollbar value. The direct [valueChanged connection](https://github.com/desktop-app/lib_ui/blob/8188a76f9ddbc04b4517c6ef0db3f7170062359f/ui/widgets/scroll_area.cpp#L240-L243) calls scrolled(), which [fires scrolls](https://github.com/desktop-app/lib_ui/blob/8188a76f9ddbc04b4517c6ef0db3f7170062359f/ui/widgets/scroll_area.cpp#L658-L660).
- At the bottom boundary with `_afterAvailable > 0`, [startScrollChecking](https://github.com/telegramdesktop/tdesktop/blob/3b88e138b8be023426990f82c548995e6aec0690/Telegram/SourceFiles/history/view/history_view_subsection_tabs.cpp#L408-L420) calls refreshAroundMiddle directly. That [changes the center and calls refreshSlice](https://github.com/telegramdesktop/tdesktop/blob/3b88e138b8be023426990f82c548995e6aec0690/Telegram/SourceFiles/history/view/history_view_subsection_tabs.cpp#L948-L951); the changed slice [fires _refreshed immediately](https://github.com/telegramdesktop/tdesktop/blob/3b88e138b8be023426990f82c548995e6aec0690/Telegram/SourceFiles/history/view/history_view_subsection_tabs.cpp#L966-L972).
- The filling subscriber creates fresh [vertical thumbnails](https://github.com/telegramdesktop/tdesktop/blob/3b88e138b8be023426990f82c548995e6aec0690/Telegram/SourceFiles/history/view/history_view_subsection_tabs.cpp#L481-L505); [MakeEmojiThumbnail](https://github.com/telegramdesktop/tdesktop/blob/3b88e138b8be023426990f82c548995e6aec0690/Telegram/SourceFiles/ui/dynamic_thumbnails.cpp#L1107-L1118) always allocates a new shared object. Thus [setSections](https://github.com/telegramdesktop/tdesktop/blob/3b88e138b8be023426990f82c548995e6aec0690/Telegram/SourceFiles/ui/controls/subsection_tabs_slider.cpp#L635-L655) cannot match the old buttons by thumbnail pointer. Its local vector owns the old focused button and destroys it on [function exit](https://github.com/telegramdesktop/tdesktop/blob/3b88e138b8be023426990f82c548995e6aec0690/Telegram/SourceFiles/ui/controls/subsection_tabs_slider.cpp#L700-L714).

The original focusInEvent then resumes and calls the base on freed storage. The configured Qt base focus handler reads the widget and updates it. The pinned lib_rpl revision `c57cccffb01d85570decd7fccb88419c9a682e63` invokes event handlers synchronously; this chain contains no scheduling boundary. Queue the reveal through an existing guarded queued helper, revalidate the focused logical tab when it executes, and avoid destroying a receiver during Qt focus dispatch. A weak check after the delegate alone does not protect Qt's surrounding dispatch.

**P2 — Preserve the focused thread, not the recycled QWidget.** The new [hadFocus/restore logic](https://github.com/telegramdesktop/tdesktop/blob/3b88e138b8be023426990f82c548995e6aec0690/Telegram/SourceFiles/ui/controls/subsection_tabs_slider.cpp#L631-L635) and [fallback](https://github.com/telegramdesktop/tdesktop/blob/3b88e138b8be023426990f82c548995e6aec0690/Telegram/SourceFiles/ui/controls/subsection_tabs_slider.cpp#L702-L712) remember only whether any widget had focus. Horizontal entries have no userpic, so the existing reuse loop consumes old.back(): unchanged topics `[All, A, B, C]` receive widgets `[oldC, oldB, oldA, oldAll]`. Focus on A now belongs to B. An unread badge change triggers [scheduleRefresh](https://github.com/telegramdesktop/tdesktop/blob/3b88e138b8be023426990f82c548995e6aec0690/Telegram/SourceFiles/history/view/history_view_subsection_tabs.cpp#L869-L880), rebuilds the same logical order, and the surviving widget makes the fallback skip restoration. Its [new click callback](https://github.com/telegramdesktop/tdesktop/blob/3b88e138b8be023426990f82c548995e6aec0690/Telegram/SourceFiles/ui/controls/subsection_tabs_slider.cpp#L668-L679) activates B's index, and the [owner opens that thread](https://github.com/telegramdesktop/tdesktop/blob/3b88e138b8be023426990f82c548995e6aec0690/Telegram/SourceFiles/history/view/history_view_subsection_tabs.cpp#L303-L318) on Enter.

Vertical topic updates recreate every thumbnail and instead move focus back to the active index, losing the independently browsed topic. Group both cases under one repair: capture and restore a stable thread/topic identity across the rebuild, after names and active state are installed, and use a deliberate fallback only if that thread disappeared. Reusing widgets by that identity is another option. Titles and indices are not identities; duplicate titles and reordering are valid.

**P2 — Supply selection notifications on backends that ignore this state mask.** [activeChangedForAccessibility](https://github.com/telegramdesktop/tdesktop/blob/3b88e138b8be023426990f82c548995e6aec0690/Telegram/SourceFiles/ui/controls/subsection_tabs_slider.cpp#L743-L758) emits only selected StateChanged events. [Upstream Qt 6.11.2 Windows](https://github.com/qt/qtbase/blob/v6.11.2/src/plugins/platforms/windows/uiautomation/qwindowsuiamainprovider.cpp#L75-L115) does not translate that bit; the [configured Qt 5 AT-SPI handler](C:/Telegram/Libraries/qt_5.15.19/qtbase/src/platformsupport/linuxaccessibility/atspiadaptor.cpp:1069) also ignores it and requires explicit selection events. Thus activation changes the queryable selection but sends no selection-change notification through those paths; clients tracking selection events are not informed. No speech failure was reproduced. Configured Qt 5 Windows already has the project's 0038 selected-state patch, so it is not affected in the same way. Use a shared library/bridge notification strategy that covers these backends, preserves that Qt 5 coverage, and avoids duplicate or false Add events. #345's unconditional helper is related context, not a ready replacement or compile prerequisite.

PR #31111, opened 2026-08-04
Name the tab strips for screen readers

rezabakhshilaktasaraei:
Complements #31097: that PR turns the painted sections of Ui::DiscreteSlider into accessible tabs, but the containers themselves had no accessible names, so a screen reader landing on them announced a bare tab control. This PR names them at every call site - the chat list search tabs, the emoji / stickers / GIFs panel, the stickers box, the notifications count and proxy auto-switch sliders, the Stars transaction history, the channel earn history, the boosts list and the message options box.

Almost all names reuse existing keys near each slider (the subsection title or box context), so only one new key is added (lng_sr_message_options for the link / reply / forward options box, which has no single visible title when the tabs are shown). The folder tabs strip is not touched here - it is named in #31098.

Builds and runs fine without #31097 (an accessible name on a widget without an accessible role is simply unused), so there is no compile dependency, but the names only become audible together with it.

### DIFF
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -8459,6 +8459,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 "lng_sr_bot_verified_badge" = "Verified Bot";
 "lng_sr_profile_menu" = "Profile menu";
 "lng_sr_close_panel" = "Close panel";
+"lng_sr_message_options" = "Message options";
 
 "lng_ai_compose_title" = "AI Editor";
 "lng_ai_compose_apply" = "Apply";
--- a/Telegram/SourceFiles/boxes/connection_box.cpp
+++ b/Telegram/SourceFiles/boxes/connection_box.cpp
@@ -1181,6 +1181,8 @@ void ProxiesBox::setupContent() {
 			_proxyRotationOptions->entity(),
 			st::settingsSlider),
 		st::settingsBigScalePadding);
+	_proxyRotationTimeout->setAccessibleName(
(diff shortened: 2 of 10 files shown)


john-preston:
**Verdict: CHANGES_REQUIRED — low-priority scope completion only.** The nine added names and new language key have no identified source-level safety defect. One existing tab strip is omitted from the PR's explicit promise to name “every call site.” This is a pre-existing gap left incomplete, not a regression caused by the additions.

**P3 — Complete the bot transaction-history naming call site.** [info_bot_earn_list.cpp:315–322](https://github.com/telegramdesktop/tdesktop/blob/e92c5d314357b93a49b0c237cffe9af386d7e0e6/Telegram/SourceFiles/info/bot/earn/info_bot_earn_list.cpp#L315-L322) creates another `CustomWidthSlider`, but this file has no accessible-name assignment. Open a bot's earnings history with a nonempty history and incoming or outgoing entries: its All/Incoming/Outgoing strip is shown and, with #31097, exposed as tabs, yet it still lacks the container context supplied to the corresponding personal and channel histories. The children remain individually labeled; this is a context/completeness issue, not loss of navigation. Add `slider->entity()->setAccessibleName(tr::lng_channel_earn_history_title(tr::now))` immediately after construction, matching [settings_credits.cpp:316](https://github.com/telegramdesktop/tdesktop/blob/e92c5d314357b93a49b0c237cffe9af386d7e0e6/Telegram/SourceFiles/settings/sections/settings_credits.cpp#L316-L317). The description expressly promises every call site and excludes only the folder strip, handled by #31098; that makes this a small completion request rather than an unrelated backlog item.

PR #30798, opened 2026-06-10
Throttle concurrent proxy checks

somiljain2006:
Fixes #30776 

Introduces an application-level queue in `ProxiesBoxController` to limit the number of concurrent proxy availability checks.

### Changes

* Added `_pendingCheckQueue` to defer proxy checks instead of starting them immediately.
* Limited concurrent checks with `kMaxConcurrentChecks = 50`.
* Duplicate queue entries are ignored when dispatched if the proxy is already being checked.
* Uses lazy deletion: deleted items remain in the queue and are discarded when popped, avoiding expensive queue modifications during bulk deletion.

### Testing (macOS)

While I could not reproduce the original Wayland-specific crash on macOS, I tested the behavior with approximately 3000 imported proxies.

**Unmodified build**

* Application became noticeably unresponsive.
* Open file descriptors reached approximately 218 during testing.

**Patched build**

* Open file descriptors remained bounded (observed peak ~188, settling around ~153 as checks completed).
* Application remained responsive while processing the proxy list.
* Successfully handled bulk imports of approximately 3000 proxies without crashes.

**Additional stress test**

* Launched Telegram under `ulimit -n 128`.
* Imported approximately 3000 proxies.
* Application remained stable, and file descriptor usage stayed below the imposed limit.

The current implementation appears to start proxy checks immediately for every imported proxy. By limiting the number of simultaneous checks and queueing the remainder, resource usage remains bounded even when very large proxy lists are imported.

PR #31112, opened 2026-08-04
Throttle concurrent proxy checks

### LINKED ISSUE #30776: Too many proxies result in crash with "Too many open files"
### Steps to reproduce

**NB**: tho the crash is being reported with distro-binary, I **did** built myself 6.8.2 with debug symbols to get stacktrace for the report. However, launching this debug version results in unauthorized TG with `0` proxies; and TG doesn't allow to export proxies list, so I kind of stuck here. If you need me to reproduce the crash with this newly built version, I kindly ask you to explain "how".

--------

A country I'm in blocks IP addresses of proxies, so I ended up using huge lists [like this one](https://raw.githubusercontent.com/Argh94/Proxy-List/refs/heads/main/MTProto.txt) and adding them all at once. I however do not remove older proxies, because they sometimes register new IP addresses so they start working again, at which I point I switch.

So, I think at this point I gathered around a thousand proxies, if not more *(TG has no exact counter)*.

Now, adding more proxies via a list [like this one](https://raw.githubusercontent.com/Argh94/Proxy-List/refs/heads/main/MTProto.txt) results in TG crashing with:

```
dell-g15 Telegram[1769910]: error marshalling arguments for create_pool: dup failed: Too many open files
dell-g15 Telegram[1769910]: Error marshalling request for wl_shm.create_pool: Too many open files
dell-g15 Telegram[1769910]: The Wayland connection experienced a fatal error: Too many open files
````

Moreover, sometimes it even crashes on start if I just click the menu with proxies list.

**Steps:**

1. Copy a list with ≈2000 unique proxy addresses
2. Open "Proxies" menu in TG, and use "Add proxy from clipboard" to add the 2000 proxies
3. *(if it didn't crash yet)* exit TG, launch it again, and redo steps 1 and 2 with another 1000 unique proxies *(at this point it certainly should crash)*

### Expected behaviour

The list is added and TG keeps working

### Actual behaviour

Telegram crashes with:

```
dell-g15 Telegram[1769910]: error marshalling arguments for create_pool: dup failed: Too many open files
dell-g15 Telegram[1769910]: Error marshalling request for wl_shm.create_pool: Too many open files
dell-g15 Telegram[1769910]: The Wayland connection experienced a fatal error: Too many open files
```

### Operating system

Archlinux

6.8.2

### Installation source

Other (unofficial) source

### Crash ID

_No response_

### Logs

```shell

```

Suthankan1:
Fixes #30776.

## Problem

Importing a very large proxy list can start availability checks for every proxy at once. Each logical check may create IPv4 and IPv6 connections, so thousands of proxies can exhaust the process file-descriptor limit. On Wayland this can also break the compositor connection and terminate Telegram Desktop.

## Solution

- Limit proxy availability checks to 50 logical proxies at a time.
- Place additional checks in a FIFO queue and start the next checks whenever capacity becomes available.
- Lazily discard queue entries for proxies that were deleted, replaced, or no longer need checking.
- Validate callback connection pointers before applying results, preventing callbacks from superseded checks from changing current proxy state.
- Guard queue processing against synchronous callback re-entry.

Queued proxies remain in the existing `Checking` state. Successful or final failed checks update the proxy normally and immediately refill the available check slots.

The limit applies to logical proxy checks rather than individual IPv4/IPv6 connections, preserving the existing dual-stack checking behavior.

## Safety and compatibility

- Deleting or replacing a queued proxy is safe.
- Stale callbacks cannot overwrite a newer check result.
- A synchronous connection failure cannot recursively corrupt queue processing.
- No storage format, networking protocol, or public API is changed.

## Relation to #30798

This addresses the same underlying resource-exhaustion problem as the earlier closed PR, while also covering queued-entry invalidation, superseded callbacks, and re-entrant completion handling.

## Testing

- `git diff --check`
- GitHub Actions Linux build passed
- GitHub Actions macOS and packaged macOS builds passed
- GitHub Actions Snap Ubuntu build passed
- GitHub Actions Windows x64/x86/ARM64 matrix passed

### DIFF
--- a/Telegram/SourceFiles/boxes/connection_box.cpp
+++ b/Telegram/SourceFiles/boxes/connection_box.cpp
@@ -67,6 +67,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 namespace {
 
 constexpr auto kSaveSettingsDelayedTimeout = crl::time(1000);
+constexpr auto kMaxConcurrentProxyChecks = 50;
 
 using ProxyData = MTP::ProxyData;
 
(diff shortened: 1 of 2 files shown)


Suthankan1:
Reviewer walkthrough:

1. `refreshChecker()` resets the previous check and queues the proxy ID.
2. `processCheckerQueue()` starts queued checks while fewer than 50 logical proxy checks are active.
3. Queue entries are resolved by ID when processed, so deleted or superseded entries are skipped safely.
4. Completion callbacks verify that their connection is still the proxy's current IPv4 or IPv6 checker before changing state.
5. Each completed logical check drains the queue, keeping the limit filled without starting an unbounded number of sockets.

This specifically protects the large imported-proxy-list case from file-descriptor exhaustion while preserving the existing proxy status and dual-stack behavior.

CI status at the time of this update: all required checks have passed, including Rocky Linux 8, both macOS workflows, Snap Ubuntu, the full Windows x64/x86/ARM64 matrix, and CLA.

PR #31118, opened 2026-08-06
Fix poll close timer day count

### LINKED ISSUE #30955: Weird display of poll ttl in days
### Steps to reproduce

1. create a poll with a scheduled end time, in a different day (eg. tomorrow)
2. look at the "ttl time" tdesktop displays ("results in ...  days")

### Expected behaviour

If results are expected to be shown tomorrow, tdesktop should say "results in 1 day" or "results tomorrow"

### Actual behaviour

tdesktop has a weird way to calculate ttl:

Image

As you can see from the above screenshot, the current datetime is 10th of July, 09:09. The poll is scheduled to end at 13:30 of **tomorrow** 11th of July. The "_results in ...  days_" says "_2 days_", which is confusing, because if I read "_2 days_", I expect the poll not to end tomorrow, but tomorrow's following day (12th of July, today + 2 days)

If you think about it, when the poll ttl is >24 hours, tdesktop will show "results in 2 days". the moment the poll ttl goes below the 24 hours threshold, we jump from "_results in 2 days_" to "_ends in hh:mm:ss_". I think it is a little confusing to simply calc the number of days as `round_up(ttl_hours / 24)` (simplified)

Similarly, the same way of calculating days is used even if the poll ends not tomorrow, but later. Example: now (10 july, 09:00) I create a poll that closes 12 july, 13:30. Tdesktop shows "results in 3 days", even if the closing time is _today + 2 days_

### Operating system

Windows 11

6.9.4

### Installation source

### Crash ID

_No response_

### Logs

```shell

```

huangtina22:
## Summary
- Poll footers used ceiling division for remaining days (`(left + 86399) / 86400`), so any close time more than exactly 24 hours away showed at least "2 days".
- Switch to full days (`left / 86400`), matching other "in N days" UI (e.g. business hours) and the expected "ends tomorrow → 1 day" reading.
- Below 24 hours the existing `hh:mm:ss` path is unchanged.

Fixes #30955

## Test plan
- [ ] Create a poll that closes tomorrow afternoon; footer should say "1 day" (or "results in 1 day"), not "2 days"
- [ ] Create a poll that closes in ~30 hours; still "1 day"
- [ ] Create a poll that closes in ~50 hours; "2 days"
- [ ] Create a poll that closes in under 24 hours; still shows the countdown timer

### DIFF
--- a/Telegram/SourceFiles/history/view/media/history_view_poll.cpp
+++ b/Telegram/SourceFiles/history/view/media/history_view_poll.cpp
@@ -4495,7 +4495,7 @@ QString Poll::Footer::closeTimerText() const {
 	const auto hideResults = (_owner->_flags
 		& PollData::Flag::HideResultsUntilClose);
 	if (left >= 86400) {
-		const auto days = (left + 86399) / 86400;
+		const auto days = left / 86400;
 		return hideResults
 			? tr::lng_polls_results_in_days(tr::now, lt_count, days)
 			: tr::lng_polls_ends_in_days(tr::now, lt_count, days);


23rd:
Great, thanks.

huangtina22:
Thanks.

PR #31123, opened 2026-08-07
Name the hint dismiss buttons for screen readers

rezabakhshilaktasaraei:
The floating hints (the stories onboarding hint over the chats list and the AI compose hint) have a close button that permanently dismisses them. With a screen reader it was announced as just button, with no way to tell what it does.

This adds an lng_tooltip_dismiss (Dismiss) key and passes it as the accessible name of both hint close buttons.

Depends on desktop-app/lib_ui#335.


rezabakhshilaktasaraei:
Not needed anymore - lib_ui names the tooltip close button itself now via Integration::phraseButtonClose (desktop-app/lib_ui#335), so no new string or call-site changes are required here.

PR #31124, opened 2026-08-07
Order the chat controls and folders sidebar visually in the Tab chain

rezabakhshilaktasaraei:
Uses the new setVisualTabOrder from desktop-app/lib_ui#336 in two places:

- HistoryWidget: one line in the constructor replaces manual setTabOrder wiring for the write row. Several of its controls are created on demand, so creation order scattered them over the focus chain; now Tab follows the visual layout (top bar, pinned bar, message list, then attach - field - send row left to right), including buttons that appear later like silent broadcast or TTL.

- Folders sidebar: the manual wiring in setListTabStop is replaced by the visual order of the two containers, with refreshVisualTabOrder() called when the roving Tab-stop moves to another folder (arrows or activation change it without any Tab passing through the sidebar, which used to leave the chain pointing at the demoted button).

Verified with NVDA: composer order in all chat types, list position surviving chat switches, folders reachable from both Tab directions after reordering, and Tab still trapped inside boxes.

Depends on desktop-app/lib_ui#336.

### DIFF
--- a/Telegram/SourceFiles/dialogs/dialogs_widget.cpp
+++ b/Telegram/SourceFiles/dialogs/dialogs_widget.cpp
@@ -1192,8 +1192,10 @@ void Widget::scrollToDefaultChecked(bool verytop) {
 
 void Widget::setupScrollUpButton() {
 	// The button floats over the bottom of the list, but it is created long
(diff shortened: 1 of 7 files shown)


john-preston:
Thanks!

PR #31125, opened 2026-08-07
Loop videos with sound in the media viewer

### LINKED ISSUE #17401: [Feature Request] Video looping in the Media player

Telegram for Android loops the videos by default.

TelegramDesktop doesn't seem to be able to loop them at all.

Loop the videos in the media viewer (either always or by an option).

Well, open a video in an external player that supports looping, or hit the play button manually every time? ;)

### Additional context

C'mon, am I really first with this FR?

enomado:
Closes #17401.

Videos without sound (GIFs, silent videos) are already looped natively by the streaming player through `options.loop`. Videos with sound stop on the last frame, because `options.loop` is not allowed in `Mode::Both` — there is an `Expects()` for exactly that in `Streaming::Player::play`. This adds an option that loops them by restarting the player at position 0 when the stream finishes.

That matches Telegram for Android, where short videos loop by default and longer ones loop when the option is on (see the thread in #17401).

The option is off by default and is toggled from two places:

* the media viewer context menu — "Loop Video" / "Don't Loop Video", shown only for videos the option actually affects: not for silent videos and GIFs that already loop, not for stories, and not without a known duration;
* Settings → Advanced → Performance.

Implementation notes:

* The value is stored through the generic KV prefs facility (`writePref` / `readPref`), so nothing is appended to the `Core::Settings` binary stream and there is no serialization ordering to preserve.
* `_streamingStartPaused` is cleared before the restart, the same way `playbackPauseResume()` clears it before its own. The flag survives from the last restart that was made while paused — a seek or a frame step — and without the reset the loop would stop paused on the first frame.
* The menu icon reuses the existing `player/player_repeat` asset, no new assets are added.
* Built and tested on Linux, Debug configuration, through the Docker `centos_env` environment.

I am aware of the "no new features, no new user interface elements" rule in CONTRIBUTING.md. This is a five-year-old feature request, so the implementation is attached as a starting point rather than as a finished proposal — happy to drop the settings toggle and keep only the context menu entry, to make looping unconditional with no option at all, or to change the wording, whatever fits the product decision.

### DIFF
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -887,6 +887,9 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 "lng_settings_system_integration" = "System integration";
 "lng_settings_performance" = "Performance";
 "lng_settings_enable_hwaccel" = "Hardware accelerated video decoding";
+"lng_settings_loop_video_in_viewer" = "Loop videos in the media viewer";
+"lng_mediaview_loop_video_on" = "Loop Video";
+"lng_mediaview_loop_video_off" = "Don't Loop Video";
 "lng_settings_enable_opengl" = "Enable OpenGL rendering for media";
 "lng_settings_angle_backend" = "ANGLE graphics backend";
 "lng_settings_angle_backend_auto" = "Auto";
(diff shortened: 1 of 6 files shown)


ilya-fedin:
Conflicts

Issue #26749, opened 2023-09-03
Desktop video player doesn't play first audio track in mp4


algielen:
### Steps to reproduce

1. Create video file with two audio tracks, in my case MP4 created with Shadowplay and the feature "Separate both tracks"
2. Send video file to a Telegram conversation
3. Play the video on desktop in the integrated video player

ffprobe gives the following info for the file
```
Input #0, mov,mp4,m4a,3gp,3g2,mj2, from '.\filename.mp4':
  Metadata:
    major_brand     : mp42
    minor_version   : 0
    compatible_brands: isommp42
    creation_time   : 2023-09-03T17:37:52.000000Z
    date            : 2023
  Duration: 00:00:30.22, start: 0.000000, bitrate: 34165 kb/s
  Stream #0:0[0x1](und): Video: h264 (High) (avc1 / 0x31637661), yuv420p(tv, smpte170m, progressive), 1920x1080 [SAR 1:1 DAR 16:9], 33768 kb/s, 60 fps, 60 tbr, 90k tbn (default)
    Metadata:
      creation_time   : 2023-09-03T17:37:52.000000Z
      handler_name    : VideoHandle
      vendor_id       : [0][0][0][0]
  Stream #0:1[0x2](und): Audio: aac (LC) (mp4a / 0x6134706D), 48000 Hz, stereo, fltp, 192 kb/s (default)
    Metadata:
      creation_time   : 2023-09-03T17:37:52.000000Z
      handler_name    : SoundHandle
      vendor_id       : [0][0][0][0]
  Stream #0:2[0x3](und): Audio: aac (LC) (mp4a / 0x6134706D), 48000 Hz, stereo, fltp, 192 kb/s (default)
    Metadata:
      creation_time   : 2023-09-03T17:37:52.000000Z
      handler_name    : SoundHandle
      vendor_id       : [0][0][0][0]
```

### Expected behaviour

Video player should load the first audio track, as on Telegram mobile, or as in VLC and mpv

### Actual behaviour

Video player loads the second audio track, which is not the intended one

### Operating system

Windows 11 (22H2)

### Version of Telegram Desktop

4.9.4 x64

### Installation source

Static binary from official website

PR #31131, opened 2026-08-09
Media viewer: add audio track selection for videos

Arunendra21:
## Problem

A video file can contain more than one audio stream — most commonly multiple
language tracks. The desktop media viewer always selects a single "best" audio
stream via `av_find_best_stream()` in the streaming `File` and drops packets
from every other audio stream, so there is currently **no way to switch the
audio track** while watching a video. (This is also the root of the wrong-track
behaviour reported in #26749.)

## Solution

Add an open-time audio-track selector, reusing the existing stream-restart path
rather than introducing live mid-stream switching:

- **Enumerate** every audio stream in `File::Context::start()` together with its
  `language`/`title` metadata into a new `AudioTrackInfo`, and carry the list up
  through `Information::audio` (`tracks` + `currentIndex`).
- **Select**: `PlaybackOptions::audioStreamIndex` / `StartOptions::audioStreamIndex`
  let a caller request a specific audio stream. `initStream()` passes it as the
  `wanted_stream_nb` of `av_find_best_stream()` after validating it points at an
  audio stream, otherwise it falls back to automatic selection.
- **UI**: the media-viewer context menu now lists the available audio tracks when
  a video has sound and more than one audio stream. Choosing a track sets
  `_chosenAudioTrack` and reopens the stream at the current position on that track
  via the existing `restartAtSeekPosition()`.

**Default behaviour is unchanged.** With `audioStreamIndex == -1` the code path is
identical to before (`av_find_best_stream(..., -1, ...)`), and single-track videos
show no extra menu entry.

## Files touched

- `media/streaming/media_streaming_common.h` — `AudioTrackInfo`, `PlaybackOptions::audioStreamIndex`, `AudioInformation::tracks/currentIndex`
- `media/streaming/media_streaming_file.{h,cpp}` — `StartOptions::audioStreamIndex`, track enumeration, requested-stream selection
- `media/streaming/media_streaming_file_delegate.h` + `media_streaming_player.{h,cpp}` — plumb the track list through `fileReady()` into `Information`
- `media/view/media_view_overlay_widget.{h,cpp}` — context-menu selector + `applyAudioTrack()`

## How to test

1. Open a video that has ≥2 audio streams (e.g. a file muxed with two language
   tracks, or an NVIDIA ShadowPlay "separate tracks" capture).
2. Right-click the video in the media viewer → the audio tracks are listed, the
   active one marked.
3. Pick another track → playback resumes at the same position with the selected
   audio.
4. Single-track videos are unaffected (no menu entry).

## Status / notes for reviewers

- Opened as a **draft**: I was unable to run a full local build of the client to
  compile- and runtime-verify this, so I'd appreciate a sanity check on the
  approach before polishing.
- The track labels are placeholder strings — if the direction is acceptable I'll
  add a proper `tr::` lang key for the "Audio track N" fallback and, if preferred,
  move the selector into the playback-controls dropdown next to the quality/speed
  menus instead of (or in addition to) the context menu.
- Threading: `_audioTracks` / `_audioTrackIndex` are written on the File thread in
  `fileReady()` and read on the main thread after the `streamReady()` dispatch,
  matching the existing pattern used for `_totalDuration`.

Related: #26749

### DIFF
--- a/Telegram/SourceFiles/media/streaming/media_streaming_common.h
+++ b/Telegram/SourceFiles/media/streaming/media_streaming_common.h
@@ -39,12 +39,22 @@ enum class Mode {
 	Inspection,
 };
 
+// Describes one selectable audio stream inside a media file. Filled in for
+// every audio stream that libavformat reports, so the UI can offer a way to
+// switch between them (for example multi-language tracks).
+struct AudioTrackInfo {
+	int index = -1; // AVStream index inside the format context.
+	QString language; // ISO code from the "language" metadata, if any.
+	QString title; // "title" metadata, if any.
+};
+
 struct PlaybackOptions {
 	Mode mode = Mode::Both;
 	crl::time position = 0;
 	crl::time durationOverride = 0;
 	float64 speed = 1.; // Valid values between 0.5 and 2.
 	AudioMsgId audioId;
+	int audioStreamIndex = -1; // -1 means automatic best-audio selection.
 	bool syncVideoByAudio = true;
 	bool waitForMarkAsShown = false;
 	bool hwAllowed = false;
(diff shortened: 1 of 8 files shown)


ilya-fedin:
CLA isn't signed in a month

PR #31145, opened 2026-08-15
Name the chat corner buttons for screen readers

rezabakhshilaktasaraei:
The jump to bottom, mentions, reactions and poll votes buttons had no accessible name, so screen readers announced all four as just "button".

They now get a name each, and the unread counter that is painted on them as a badge is exposed as the description, so it is announced together with the name and updates with the badge.


john-preston:
Thanks!

PR #31148, opened 2026-08-15
Show back button in fullscreen mini apps on Linux

damnkrat:
Before:
image
After:
image

### DIFF
--- a/Telegram/Resources/bot_webview_shell_html/page.css
+++ b/Telegram/Resources/bot_webview_shell_html/page.css
@@ -390,9 +390,13 @@ select {
 	border-bottom: 0;
 	pointer-events: none;
 }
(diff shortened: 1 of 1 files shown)


ilya-fedin:
Can you explain why this change is needed?

damnkrat:
Because for some reason in fullscreen mode back button is absent only on linux. You can check it in [test bot](https://t.me/asmico_attach_bot) by pressing Show Back Button, than Request Fullscreen. As you can see back button exists only in non-fullscreen. That makes some apps less intuitive, ass you can back in them only by rmb menu

Before change:
image
After change:
image

Windows(not my screenshot):
image

john-preston:
Thanks!

PR #31166, opened 2026-08-24
pass proxy settings to webview

### LINKED ISSUE #28692: Add proxy support for WebView

Telegram has been unofficially blocked in some regions of Russia. At the moment, the WebView module does not support local proxy server settings. This is a bit inconvenient, I have to set up a proxy server globally in the system or use a vpn.

Perhaps the WebView module should support proxy server settings from the application. The SOCKS protocol will do.

Nope.

### Additional context

_No response_

cnt0:
Partially fixes https://github.com/telegramdesktop/tdesktop/issues/28692 (for SOCKS5 proxies, but not for MTPROTO)

Meant to be used with https://github.com/desktop-app/lib_webview/pull/151

### DIFF
--- a/Telegram/CMakeLists.txt
+++ b/Telegram/CMakeLists.txt
@@ -531,6 +531,8 @@ PRIVATE
     core/core_settings.h
     core/core_settings_proxy.cpp
     core/core_settings_proxy.h
+    core/core_webview_proxy.cpp
+    core/core_webview_proxy.h
     core/crash_report_window.cpp
     core/crash_report_window.h
     core/crash_reports.cpp
--- a/Telegram/SourceFiles/core/core_webview_proxy.cpp
+++ b/Telegram/SourceFiles/core/core_webview_proxy.cpp
@@ -1,0 +1,29 @@
+/*
+This file is part of Telegram Desktop,
+the official desktop application for the Telegram messaging service.
+
+For license and copyright information please follow this link:
+https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
+*/
+#include "core/core_webview_proxy.h"
+
+#include "core/application.h"
+#include "core/core_settings.h"
+
+namespace Core {
+
+std::optional<Webview::ProxySettings> CurrentWebviewProxy() {
+	const auto &proxy = App().settings().proxy();
+	if (!proxy.isEnabled()) {
+		return std::nullopt;
+	}
+	const auto data = proxy.selected();
+	return Webview::ProxySettings{
+		.host = data.host.toStdString(),
+		.port = std::to_string(data.port),
+		.username = data.user.toStdString(),
+		.password = data.password.toStdString(),
+	};
+}
+
+} // namespace Core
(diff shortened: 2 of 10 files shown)


maksimtech:
Thanks for working on this — the webview gap has been open a long time.

One thing worth flagging while this is being looked at: the same MTProto caveat you mention applies to a path that isn't a webview at all, so extending the fix there might be worth considering.

**`Core::ResolveLocationAddress` falls back to a direct Mapbox request.**

`Telegram/SourceFiles/core/current_geo_location.cpp:199-215` tries the platform geocoder first and falls back to `ResolveLocationAddressGeneric` when it returns nothing:

```cpp
auto done = [=, done = std::move(callback)](GeoAddress result) mutable {
    if (!result && !token.isEmpty()) {
        ResolveLocationAddressGeneric(location, language, token, std::move(done));
    } else {
        done(result);
    }
};
Platform::ResolveLocationAddress(location, language, std::move(done));
```

On Linux that fallback is not an edge case — `platform/linux/current_geo_location_linux.cpp:98-102` is a stub:

```cpp
void ResolveLocationAddress(
        const Core::GeoLocation &location,
        const QString &language,
        Fn<void(Core::GeoAddress)> callback) {
    callback({});
}
```

So on Linux the generic path always runs, and it issues a plain request to `https://api.mapbox.com/search/geocode/v6/reverse?longitude=…&latitude=…` (`current_geo_location.cpp:70-71`) through a bare `QNetworkAccessManager` (`:93`, used at `:121`) that sets no proxy of its own. It therefore relies entirely on Qt's application-wide proxy.

**Which is where the MTProto case bites.** `Sandbox::refreshGlobalProxy()` (`core/sandbox.cpp:672-688`):

```cpp
if (proxy.type == MTP::ProxyData::Type::Socks5
    || proxy.type == MTP::ProxyData::Type::Http) {
    QNetworkProxy::setApplicationProxy(
        MTP::ToNetworkProxy(MTP::ToDirectIpProxy(proxy)));
} else if (!Core::IsAppLaunched()
    || Core::App().settings().proxy().isSystem()) {
    QNetworkProxyFactory::setUseSystemConfiguration(true);
} else {
    QNetworkProxy::setApplicationProxy(QNetworkProxy::NoProxy);
}
```

With an MTProto proxy selected, `isEnabled()` is true but the type matches neither branch, and `isSystem()` is false (`Settings::Enabled` and `Settings::System` are mutually exclusive — `core/core_settings_proxy.cpp`), so it lands on `setApplicationProxy(QNetworkProxy::NoProxy)`.

The result is that reverse geocoding reaches Mapbox over a direct connection carrying the user's real IP, even though the user has a proxy configured and every MTProto connection is going through it. For someone using an MTProto proxy specifically because direct connections are undesirable, sharing a location is a quiet exception.

**Suggestion:** an explicit `setProxy()` on that `QNetworkAccessManager`, derived from the current settings the same way the webview path is getting here, would close it. It is a small, self-contained change and it does not depend on the per-platform webview APIs this PR deals with.

Entirely reasonable if you would rather keep this PR focused on the webview and leave that to a separate one — I mainly wanted the non-webview case on record, since `#28692` is framed around webviews and this path would likely be missed.

Checked against `v7.2.9` (`4d4da47`). I have not tested the MTProto scenario at runtime, only read the code, so a second pair of eyes on that branch would be welcome.

---
*Analysis assisted by Claude Opus 5.*

Issue #263, opened 2014-11-16
Turn of youtube preview.


luckydonald:
Hey,
Is there a way to turn off the youtube preview?

Aokromes:
yeah maybe one dropdown "small, medium, large (current size), extra large (taking the whole window width), full screen"

luckydonald:
Also, doesn't a youtube preview in a secret chat expose that information to the internet?
(Not sure if it does that, nobody around to start a secret chat with...)

sigod:
There's no Secret Chats in Telegram Desktop yet.

joshery420:
Correct, no Secret Chats in desktop yet.

Also, +1 for @Aokromes list as long as we add "No preview" that simply shows the hyperlink, since that was the original request.

yuergen:
Maybe this could be resolved together with #687.

PR #31169, opened 2026-08-24
Leave the painted list focus announcement to the platform

rezabakhshilaktasaraei:
Tabbing into the chat list, a message list, the countries list or the languages list announces the current row twice. Taking keyboard focus raises a focus event which the platform resolves through focusChild() - the current row - and each of these lists announced that row from focusInEvent as well.

Only the platform's announcement is left. What focus-in still has to do is point at the right row before it returns: the chat list, the countries and the languages lists already pick it there, and for the two message lists the helper that resolves the focused message - following it across slice changes, or picking the first unread / the newest one - is split from the announcing one, whose other caller (a chat focused before its first messages arrived) has no focus event to rely on and keeps announcing directly.

Needs desktop-app/patches#263 on Windows: Qt 5.15 forwards a focus event to focusChild() only for an accessible exposing a table interface, upstream generalized that to any element with children in 6.2.

Tested with NVDA on Windows 10 against a Qt 5.15.19 build with that patch: all five lists announce their current row once when focus arrives, and browsing inside them with the arrows is unchanged. The list container keeps reporting the focused state, since it is the widget that actually holds keyboard focus - the row is what the focus event and GetFocus resolve to.

### DIFF
--- a/Telegram/SourceFiles/boxes/language_box.cpp
+++ b/Telegram/SourceFiles/boxes/language_box.cpp
@@ -363,15 +363,10 @@ void Rows::focusInEvent(QFocusEvent *e) {
 			setSelected(chosen >= 0 ? chosen : 0, Announce::No);
 		}
 	}
+	// The row is picked above, before the base class runs: taking focus raises
+	// a focus event of its own, which the platform hands to focusChild() - the
+	// selected row - so announcing it here as well would read it twice.
 	RpWidget::focusInEvent(e);
-	const auto index = selected();
-	if (index >= 0) {
-		InvokeQueued(this, [=] {
-			if (selected() == index && hasFocus()) {
-				accessibilityChildFocused(index);
-			}
-		});
-	}
 }
 
 void Rows::keyPressEvent(QKeyEvent *e) {
@@ -1173,7 +1168,7 @@ void Rows::accessibilityChildSetFocus(quintptr identity) {
 		// The rows are virtual (no real QWidget), so the screen reader's
 		// SetFocus can't move real keyboard focus to a row. Translate it
 		// into our internal selection, then either announce it directly or
-		// grab keyboard focus (focusInEvent announces the selected row).
+		// grab keyboard focus (taking it announces the row by itself).
 		setSelected(index, hasFocus() ? Announce::Always : Announce::No);
 		_mustScrollTo.fire(rowScrollRequest(index));
 		if (!hasFocus()) {
@@ -1195,9 +1190,9 @@ void Rows::accessibilityChildActivate(quintptr identity) {
 		// state change of the element it considers focused. Take focus onto
 		// the row (as the SetFocus action does) before activating, so the
 		// "checked" announcement is heard from Invoke as well as from Space:
-		// focusInEvent announces the row when focus moves here, OnChange
-		// covers an invoke on a not-focused row of the focused list, and an
-		// invoke on the already-focused row announces the change alone.
+		// taking focus announces the row by itself, OnChange covers an invoke
+		// on a not-focused row of the focused list, and an invoke on the
+		// already-focused row announces the change alone.
 		setSelected(index, hasFocus() ? Announce::OnChange : Announce::No);
 		if (!hasFocus()) {
 			setFocus();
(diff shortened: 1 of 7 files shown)


john-preston:
**Verdict: CHANGES_REQUIRED.** The Windows duplication fix relies on forwarding that the AT-SPI backend does not perform. Preserve a row-targeted focus notification on that backend. The helper split itself is sound.

## Findings

**P2 — Preserve child focus events on AT-SPI.** The changed [HistoryInner focus handler](https://github.com/telegramdesktop/tdesktop/blob/04c0c70666e90ed00ac0c2e1a5c41f3148611a7e/Telegram/SourceFiles/history/history_inner_widget.cpp#L6834) now only updates internal focus. The same removal appears in [ListWidget](https://github.com/telegramdesktop/tdesktop/blob/04c0c70666e90ed00ac0c2e1a5c41f3148611a7e/Telegram/SourceFiles/history/view/history_view_list_widget.cpp#L6394), [dialogs](https://github.com/telegramdesktop/tdesktop/blob/04c0c70666e90ed00ac0c2e1a5c41f3148611a7e/Telegram/SourceFiles/dialogs/dialogs_inner_widget.cpp#L6461), [languages](https://github.com/telegramdesktop/tdesktop/blob/04c0c70666e90ed00ac0c2e1a5c41f3148611a7e/Telegram/SourceFiles/boxes/language_box.cpp#L366), and [countries](https://github.com/telegramdesktop/tdesktop/blob/04c0c70666e90ed00ac0c2e1a5c41f3148611a7e/Telegram/SourceFiles/ui/boxes/country_select_box.cpp#L336).

Trigger: with an AT-SPI client subscribed to focus events, Tab/Backtab into any populated affected list, or invoke a row's SetFocus while its container lacks keyboard focus. Previously the queued call to [RpWidget::accessibilityChildFocused](https://github.com/desktop-app/lib_ui/blob/e3163c75370fdb37f00ff2a6e354c2a9721eedbd/ui/rp_widget.cpp#L460) set the row index explicitly. The remaining native event identifies the QWidget/container. The complete configured Qt 5.15.19 route is:

- `QWidget::setFocus`, `qwidget.cpp:6349–6355`: deliver widget focus, then construct `QAccessibleEvent(f, Focus)` without `setChild`.
- `QAccessible::updateAccessibility`, `qaccessible.cpp:873–896`, and `QAccessibleEvent::accessibleInterface`, `1778–1795`: resolve only an explicitly supplied child; otherwise retain the widget interface. Pinned application, lib_ui and lib_base have no installed update-handler override.
- `QSpiAccessibleBridge::notifyAccessibilityUpdate`, `bridge.cpp:104–109`: forward the same event.
- `AtSpiAdaptor::notify`, `atspiadaptor.cpp:932–935`, then `sendFocusChanged`, `1163–1184`: emit the focused-state and Focus signals using that interface's own object path. `pathForInterface` only obtains its accessible ID; it does not descend to `focusChild`.

This is also the official Qt 6.11.2 behavior: its [bridge](https://github.com/qt/qtbase/blob/v6.11.2/src/gui/accessible/linux/qspiaccessiblebridge.cpp#L76) passes the event unchanged, the [Focus branch](https://github.com/qt/qtbase/blob/v6.11.2/src/gui/accessible/linux/atspiadaptor.cpp#L1056) takes `event->accessibleInterface()`, and [sendFocusChanged](https://github.com/qt/qtbase/blob/v6.11.2/src/gui/accessible/linux/atspiadaptor.cpp#L1371) emits on that object's path. The confirmed regression is loss of the row-targeted entry notification, not a measured claim about Orca's exact spoken output. The first subsequent arrow still explicitly announces a row, but it cannot replace the missing entry event.

Fix this through one backend-aware strategy: retain the queued child notification on bridges without forwarding, while suppressing duplicates where forwarding is available; alternatively supply and integrate equivalent AT-SPI forwarding first. Do not infer support from a generic Qt version or `QAccessible::isActive()`.

Windows with patches #263 does redirect this event through `focusChild()`. Cocoa requires separate reasoning: configured `qcocoaaccessibility.mm:69–70` posts a focus-change notification, while `qcocoaaccessibilityelement.mm:624–637` answers the focused-element query through `focusChild()`. The AT-SPI finding therefore does not establish the same Cocoa failure; VoiceOver still needs validation.

PR #31176, opened 2026-08-26
Expose the folders sidebar as tabs, not a list

rezabakhshilaktasaraei:
﻿The folders sidebar was exposed to screen readers as a list, but the UI calls these folders tabs, so a screen reader user expects the tab role - the topic tabs strip and the horizontal folder tabs already report it. The sidebar container is a tab control now and every real folder a tab, the active one reported as selected; a locked (premium) folder is a tab that opens the Premium box on Enter instead of becoming current, which reads fine in practice. Only the "Edit" button stays a plain button.

The second fix is about where Tab enters the strip: browsing with the arrows used to leave the roving Tab stop on the last browsed folder, so Tabbing back into the sidebar landed there instead of on the active folder. Once focus leaves the list, the stop parks back on the active one, matching the topic tabs strip.

The context menu key also used to open the folder menu at the mouse cursor, which may sit nowhere near the folder or outside the window altogether; a menu asked for from the keyboard is anchored on the folder button now.

The favorite link button (the experimental button at the bottom of the strip that opens a configured link) only painted its label, so a screen reader announced it as an anonymous button - the label doubles as its accessible name now.

The unread dot on the main menu button - unread messages in other accounts - is only painted too; the button carries it as its accessible description now, both on the sidebar variant and on the chat list toggle shown without the sidebar.

Needs desktop-app/lib_ui#348 for the PageTab button role and for the selection state of SideBarButton tabs.

Tested with NVDA on Windows 10: the strip announces as a tab control, folders as tabs with the active one selected, selection is announced on Enter, locked folders open the Premium box, and leaving the sidebar and Tabbing back in lands on the active folder.

### DIFF
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -6096,6 +6096,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 
 "lng_sr_folder_locked" = "{text}, Premium required";
 "lng_sr_folder_locked_about" = "Press to view subscription options.";
+"lng_sr_menu_unread_other_accounts" = "Unread messages in other accounts";
 
 "lng_media_save_progress" = "{ready} of {total} {mb}";
 "lng_mediaview_save_as" = "Save As...";
--- a/Telegram/SourceFiles/dialogs/dialogs_widget.cpp
+++ b/Telegram/SourceFiles/dialogs/dialogs_widget.cpp
@@ -1693,6 +1693,15 @@ void Widget::setupMainMenuToggle() {
 			? &st::dialogsMenuToggleUnread
 			: &st::dialogsMenuToggleUnreadMuted;
 		_mainMenu.toggle->setIconOverride(icon, icon);
+		// The unread dot is only painted - surface it to a screen reader
+		// through the button's description.
+		const auto description = state.count
+			? tr::lng_sr_menu_unread_other_accounts(tr::now)
+			: QString();
+		if (_mainMenu.toggle->accessibleDescription() != description) {
+			_mainMenu.toggle->setAccessibleDescription(description);
+			_mainMenu.toggle->accessibilityDescriptionChanged();
+		}
 	}, _mainMenu.toggle->lifetime());
 }
 
--- a/Telegram/SourceFiles/window/window_filters_favorite.cpp
+++ b/Telegram/SourceFiles/window/window_filters_favorite.cpp
@@ -458,6 +458,8 @@ void FolderFavoriteButton::applyPeer(not_null<PeerData*> peer) {
 		Data::PeerUpdate::Flag::Name
 	) | rpl::on_next([=] {
 		_label.setText(_st.style, peer->name());
+		setAccessibleName(peer->name());
+		accessibilityNameChanged();
 		resizeToWidth(width());
 		update();
 	}, _resolveLifetime);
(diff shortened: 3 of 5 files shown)


john-preston:
**Verdict: CHANGES_REQUIRED.** Keep this small migration, but guard deferred focus parking against disabled screen-reader mode and remove duplicate accessibility notifications. The selected/tab semantics and keyboard menu positioning otherwise fit the existing controls. This is static review, not runtime certification.

**P2 — Deferred parking re-enables folder focus after mode is disabled.** The new [callback](https://github.com/telegramdesktop/tdesktop/blob/319e974b8efe1aa94c07eb0078083c1960af5a4f/Telegram/SourceFiles/window/window_filters_menu.cpp#L360-L367) checks only whether a folder currently holds focus. Reproduction sequence from the source: enable reader mode, focus a folder, stop the reader so ScreenReaderModeActiveValue emits false, then click the chat input. The existing [mode subscriber](https://github.com/telegramdesktop/tdesktop/blob/319e974b8efe1aa94c07eb0078083c1960af5a4f/Telegram/SourceFiles/window/window_filters_menu.cpp#L674-L681) first sets every folder to NoFocus. In configured Qt 5.15.19, `QWidget::setFocusPolicy` only changes the policy/proxy, so the focused folder can remain focused until that click. Its new [FocusOut handler](https://github.com/telegramdesktop/tdesktop/blob/319e974b8efe1aa94c07eb0078083c1960af5a4f/Telegram/SourceFiles/window/window_filters_menu.cpp#L692-L694) queues parking. With focus now outside, parking calls setListTabStop, which unconditionally gives the active folder TabFocus and, if different, the previous stop ClickFocus. The sidebar returns to the ordinary Tab/mouse focus path while reader mode remains off. PageTab is deliberately excluded from the library FocusManager's automatic registration, so no subsequent manager pass repairs this.

Check `Ui::ScreenReaderModeActive()` **inside the deferred callback**, together with listFocused, before changing policies. Checking only when scheduling would leave the transition-before-execution case open. The configured Qt evidence is `C:/Telegram/Libraries/qt_5.15.19/qtbase/src/widgets/kernel/qwidget.cpp:7635`; this sequence was traced, not executed.

**P3 — Use the setters' existing name/description notifications.** The [favorite peer-name update](https://github.com/telegramdesktop/tdesktop/blob/319e974b8efe1aa94c07eb0078083c1960af5a4f/Telegram/SourceFiles/window/window_filters_favorite.cpp#L460-L462), [sidebar menu description](https://github.com/telegramdesktop/tdesktop/blob/319e974b8efe1aa94c07eb0078083c1960af5a4f/Telegram/SourceFiles/window/window_filters_menu.cpp#L247-L249), and [chat-list menu description](https://github.com/telegramdesktop/tdesktop/blob/319e974b8efe1aa94c07eb0078083c1960af5a4f/Telegram/SourceFiles/dialogs/dialogs_widget.cpp#L1701-L1703) each call a QWidget setter and then explicitly emit the same event. Configured Qt's setters already dispatch NameChanged/DescriptionChanged (`qwidget.cpp:11409` and `:11439`); the RpWidget helpers simply dispatch that event again. Each qualifying update therefore produces two source notifications. Duplicate spoken output depends on the backend/reader and was not measured. Remove the three explicit notification calls; keep the description equality checks and setter calls. The setter in favorite `apply()` already supplies its notification correctly.

PR #31182, opened 2026-08-27
Open keyboard-invoked list menus on their rows

rezabakhshilaktasaraei:
﻿Pressing the context menu key on a focused chat or message already opened the menu of the right row - that part came with the earlier keyboard and screen reader work - but the menu itself popped at the position Qt puts into a keyboard-invoked event: the center of an empty input method rect, which lands near the top left corner of the list, nowhere near the row. For messages the reactions selector attaches based on the same position, so it drifted along.

The chat list menu now opens under the selected row, at the position the chat menu shortcut already computes for the events it synthesizes (that path is unchanged, and covers the default list and the search result rows). The message lists rebuild the keyboard event anchored on the visible part of the focused message before the shared handler runs, so everything reading the event position - the menu, the reactions selector, the poll option tabs - lands on the message. The chat type filter menu in search results also anchors on its link instead of the mouse cursor, which an Enter press may have left anywhere.

A mouse-invoked menu keeps opening at the cursor everywhere.

Needs desktop-app/lib_ui#348 for the Ui::ContextMenuPosition helper, the same policy the tab strips adopted. Only the placement of the popups changes - the reachability and content of the menus stay as they are, so screen reader interaction is unaffected.

### DIFF
--- a/Telegram/SourceFiles/dialogs/dialogs_inner_widget.cpp
+++ b/Telegram/SourceFiles/dialogs/dialogs_inner_widget.cpp
@@ -3954,7 +3954,34 @@ void InnerWidget::contextMenuEvent(QContextMenuEvent *e) {
 	if (_menu->empty()) {
 		_menu = nullptr;
 	} else {
-		_menu->popup(e->globalPos());
+		// A keyboard-invoked event carries the center of an empty input
+		// method rect for a position, which may sit nowhere near the row -
+		// anchor the menu under the selected row instead, at the position
+		// showPeerMenu computes for the events it synthesizes.
+		const auto rowBottom = [&] {
+			if (_state == WidgetState::Default) {
+				if (_selected) {
+					return _selected->top() + _selected->height();
+				}
+			} else if (_state == WidgetState::Filtered) {
+				if (base::in_range(_filteredSelected, 0, _filterResults.size())) {
+					const auto &result = _filterResults[_filteredSelected];
+					return filteredOffset() + result.top + result.row->height();
+				} else if (base::in_range(_previewSelected, 0, _previewResults.size())) {
+					return previewOffset() + (_previewSelected + 1) * _st->height;
+				} else if (base::in_range(_searchedSelected, 0, _searchResults.size())) {
+					return searchedOffset() + (_searchedSelected + 1) * _st->height;
+				}
+			}
+			return -1;
+		}();
+		const auto &padding = st::defaultDialogRow.padding;
+		const auto position = (fromMouse || rowBottom < 0)
+			? e->globalPos()
+			: mapToGlobal(QPoint(
+				width() - padding.right(),
+				rowBottom + padding.bottom()));
+		_menu->popup(position);
 		e->accept();
 	}
 }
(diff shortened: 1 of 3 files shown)


john-preston:
**Verdict: CHANGES_REQUIRED.** Preserve the focused-row approach, but handle wholly offscreen message anchors and correct the new default-chat coordinate calculation. The filter-link change also moves ordinary mouse popups to a fixed point; preserve the pointer path when distinguishing its activation source. No runtime tests were performed.

**Findings.**

1. **P2 — Keep a wholly offscreen message anchor inside the visible list.** [HistoryInner lines 2767–2770](https://github.com/telegramdesktop/tdesktop/blob/8413d75a866d17dfae4c56da0ebb5342bc12cd6f/Telegram/SourceFiles/history/history_inner_widget.cpp#L2767-L2770), duplicated at [ListWidget lines 3961–3964](https://github.com/telegramdesktop/tdesktop/blob/8413d75a866d17dfae4c56da0ebb5342bc12cd6f/Telegram/SourceFiles/history/view/history_view_list_widget.cpp#L3961-L3964). Focus a loaded message, wheel-scroll it wholly out of view while retaining keyboard focus, then press the context-menu key. Scroll-area updates change the visible bounds without clearing that focused item. When the rectangles no longer intersect, the wrapper keeps the original offscreen rectangle and maps its center to global coordinates. The plain popup path receives that point. [PopupMenu::prepareGeometryFor](https://github.com/desktop-app/lib_ui/blob/8188a76f9ddbc04b4517c6ef0db3f7170062359f/ui/widgets/popup_menu.cpp#L1008-L1051) chooses `screenAt(p)` and an empty bounds rectangle when no screen contains it; its later bounds corrections run only for nonempty screen bounds. There is no caller-side clamp here or automatic call to ConstrainToParentScreen. Thus this new path can request a menu outside the list/window, on another monitor, or outside all screens. This is source-traced bad placement, not an observed runtime disappearance. For a disjoint rectangle, use a point clamped to the visible list or a visible-list fallback while retaining the actual focused target. Apply the same policy to both wrappers.

2. **P3 — Include the painted-list offset in the new default-chat anchor.** [dialogs_inner_widget.cpp line 3964](https://github.com/telegramdesktop/tdesktop/blob/8413d75a866d17dfae4c56da0ebb5342bc12cd6f/Telegram/SourceFiles/dialogs/dialogs_inner_widget.cpp#L3961-L3965). With a collapsed Archive row, a skipped top dialog, or a community header, `_selected->top()` is a raw-list coordinate. The painted bottom is `dialogsOffset() + top() + height()`: [dialogsOffset](https://github.com/telegramdesktop/tdesktop/blob/8413d75a866d17dfae4c56da0ebb5342bc12cd6f/Telegram/SourceFiles/dialogs/dialogs_inner_widget.cpp#L708-L714) accounts for those structures, and [accessibilityChildRect](https://github.com/telegramdesktop/tdesktop/blob/8413d75a866d17dfae4c56da0ebb5342bc12cd6f/Telegram/SourceFiles/dialogs/dialogs_inner_widget.cpp#L6918-L6925) already uses the correct conversion. The newly added native-key anchor is displaced by that offset. Add it or reuse a shared painted-row rectangle. The old showPeerMenu shortcut already omitted this offset; this finding is incomplete new anchoring, not a claim that this PR first broke that shortcut. Collapsed Archive itself also retains the prior fallback position, rather than gaining row anchoring.

PR #31184, opened 2026-08-27
Keep the search tabs on manual activation

rezabakhshilaktasaraei:
With #31097 the tab strips switch to the tab the arrows land on, the way native Windows tab controls do. The Chats / Channels / Apps tabs above the search suggestions are an exception: switching them changes the search scope and can run new searches, so passing through them while browsing with a screen reader would fire intermediate searches. Keep these tabs on the browse-first model - the arrows move between them, Enter or Space commits.

Uses the opt-out added in #31097, so it compiles only on top of it.

### DIFF
--- a/Telegram/SourceFiles/dialogs/ui/dialogs_suggestions.cpp
+++ b/Telegram/SourceFiles/dialogs/ui/dialogs_suggestions.cpp
@@ -1410,6 +1410,9 @@ Suggestions::Suggestions(
 Suggestions::~Suggestions() = default;
 
 void Suggestions::setupTabs() {
+	// Switching a tab changes the search scope and can run new searches, so
+	// the arrows only browse these tabs, committing on Enter / Space.
+	_tabs->setAccessibilityActivateOnBrowse(false);
 	_tabsScroll->setCustomWheelProcess([=](not_null<QWheelEvent*> e) {
 		const auto pixelDelta = e->pixelDelta();
 		const auto angleDelta = e->angleDelta();


john-preston:
**Verdict: CHANGES_REQUIRED.** Keep manual activation, but reveal the browsed search tab without activating it. There is one functional finding.

**P2 — Scroll the manually browsed tab into view.** The [new opt-out at line 1415](https://github.com/telegramdesktop/tdesktop/blob/fa66869604c3a207a3ba0f18c396718edd196547/Telegram/SourceFiles/dialogs/ui/dialogs_suggestions.cpp#L1413-L1415) removes the activation event on which this container currently relies for scrolling. In #31097, [manual browsing returns after setAccessibilitySelected](https://github.com/telegramdesktop/tdesktop/blob/24a066315c7834ab01c3535e576f753b3b9403b3/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp#L325-L330), and that setter [emits accessibilitySectionBrowsed and a focus notification](https://github.com/telegramdesktop/tdesktop/blob/24a066315c7834ab01c3535e576f753b3b9403b3/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp#L371-L384), without activating a section. Suggestions has no subscription to that browse producer. Its [existing automatic scroll subscriber listens only to sectionActivated](https://github.com/telegramdesktop/tdesktop/blob/fa66869604c3a207a3ba0f18c396718edd196547/Telegram/SourceFiles/dialogs/ui/dialogs_suggestions.cpp#L1427-L1449).

This is a reachable layout, not just a theoretical long translation: [TabKeysFor supplies ten tabs, or eleven with Downloads](https://github.com/telegramdesktop/tdesktop/blob/fa66869604c3a207a3ba0f18c396718edd196547/Telegram/SourceFiles/dialogs/ui/dialogs_suggestions.cpp#L2311-L2329). [updateControlsGeometry fits the slider to all labels while constraining the viewport to the column width](https://github.com/telegramdesktop/tdesktop/blob/fa66869604c3a207a3ba0f18c396718edd196547/Telegram/SourceFiles/dialogs/ui/dialogs_suggestions.cpp#L2364-L2369). With reader mode active, open search suggestions in a narrow column, leave Chats selected, focus the strip, and press End or browse right to an initially offscreen tab. The accessible browse/focus position moves there while the horizontal offset remains unchanged; the tab is revealed only after committing. This loses visual tracking for users combining speech, magnification and keyboard navigation. It remains possible to activate the tab, so this is not described as an unreachable search scope.

Qt does not supply the missing scrolling: these are painted virtual children, and arrow browsing keeps the same QWidget focused. The inspected configured Qt 5.15.19 QScrollArea only ensures widget visibility during real focus traversal; its focus event path also has no virtual-section geometry. Ui::ScrollArea adds no browse-event subscription.

Subscribe to `_tabs->accessibilitySectionBrowsed()` after defining `scrollToIndex`, call `scrollToIndex(index, anim::type::normal)`, and bind the subscription to `_tabs->lifetime()`. Keep the existing activation subscription for mouse and programmatic changes. #31098 uses this same separation for folder tabs; it is a reference implementation, not an additional dependency. Do not route browsing through `setActiveSection` or `switchTab`, which would restore the unwanted intermediate scope changes.

PR #31185, opened 2026-08-27
Fix stuck window activity after a lost X11 focus event

kost-t-human:
On X11 the window can get stuck in an inactive state while it is visible and actually holds the input focus. Messages in an opened chat are not marked as read, GIFs and animated stickers stay frozen, and the tray icon click toggles the window the wrong way. Clicking or scrolling inside the window does not help, only switching to another window and back does.

## Root cause

`Window::MainWindow::computeIsActive()` relies on `isActiveWindow()`, and the two halves of Qt keep two different ideas of it.

QtWidgets resets `active_window` straight from the `hide()` call, through `hide_sys()` and `deactivateWidgetCleanup()`. No window system event is involved, it happens on every hide of the active window.

QtGui keeps `focus_window` from X11 focus events, and nothing clears it when the window hides. That is left to a 100 ms timer the xcb plugin starts when the FocusOut arrives, and any FocusIn the plugin accepts stops that timer before it fires.

So if the window is shown again before that timer fires, `focus_window` never passes through null, `processFocusWindowEvent()` returns early on `previous == newFocus`, `notifyActiveWindowChange()` never runs, and `activeWindow()` stays null until the focus really changes. Everything reading `MainWindow::isActive()`, marking as read, animation pausing, the tray toggle, is stuck with it.

## Steps to reproduce

The gap between the hide and the show has to be shorter than what is left of those 100 ms, which is about 80 ms here, so this is hard to hit by hand and easy to hit from a script. `xev` on the window, a healthy cycle and a broken one from the same run:

```
15:57:15.752  FocusOut  NotifyNonlinear   hide, doFocusOut starts the timer
15:57:16.358  FocusIn   NotifyNonlinear   show 600 ms later, accepted, window goes active

15:57:17.669  FocusOut  NotifyNonlinear   hide again
15:57:17.708  FocusIn   NotifyNonlinear   show 39 ms later, accepted, but previous == newFocus
```

After the second cycle the window is on screen and holds the input focus, and the client considers itself inactive until the focus moves away for real. A way to do it by hand, plus the plain Qt reproduction with no Telegram code involved, is in the comments.

## Fix

Destroy the native window when hiding on X11. `QWindowPrivate::destroy()` clears the focused window right there, so the two halves of Qt stop disagreeing, and showing the window back creates it again and the activation is delivered as usual.

`savePosition()` needs its visibility check moved above the `windowHandle()` use, because there is no handle while the window is hidden now.

## Testing

Built and used on Kubuntu with KWin 5.27 on X11. On a build carrying only #31189 the scripted reproduction gets stuck on every run, four out of four. With this patch on top it stays clean, three runs with a 30 ms gap and three more with no gap at all.

Also checked what the destroyed window costs: geometry and maximized state survive the hide and show, the position still reaches the settings across a restart, and the rss, the fd count and the window count stay flat over 30 cycles.

## Related

Related to #31189, which fixes the window manager refusing to activate the window from the tray. That refusal was the most common path into the state described here, and with it fixed the scenario became much rarer. This is still a separate defect: the state is reachable from an ordinary hide and show cycle with nothing refusing the activation, which is what the reproduction in the comments does.

### DIFF
--- a/Telegram/SourceFiles/platform/linux/main_window_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/main_window_linux.cpp
@@ -522,6 +522,9 @@ bool MainWindow::eventFilter(QObject *obj, QEvent *evt) {
 		}
 	} else if (obj == this && t == QEvent::Hide) {
 		_exposed = false;
+		if (IsX11()) {
+			destroy();
+		}
 	} else if (obj == this && t == QEvent::ThemeChange) {
 		updateWindowIcon();
 	}
--- a/Telegram/SourceFiles/window/main_window.cpp
+++ b/Telegram/SourceFiles/window/main_window.cpp
@@ -899,13 +899,14 @@ QRect MainWindow::computeDesktopRect() const {
 }
 
 void MainWindow::savePosition(Qt::WindowState state) {
+	if (!isVisible() || !positionInited()) {
+		return;
+	}
+
 	if (state == Qt::WindowActive) {
 		state = windowHandle()->windowState();
 	}
-
-	if (state == Qt::WindowMinimized
-		|| !isVisible()
-		|| !positionInited()) {
+	if (state == Qt::WindowMinimized) {
 		return;
 	}
 	if (const auto saved = Core::App().savedWindows()) {


kost-t-human:
Updated the description: linked #31189, which fixes the window manager refusing the activation — that turned out to be the most common path into the state described here. Also simplified the reproduction steps, waiting for an incoming message is not required.

ilya-fedin:
Do we really need to work around a bug on a dying platform reproduced on a window manager that will remove X11 support in 1.5 months? And then if the bug is KWin-specific, you make active window check way less performant (roundtrip to the X server vs reading Qt's cached value) for everyone on X11.

kost-t-human:
Good points. It looks like #31189 actually resolves the root cause -- the stale timestamp in the activation request -- without adding an X server roundtrip. It simply implements ActivateThisProcess() on Linux using the existing lib_base function to get the timestamp, similar to how it works on macOS.
On the KWin-specific part -- I only tested on KWin, so I can't rule that out. What the logs show is a disagreement inside Qt itself rather than something the window manager does: after hiding to tray, QApplication::activeWindow() is reset by hide(), while QGuiApplication::focusWindow() keeps pointing at the very window Qt has just unmapped. I did not check the raw X event stream, though, so which side drops the focus event is still a guess on my part.

ilya-fedin:
> It looks like #31189 actually resolves the root cause -- the stale timestamp in the activation request -- without adding an X server roundtrip. It simply implements ActivateThisProcess() on Linux using the existing lib_base function to get the timestamp, similar to how it works on macOS.

Yeah, I asked Claude to check KWin code and it confirms #31189 while this change works by accident due to _isActiveTimer and then the current way you check is for taskbar impls to check the current active window while the app itself should check `_NET_WM_STATE_FOCUSED` on its own window.

kost-t-human:
I can rework this to track `_NET_WM_STATE_FOCUSED` on our own window via PropertyNotify. That removes the roundtrip and gives a proper update trigger instead of relying on _isActiveTimer. 
I doubt it's worth it. With only #31189 applied, the scenario that used to produce unread messages several times an hour hasn't reproduced in 2 hour of use.

kost-t-human:
Reworked it. Now it reads `_NET_WM_STATE_FOCUSED` on our own window and keeps it up to date via `PropertyNotify` on `_NET_WM_STATE`, so there is no per-call roundtrip and the update no longer depends on `_isActiveTimer`. Verified the state flips within the same second the X server sees it.

Still your call whether this is worth having. Thanks again for the review!

ilya-fedin:
> Reworked it.

Did you reproduce after all?

ilya-fedin:
fwiw that was a review from claude but personally I was thinking that calling `QWindowSystemInterface::handleFocusWindowChanged(nullptr);` on hide is maybe a better workaround.

kost-t-human:
> Did you reproduce after all?

No, I didn't reproduce it. I reworked it because he pointed out a weak spot in the implementation (which struck me as the right move), not because I managed to trigger the bug again.
Once I had the context, it was quick. The hard part was figuring out what was causing the bug in the first place.

kost-t-human:
`QWindowSystemInterface::handleFocusWindowChanged(nullptr)` on hide, guarded by `focusWindow() == windowHandle()` so that hiding an inactive window does not drop the focus of another one. 11 lines, no X access at all, and `window/main_window.h` and `.cpp` are untouched now. Much better than what I had, thanks.

ilya-fedin:
Nice. Run the build with just the fix from #31189 for some time, if you reproduce the bug without this fix, please tell so and then check whether this variant actually fixes the issue. If it fixes, tell so and it could be merged. If #31189 is enough after all, just close the PR.

kost-t-human:
> Nice. Run the build with just the fix from #31189 for some time, if you reproduce the bug without this fix, please tell so and then check whether this variant actually fixes the issue. If it fixes, tell so and it could be merged. If #31189 is enough after all, just close the PR.

Sorry for the length. I tried to cut it down, but the short version reads like a list of claims you would have to take on faith, and I would rather show the reasoning and the numbers so you can check any of it yourself.

Short answer first: yes on both counts, with one caveat: the reproduction drives the same user actions from a script rather than by hand, because by hand the window is only about 80 ms wide. It reproduces on a build that has only #31189, on demand rather than by luck, and it stops reproducing with this patch on top. The rest is how I got there and what exactly goes wrong.

I ran a build with only #31189 for a while and could not hit it by hand, so I went looking for the exact mechanism instead. It turns out the state is reproducible on demand, with no Telegram code involved at all.

## What actually happens

The two halves of Qt keep two different ideas of "this window is active", and once they disagree, nothing reconciles them until the focus genuinely moves to another window and back. Line numbers below are from 6.11.1, the version tdesktop builds.

* `QWidgetPrivate::hide_helper()` reaches `deactivateWidgetCleanup()` through `hide_sys()` (`qwidget.cpp:8253`, `:8310`, then `:2350`), which resets `active_window` to null whenever the widget being hidden is the active window. This is driven straight from the `hide()` call, not by any window system event.
* On the QtGui side nothing clears `focus_window` when the window hides. The only thing that eventually does is the timer that `QXcbWindow::doFocusOut()` starts when the FocusOut arrives (`qxcbwindow.cpp:854`), with an interval of 100 ms (`qxcbconnection.cpp:92`).
* Any FocusIn that the plugin accepts stops that timer before it fires (`qxcbwindow.cpp:2294`), and one with detail NotifyPointer is dropped even earlier, without stopping it (`:2291`).
* `QGuiApplicationPrivate::processFocusWindowEvent()` returns early when `previous == newFocus` (`qguiapplication.cpp:2706`), so `notifyActiveWindowChange()` never runs and `active_window` is never restored.

So if `focus_window` does not pass through nullptr between the hide and the show, `isActiveWindow()` stays false while the window is visible and holds the input focus, and it stays false until the focus really changes. Everything that reads `MainWindow::isActive()` is stuck with it: marking as read, animation pausing, the tray toggle.

The teardown path at least notices the case: `QXcbWindow::destroy()` calls `doFocusOut()` when the window being destroyed is the focus window (`qxcbwindow.cpp:530`). That still goes through the same timer, and `hide()` does not do even that much.

## The mechanism is in Qt, not in Telegram

A small PyQt6 script, no Telegram code at all, on system Qt 6.4.2 (this part of the xcb plugin is unchanged in 6.11.1, apart from `handleWindowActivated` being renamed to `handleFocusWindowChanged`). KWin is the only window manager I have here, but the decision is made inside Qt, the window manager just delivers the focus events. Show a widget, activate it, `hide()`, then `show()` plus `activateWindow()` N ms later, then look at `isActiveWindow()`. Ten invocations per delay, on this machine:

```
 10 ms: reproduced 10 of 10
 50 ms: reproduced 10 of 10
 80 ms: reproduced 3 of 10, then 5 of 10 on a repeat
150 ms: 0 of 10
250 ms: 0 of 10
500 ms: 0 of 10
```

In the reproduced state `_NET_ACTIVE_WINDOW` points at the window, `QGuiApplication::focusWindow()` points at the window, `QApplication::activeWindow()` is null and `isActiveWindow()` is false. The boundary sits some tens of milliseconds below the 100 ms of `focusInDelay`, because showing the window and getting the focus back eats into it, see the budget below.

<details>
<summary>the script, in case you want to run it yourself</summary>

Save as `focusdesync.py`, run as `python3 focusdesync.py 50` on an X11 session.

```python
import shutil, subprocess, sys
from PyQt6.QtCore import QTimer
from PyQt6.QtGui import QGuiApplication
from PyQt6.QtWidgets import QApplication, QWidget

DELAY = int(sys.argv[1]) if len(sys.argv) > 1 else 50

app = QApplication(sys.argv[:1])
if app.platformName() != 'xcb':
    sys.exit(f'this race is X11 only, but Qt runs on "{app.platformName()}", '
             'try QT_QPA_PLATFORM=xcb in an X11 session')
if not shutil.which('xprop'):
    sys.exit('xprop not found, install x11-utils')
w = QWidget(); w.resize(420, 140); w.setWindowTitle('focus desync probe')
w.show(); w.raise_(); w.activateWindow()


def xactive():
    out = subprocess.run(['xprop', '-root', '_NET_ACTIVE_WINDOW'],
                         capture_output=True, text=True).stdout
    return out.strip().split('# ')[-1].split(',')[0]


def state(tag):
    fw = QGuiApplication.focusWindow()
    print(f"{tag:11} isActiveWindow={w.isActiveWindow()!s:5} "
          f"activeWindow={'set' if QApplication.activeWindow() else 'NULL':4} "
          f"focusWindow={'ours' if fw is w.windowHandle() else 'other':5} "
          f"x11_active={xactive()} ours={hex(int(w.winId()))}", flush=True)


waits = [0]


def arm():
    if w.isActiveWindow():
        state('armed')
        w.hide()
        QTimer.singleShot(DELAY, back)
        return
    waits[0] += 1
    if waits[0] > 40:
        print('could not activate the window, nothing was tested')
        app.quit()
        return
    QTimer.singleShot(100, arm)


def back():
    w.show(); w.raise_(); w.activateWindow()
    QTimer.singleShot(600, verdict)


def verdict():
    state('after show')
    if w.isActiveWindow():
        line = 'healthy'
    elif xactive() == hex(int(w.winId())):
        line = 'REPRODUCED, the window is focused and considered inactive'
    else:
        line = 'INCONCLUSIVE, the window never got the X focus back'
    print(f"\n{DELAY} ms: {line}")
    app.quit()


QTimer.singleShot(300, arm)
sys.exit(app.exec())
```

</details>

## Reproduced on the real client, with #31189 but without this patch

A small script drives two ordinary user paths over DBus. Hide through the tray icon (`org.kde.StatusNotifierItem.Activate`) and show through `org.freedesktop.Application.Activate`, the GApplication entry point that a desktop entry launch goes through. A notification click comes back through the notifications interface as `ActionInvoked`, or, when GNotification is in use, through `ActivateAction` on the same `org.freedesktop.Application` interface as the show path, and a second launch of the binary takes the single instance socket. All of these show paths bypass the `doubleClickInterval()` throttle that plain tray clicks go through (400 ms by default, `tray.cpp`), so the show can land arbitrarily soon after the hide. The script calibrates itself, one slow cycle first, then a fast one. It reads its verdict from the first tray menu item, so it needs no debug build. I left it out of the comment because it is a hundred lines of DBus plumbing, happy to paste it if you want it.

`xev` on the window during one such run:

```
15:57:15.752  FocusOut  NotifyNonlinear   hide, doFocusOut starts the timer
15:57:16.358  FocusIn   NotifyNonlinear   show 600 ms later, accepted

15:57:17.669  FocusOut  NotifyNonlinear   hide again
15:57:17.672  FocusIn   NotifyPointer     focus reverted to PointerRoot, dropped by the plugin
15:57:17.708  FocusIn   NotifyNonlinear   show 39 ms later, accepted, but previous == newFocus
```

And the client at the same moments. The build is `dev` plus #31189 with a small probe added: it writes a line on every visibility change, and once a second it compares the two Qt states, reporting a mismatch once it has lasted three samples and dating it back by those three seconds, which is why the duration below is longer than the gap between the two log lines:

```
15:57:15  visible [FALSE], gui focused [TRUE],  widgets active [FALSE]
15:57:16  visible [TRUE],  gui focused [FALSE], widgets active [FALSE]
15:57:17  visible [FALSE], gui focused [TRUE],  widgets active [FALSE]
15:57:17  visible [TRUE],  gui focused [TRUE],  widgets active [FALSE]
15:57:20  mismatch started, gui focused [TRUE], widgets active [FALSE], visible [TRUE], minimized [FALSE], isActive [FALSE]
15:57:58  mismatch ended after 41000ms
```

The interesting column is `gui focused` on the `visible [TRUE]` line. In the healthy cycle it is FALSE, because the timer had already cleared `focus_window`, so the FocusIn that follows is a real transition, the client goes active and no mismatch is ever reported. In the fast cycle it is TRUE, the FocusIn changes nothing, and the client stays inactive for the next 41 seconds, until the focus really moves away.

## The timing budget

Where the 100 ms goes, taken from the trace above. The script left 30 ms between its two calls and the accepted FocusIn arrived 39 ms after the FocusOut, so the FocusIn trailed the show call by about 9 ms more than the FocusOut trailed the hide call. That puts the budget for the gap itself a little over 90 ms on this machine, and in the freeze recipe below the two calls were handled 19 ms apart, comfortably inside it. The PyQt script pays more than those 9 ms, because its window has to be mapped and taken over by the window manager again before it can be focused, which is why 80 ms is already a coin flip in its table rather than a clean reproduction. Expect that line to move between runs, it sits right on the edge.

And that is why it feels so random in daily use, because two deliberate actions by hand never came closer than about 150 ms in my tries, so the window is normally missed. What changes the odds is a stall of the client main thread. The real budget is 100 ms plus however long the main thread was busy, because both focus events are then processed in the same drain of the event loop.

## Reproducing it by hand

That last point gives a manual recipe. Freeze the client, click calmly, unfreeze:

```sh
pid=$(pidof Telegram)
( sleep 5; kill -STOP $pid; sleep 3
  gdbus call --session --dest org.telegram.desktop \
    --object-path /org/telegram/desktop \
    --method org.freedesktop.Application.Activate "{}" >/dev/null 2>&1 &
  sleep 1; kill -CONT $pid ) &
```

Adjust the process name if your build installs as `telegram-desktop`. Check the bus name too: a build with the autoupdater enabled owns `org.telegram.desktop._<hash>` rather than the plain name (`specific_linux.cpp`), so run `busctl --user list | grep telegram` first and use what it prints as `--dest`, otherwise the call goes nowhere and the recipe quietly does nothing. Click the Telegram window within the first 5 seconds so that it becomes active, then click the tray icon while the client is frozen, then keep your hands off. The panel sends the same `Activate` call as always, it just sits in the queue until the client runs again. I measured the run with both calls sent from the script, three seconds apart on the outside, and inside the client they landed 19 ms apart.

One catch worth knowing. The hide is only read as "minimize to tray" if the client still considers itself active when it processes it, so nothing may take the focus in between. Switching to a terminal to type the commands is enough to break the run, which is why everything above is scheduled in advance.

To be straight about it, I never managed to hit this by hand in normal use, only with the freeze above or with the script. What I do have is one occurrence in yesterday's log, on a build without this patch, that happened on its own while I was simply using the client. My PC was under constant heavy load all of yesterday, which fits the mechanism rather well: the busier the main thread, the wider the window, because the two focus events then land in the same drain of the event loop. On an idle PC I have not seen a spontaneous one since.

## What this patch covers

The healthy cycles differ from the broken one in exactly one thing: `focus_window` passed through nullptr. The patch makes that happen on every hide, instead of leaving it to a timer that a returning FocusIn can cancel.

I then built this branch with #31189 cherry-picked on top and ran the same reproduction against it. Three runs with a 30 ms gap and three runs with no gap at all, all of them clean, while the build with only #31189 got stuck on all four runs I did with it. So the answer to your question is yes on both counts: it reproduces without this patch and it stops reproducing with it.

There is a second way into the same state that this patch does not touch, and I think it is worth writing down. On my machine, while the window is being hidden the focus reverts to PointerRoot, so if the pointer happens to be over the window, the window gets FocusIn with detail NotifyPointer, which the plugin drops (`qxcbwindow.cpp:2291`). The client becomes active only thanks to the explicit FocusIn NotifyNonlinear that KWin sends afterwards. When that one does not arrive, the window sits on screen with the keyboard focus while the client thinks it is inactive. That is the path #31189 closed, by giving the activation request a fresh timestamp so that the window manager accepts it and sends that explicit focus event. It also explains why the bug became so much rarer after it.

ilya-fedin:
> * On the QtGui side nothing clears `focus_window` when the window hides. The only thing that eventually does is the timer that `QXcbWindow::doFocusOut()` starts when the FocusOut arrives (`qxcbwindow.cpp:854`), with an interval of 100 ms (`qxcbconnection.cpp:92`).

Claude was telling me the same theory but it doesn't fit with this:

> The same desynchronization also happens the other way around: the window stayed `active` for four minutes while the focus belonged to another application, which is what makes the tray icon click hide the window instead of showing it.

A 100ms timer couldn't cause a 4 minute desynchronization.

ilya-fedin:
Only maybe that the 4 minute desynchronization is a separate bug

ilya-fedin:
> 15:57:17.672  FocusIn   NotifyPointer     focus reverted to PointerRoot, dropped by the plugin

And xev doesn't show explicit focus after that? Claude says KWin has a guard against PointerRoot focus (src/events.cpp:206-222) but you may need a record with `xev -root` to see that

kost-t-human:
> > 15:57:17.672  FocusIn   NotifyPointer     focus reverted to PointerRoot, dropped by the plugin
> 
> And xev doesn't show explicit focus after that? Claude says KWin has a guard against PointerRoot focus (src/events.cpp:206-222) but you may need a record with `xev -root` to see that

It does, next line of the same trace: 
`15:57:17.708 FocusIn NotifyNonlinear`,
36 ms after the pointer one. So the explicit focus does come and the plugin takes
it. The problem is that `focus_window` still points at our window by then, so
`processFocusWindowEvent` returns early and `active_window` stays null.

About the kwin guard, you are right. Recorded both windows at once on a hide to
tray, with the mouse over the client window. Root window:

```
21:26:29.184  FocusIn   NotifyPointerRoot
21:26:29.184  FocusIn   NotifyPointer
```

Client window, same hide:

```
21:26:29.223  FocusOut  NotifyNonlinear
21:26:29.223  FocusIn   NotifyPointer
```

So the focus does land on PointerRoot, and while it sits there the pointer chain
gets the NotifyPointer half of it, root and client both, which is the half the
plugin drops. The explicit focus comes back when the window is shown again. Two
separate blocks because the timestamps come from two xev processes, exact within
each stream but not across them.

kost-t-human:
> Only maybe that the 4 minute desynchronization is a separate bug

That one is my measurement error. The probe compared the client against _NET_ACTIVE_WINDOW and _NET_WM_STATE_FOCUSED, which is the wm idea of the active window, not the input focus. Checked the logs i still have from that day, longest such state there is 3 seconds. Removing that sentence from the description

ilya-fedin:
> The explicit focus comes back when the window is shown again.

Do you mean KWin's PointerRoot guard doesn't work until the window is shown?

> The teardown path at least notices the case: `QXcbWindow::destroy()` calls `doFocusOut()` when the window being destroyed is the focus window (`qxcbwindow.cpp:530`). That still goes through the same timer, and `hide()` does not do even that much.

That's a very good observation. It means we don't need private Qt API to fix the bug, we could just call `destroy();`.

kost-t-human:
> Do you mean KWin's PointerRoot guard doesn't work until the window is shown?

No, KWin does not leave the focus on PointerRoot while the window is hidden. Polled `XGetInputFocus` every 2 ms across a hide to tray, and it does land on PointerRoot, but the next window takes it about 5 ms later:

```
22:45:14.934  0x60001c      telegram, active
22:45:16.437  PointerRoot   hidden to tray
22:45:16.442  0x500000d     5 ms later the next window has the focus
22:45:26.458  0x60001c      shown back 10 s later
```

Two runs, 5 ms and 6 ms. The client does get its `FocusIn NotifyPointer` out of it, it just does not last. Which code moves the focus on, your guard or the ordinary activate the next client on unmap, i did not instrument. Either way it does not affect this PR, the race does not depend on where the focus goes while the window is hidden.

> we could just call destroy();

Tested it in the plain Qt repro and it does fix it, though not through the line i quoted. `QXcbWindow::destroy()` only starts the same 100 ms timer, and a show inside those 100 ms cancels it again. What fixes it is `QWindowPrivate::destroy()` one level up, `focus_window = q->parent()`, synchronous, null for a top level, same in 6.4.2 and in dev.
But `destroy()` drops the X window and its surface on every hide to tray, `SurfaceAboutToBeDestroyed` goes out and the winId changed in my run. So anything holding the old id or a resource tied to that surface has to be rebuilt on the next show. `handleFocusWindowChanged(nullptr)` runs the same deactivation path the plugin's own timer runs (`qxcbconnection.cpp:97)`, just immediately and with the native window untouched.

ilya-fedin:
> though not through the line i quoted

Yeah, I know, but your observation remembered me of this path.

> So anything holding the old id or a resource tied to that surface has to be rebuilt on the next show. ``

Yeah, all the code operating on native handle is made in a way to not cache winId on Linux, re-applying on each show so it should be ok. FWIW, the most minimalist patch is only 3 lines:
```
diff --git a/Telegram/SourceFiles/platform/linux/main_window_linux.cpp b/Telegram/SourceFiles/platform/linux/main_window_linux.cpp
index abf764ba2d..dd644f281f 100644
--- a/Telegram/SourceFiles/platform/linux/main_window_linux.cpp
+++ b/Telegram/SourceFiles/platform/linux/main_window_linux.cpp
@@ -522,6 +522,9 @@ bool MainWindow::eventFilter(QObject *obj, QEvent *evt) {
 		}
 	} else if (obj == this && t == QEvent::Hide) {
 		_exposed = false;
+		if (IsX11()) {
+			destroy();
+		}
 	} else if (obj == this && t == QEvent::ThemeChange) {
 		updateWindowIcon();
 	}
```

kost-t-human:
Built dev  + #31189 +  your three lines and ran the reproduction on the client. It is gone, three runs, both the 600 ms and the 30 ms cycle stay active.

It does segfault though. Any Move or Resize of the main window calls `positionUpdated()` (`mainwindow.cpp:698`), which arms `_positionUpdatedTimer` for a second (`main_window.cpp:795`). When it fires it calls `savePosition()` with the default `Qt::WindowActive`, so it takes this branch, and the deref happens before the `isVisible()` guard:

```cpp
// main_window.cpp:902
if (state == Qt::WindowActive) {
    state = windowHandle()->windowState();
}
```

After `destroy()` there is no handle. So move the window and hide it to the tray within that second and the client is gone:

```
#0  QWindow::windowState() const ()
#1  Window::MainWindow::savePosition(Qt::WindowState) ()
#2  base::Timer::timerEvent(QTimerEvent*) ()
#3  QObject::event(QEvent*) ()
```

Both scripted runs died with SIGSEGV, a gdb session caught the backtrace above, and it also reproduces by hand: drag the window a little, then click the tray icon quickly, or hit the close button. Everything else held up. The same test on the current PR build does not do it.

ilya-fedin:
Just reorder it?
```
diff --git a/Telegram/SourceFiles/window/main_window.cpp b/Telegram/SourceFiles/window/main_window.cpp
index 72ef9dd858..f3b79a1fd5 100644
--- a/Telegram/SourceFiles/window/main_window.cpp
+++ b/Telegram/SourceFiles/window/main_window.cpp
@@ -939,13 +939,14 @@ QRect MainWindow::computeDesktopRect() const {
 }
 
 void MainWindow::savePosition(Qt::WindowState state) {
+	if (!isVisible() || !positionInited()) {
+		return;
+	}
+
 	if (state == Qt::WindowActive) {
 		state = windowHandle()->windowState();
 	}
-
-	if (state == Qt::WindowMinimized
-		|| !isVisible()
-		|| !positionInited()) {
+	if (state == Qt::WindowMinimized) {
 		return;
 	}
 	if (const auto saved = Core::App().savedWindows()) {
```

kost-t-human:
Yeah, that does it. Rebuilt with both of your diffs and the crash is gone: move the window, hide it to the tray inside that second, three runs, then three more with the close button, all fine. Desync is still fixed too, the 600 ms and the 30 ms cycle both stay active.

Went looking for anything else that touches the handle while the window is hidden. `markingAsRead()` (`mainwindow.cpp:602`) sits behind `!isHidden()`, and the top bar (`history_view_top_bar_widget.cpp:278`) and the connecting widget (`window_connecting_widget.cpp:311`) already null check it themselves. Only `workmodeUpdated()` (`main_window_linux.cpp:142`) derefs `windowHandle()` unguarded, but the work mode only changes from the settings section, so the window is up by then.

Rest of it held over 30 cycles, geometry and maximized state come back, the position still reaches the settings across a restart, rss and fds flat, and nothing minded the winId changing. Both of your diffs are in the PR now, the private API patch is gone.

ilya-fedin:
Nice, can you separate them to two commits? And I would say the PR is ready for merge.

kost-t-human:
Split. Thanks for the patience with this one

ilya-fedin:
I remember the situation with focus being very buggy on X11. I was even experimenting with calling `base::Platform::ActivateThisProcessWindow` in `MainWindow::activate` on MATE multiple years ago, it was helping back then too with activation from tray but I wasn't sure it's the right thing to do and didn't think of putting it into ActivateThisProcess. It was looking quite hacky inside `activate` so I just gave up on it. Qt was always buggy on platform integration so I wasn't denying the possibility that it's just buggy X11 implementation in Qt. After all, I think we can say that's exactly what happened. If not that weird timer inside Qt + if Qt was doing the following inside SNI and D-Bus menu impls:
```
    if (QX11Info::isPlatformX11())
        QX11Info::setAppTime(QX11Info::getTimestamp());
```
We would need neither #31189 nor this PR.

kost-t-human:
Yeah. Waiting for Wayland here anyway, once everything i run is fine on it :)

PR #31188, opened 2026-08-27
Follow the scroll direction in swipe gestures

### LINKED ISSUE #29573: Touchpad gestures on Linux inconsistent with other software

ilya-fedin:
`QWheelEvent::inverted()` was checked at the wrong level.

Swipe gestures were normalized to the physical finger movement, so that the same gesture would need the same finger direction no matter how the user configured scrolling. That is only possible where the platform reports the physical direction: macOS, and Wayland compositors sending `wl_pointer.axis_relative_direction` (wl_pointer v9). Where it is not reported — X11 and XWayland always, GNOME before 49, Sway before 1.10 — `inverted()` is `false` for a touchpad with natural scrolling on, and every gesture ends up reversed. That is what #29573 is about.

The way out is not to guess the finger direction better, but to stop needing it for the gesture itself. GTK never asks for it: `AdwSwipeTracker`, which drives swipe-back in `AdwNavigationView` and `AdwLeaflet`, uses the raw scroll deltas, and GTK exposes the physical direction only behind the opt-in `GTK_EVENT_CONTROLLER_SCROLL_PHYSICAL_DIRECTION` flag, which libadwaita never sets. WebKit does the same on both GTK and macOS. Their gestures follow the scroll direction, so they are automatically consistent with the user's setting and cannot be reversed by a compositor that keeps quiet.

So the delta goes back to what it was before 16825fff41fe51211245629c77d6779ee8de3f6b, and `inverted()` moves to where it belongs — the visual effects:

* **Gesture direction follows the scroll**, like everywhere else on the desktop. This is what removes the dependency on the compositor: the natural scrolling setting is already baked into the deltas by the compositor, so nothing has to be recovered.
* **Content follows the finger**, through the new `SwipeContextData::visualTranslation()`. The reply bubble, its icon and the chat list row move with the finger, mirrored when needed, instead of the gesture being mirrored to keep them in place. The `translation` sign keeps meaning which of the two actions was started, so nothing that branches on it had to change.
* **Gestures that drag out something anchored to a side of the window keep following the finger**, through the new `SwipeHandlerInitData::fingerDirection()`. The main menu lives past the left edge, so it is always pulled to the right and pushed back to the left; the chats list quick action shares the axis with it and stays put as well. Where the main menu waits in the row of chats filters flips instead — past the first filter with natural scrolling, past the last one without it — so that the filters stay reachable in both directions.

Everything else — moving between the chats filters, tabs, photos in the media viewer, going back — follows the scroll now.

Fixes #29573
Closes #31177

### DIFF
--- a/Telegram/SourceFiles/dialogs/dialogs_widget.cpp
+++ b/Telegram/SourceFiles/dialogs/dialogs_widget.cpp
@@ -811,19 +811,27 @@ Widget::Widget(
 }
 
 void Widget::setupSwipeBack() {
-	const auto isMainList = [=] {
+	// The main menu is dragged out from the left side of the window, so it
+	// always waits past the end of the chats filters, at whichever of them
+	// the swipe towards it scrolls to nothing - the first one with the
+	// natural scrolling and the last one without it. Standing anywhere else
+	// in the filters that swipe still has a filter to move to.
+	const auto noNearChatsFilter = [=](bool isNext) {
 		const auto current = controller()->activeChatsFilterCurrent();
 		const auto &chatsFilters = session().data().chatsFilters();
-		if (chatsFilters.has()) {
-			return chatsFilters.defaultId() == current;
+		if (!chatsFilters.has()) {
+			return !current;
 		}
-		return !current;
+		return !Window::CheckAndJumpToNearChatsFilter(
+			controller(),
+			isNext,
+			false);
 	};
 
 	auto update = [=](Ui::Controls::SwipeContextData data) {
 		data.cursorTop -= _inner->y();
 		if (data.translation != 0) {
-			if (data.translation < 0
+			if (data.visualTranslation() < 0
 				&& _inner
 				&& (Core::App().settings().quickDialogAction()
 					!= Ui::QuickDialogAction::Disabled)) {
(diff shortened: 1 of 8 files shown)


wlroots:
Tested the artifact on Sway, works completely fine out of the box now! Thanks for the clean fix.

ilya-fedin:
> Tested the artifact on Sway, works completely fine out of the box now! Thanks for the clean fix.

It should still be bugged on compositors without the protocol support (e.g. swiping a message to left for reply will move it to right) but it becomes just a visual issue instead of a functional one. The only thing is I'm not sure what happens with the swipe on chat list, maybe switching folders or opening the sidebar becomes broken... but either way I assume it's a smaller issue than the current state.

PR #354, opened 2014-12-23
Added Korean Language(Full Translated)

itstake:

PR #31191, opened 2026-08-27
Expose value-picking sliders as a slider and a radio group

rezabakhshilaktasaraei:
Two of the DiscreteSlider strips do not switch displayed content - they pick a value - so with #31097 in place they can get their real roles:

- The notifications count (Settings > Notifications and Sounds) picks a number of 1..5. It is now a single slider element: the arrows change the value right away and a screen reader reads and follows the number, the same way the volume slider works.
- The proxy auto-switch timeout picks one of a few textual durations. It is now read as a radio group: Left / Right check the option the arrows land on, matching how the Radiobutton groups behave (a horizontal group leaves Up / Down free).

All other strips keep the tab roles from #31097.

Depends on #31097 and on desktop-app/lib_ui#354 (the accessible value interface), so it compiles only on top of them.

### DIFF
--- a/Telegram/SourceFiles/boxes/connection_box.cpp
+++ b/Telegram/SourceFiles/boxes/connection_box.cpp
@@ -1225,6 +1225,10 @@ void ProxiesBox::setupContent() {
 			_proxyRotationOptions->entity(),
 			st::settingsSlider),
 		st::settingsBigScalePadding);
+	// Picks one of a few textual timeout options, not a displayed section -
+	// a radio group to accessibility.
+	_proxyRotationTimeout->setAccessibilityMode(
+		Ui::DiscreteSlider::AccessibilityMode::Radio);
 	for (const auto seconds : Core::SettingsProxy::kProxyRotationTimeouts) {
 		_proxyRotationTimeout->addSection(
 			tr::lng_proxy_auto_switch_timeout(
--- a/Telegram/SourceFiles/settings/sections/settings_notifications.cpp
+++ b/Telegram/SourceFiles/settings/sections/settings_notifications.cpp
@@ -1567,6 +1567,9 @@ void BuildSystemIntegrationAndAdvancedSection(SectionBuilder &builder) {
 		const auto countSlider = advancedWrap->add(
 			object_ptr<Ui::SettingsSlider>(advancedWrap, st::settingsSlider),
 			st::settingsBigScalePadding);
+		// Picks a count, not a displayed section - a slider to accessibility.
+		countSlider->setAccessibilityMode(
+			Ui::DiscreteSlider::AccessibilityMode::Value);
 		for (int i = 0; i != kMaxNotificationsCount; ++i) {
 			countSlider->addSection(QString::number(i + 1));
 		}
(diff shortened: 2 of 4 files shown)


john-preston:
**Verdict: CHANGES_REQUIRED.** Keep the two roles. Validate incoming numeric requests, notify value changes through every activation path. The missing numeric event payload in lib_ui #354 is a separate prerequisite correction.

**P2 — Reject invalid numeric requests before choosing a section.** The queued [setter at lines 408–416](https://github.com/telegramdesktop/tdesktop/blob/9d27388746da84cf39434d96084f0c154d41bc76/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp#L408-L416) checks only mode and section presence. Starting from count 3, send an accessible value of 99: [nearest-section mapping](https://github.com/telegramdesktop/tdesktop/blob/9d27388746da84cf39434d96084f0c154d41bc76/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp#L430-L441) selects 5 and persists that setting. A negative request selects 1. NaN or either infinity leaves the first candidate selected, also changing the count to 1.

[Qt's public setter contract](https://doc.qt.io/qt-6/qaccessiblevalueinterface.html#setCurrentValue) requires out-of-range requests to be ignored. The configured Qt 5.15.19 AT-SPI implementation forwards the double directly (`src/platformsupport/linuxaccessibility/atspiadaptor.cpp:2239`). Its Windows RangeValue provider rejects ordinary out-of-range numbers, but its two comparisons do not reject NaN (`qwindowsuiarangevalueprovider.cpp:65–82`). The adapter in #354 does not validate either. Thus this cannot rely on a universal backend precheck. Inside the guarded main-thread callback, reject non-finite values and values outside the current advertised range before quantizing valid input. Preserve the valid in-range selection behavior.

**P2 — Emit value notifications for mouse activation too.** [changeValueTo at lines 376–381](https://github.com/telegramdesktop/tdesktop/blob/9d27388746da84cf39434d96084f0c154d41bc76/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp#L376-L381) is the only value-notification call, and only the new keyboard/setter paths use it. With the count currently 2, click the section labelled 4. [mouseReleaseEvent](https://github.com/telegramdesktop/tdesktop/blob/9d27388746da84cf39434d96084f0c154d41bc76/Telegram/SourceFiles/ui/widgets/discrete_sliders.cpp#L233-L246) calls setActiveSection directly. Its activation stream reaches [position->setCount](https://github.com/telegramdesktop/tdesktop/blob/9d27388746da84cf39434d96084f0c154d41bc76/Telegram/SourceFiles/settings/sections/settings_notifications.cpp#L1573-L1580), updating and saving the setting, but emits no QAccessibleValueChangeEvent. A subscribed client receives no value event even though a later query returns 4. [Slider event requirements](https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-supportslidercontroltype#required-events) apply when the supported value changes, not only when a keyboard or provider caused it.

Move the mode-specific notification into the shared active-value change path, with an actual-change check, and remove the duplicate wrapper notification. Preserve existing callback timing and tab/radio announcement policy. This omission remains after #354 is fixed: that prerequisite repairs an event's payload, whereas this path sends no event at all. No speech result was experimentally measured.

PR #31194, opened 2026-08-28
Fix Linux Mini App window shadow gutter

edwardgushchin:
## Summary

- remove the dedicated visual shadow gutter around Linux external Mini App windows;
- keep an independent transparent 8 px hit target for all edge and corner resize handles;
- retain a subtle 1 px inset outline around the actual shell surface.

## Problem

The external Linux WebView shell reserves client-side margins of 14 px left/right, 12 px top and 16 px bottom solely for a multi-layer HTML shadow. On GNOME Wayland this renders as a prominent dark rectangular frame around the Mini App instead of a native-looking window boundary.

The shadow padding is also used as the resize target, so simply zeroing the margins would otherwise make border resizing unavailable. This change separates the invisible resize hit target from the visual padding.

For the current `botWebViewPanelSize` (384 x 694) and 60 px header, the observed helper window changes from 412 x 782 to 384 x 754 while preserving the content area.

## Testing

- built the complete Telegram Desktop 7.0.9 Release package on CachyOS Linux / GNOME 50.4 / Wayland;
- installed and launched the patched package; the external Mini App helper restarted normally;
- verified live window extents through AT-SPI: 412 x 782 before, 384 x 754 after;
- verified all eight edge/corner handles use the independent 8 px hit target;
- ported the single commit onto the current `upstream/dev`; it applies without conflicts and `git diff --check` passes.

### DIFF
--- a/Telegram/Resources/bot_webview_shell_html/page.css
+++ b/Telegram/Resources/bot_webview_shell_html/page.css
@@ -33,10 +33,11 @@ select {
 	--shell-pad-right: 12px;
 	--shell-pad-bottom: 12px;
 	--shell-pad-left: 12px;
-	--shadow-pad-top: 12px;
-	--shadow-pad-right: 14px;
-	--shadow-pad-bottom: 16px;
-	--shadow-pad-left: 14px;
+	--shadow-pad-top: 0px;
+	--shadow-pad-right: 0px;
+	--shadow-pad-bottom: 0px;
+	--shadow-pad-left: 0px;
+	--resize-handle: 8px;
 	--header-height: 60px;
 	--title-pad-top: 0px;
 	--title-pad-right: 0px;
@@ -95,11 +96,7 @@ select {
 	flex-direction: column;
 	border-radius: var(--shell-radius);
 	background: var(--body-bg, #ffffff);
-	box-shadow:
-		0 0 0 1px rgba(0, 0, 0, 0.08),
-		0 1px 2px rgba(0, 0, 0, 0.18),
-		0 3px 8px -1px rgba(0, 0, 0, 0.16),
-		0 8px 14px -5px rgba(0, 0, 0, 0.20);
+	box-shadow: inset 0 0 0 1px rgba(0, 0, 0, 0.10);
 	overflow: hidden;
 }
 #root.no-window-alpha #shell {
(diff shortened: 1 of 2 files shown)


ilya-fedin:
> On GNOME Wayland this renders as a prominent dark rectangular frame around the Mini App instead of a native-looking window boundary.

I don't quite understand what is meant here. Could you provide a screenshot?

edwardgushchin:
Sure — here is the affected external Mini App window on GNOME Wayland before the patch. The dark rectangular area surrounding the Mini App is the application-owned HTML shadow gutter (14 px left/right, 12 px top, and 16 px bottom):

telegram-mini-app-shadow-gutter-before

ilya-fedin:
Can you explain why it doesn't work the current way? What happens to telling the system about shadow size via `gdk_window_set_shadow_width` after this change? Are the hints seen by the system still the same as the shadow being rendered?

edwardgushchin:
You’re right: the current implementation is internally consistent. The 14/12/14/16 px HTML shadow margins are passed to lib_webview as windowMargins and then to gdk_window_set_shadow_width() or its GTK 4 equivalent. Therefore, the compositor is already told the same extents that are being rendered.

With this patch, both the rendered outer shadow and the corresponding system hints become zero. The remaining 1 px outline is inset, while the 8 px resize targets are inside the client surface. Therefore, there is no mismatch, but the patch does not delegate shadow rendering to the compositor — it simply removes Telegram’s client-rendered shadow.
The issue I was trying to address is visual: on GNOME Wayland, this HTML shadow appears as a heavy rectangular gutter around the window.

Perhaps the better design question is whether Telegram should render its own HTML shadow for this frameless external shell at all, or leave any outer shadow to the compositor/platform. If preserving the client-side shadow is preferred, I can instead rework its rendering while keeping the existing GDK extents aligned with it.

ilya-fedin:
> or leave any outer shadow to the compositor/platform or leave any outer shadow to the compositor/platform

But you can't make the platform draw only shadow but not the frame on Linux. AFAIK only mutter and it forks do that and only on X11. And then you can't resize with the shadow it draws, you have to draw a border inside your window anyway. With other WMs on X11 (be it kwin, openbox or something else) and with any compositor on Wayland once you remove system window frame, no shadow is drawn by the system. And then GNOME doesn't support drawing window frame and shadow on Wayland at all AFAIK, it forces CSD on all apps (hence when you enable Qt window frame in tdesktop settings, you get an Adwaita-looking frame drawn by Qt rather than real frame from your GTK theme like it happens on X11).

edwardgushchin:
Thanks, understood. Then delegating only the shadow to the compositor is not a portable option for this frameless window, and the client has to own both the shadow and the resize affordance.
In that case, do you consider the appearance shown in the screenshot to be expected? If not, would a change that preserves the existing client-side shadow margins and GDK hints, but only adjusts the CSS shadow rendering, be an acceptable direction?

ilya-fedin:
> In that case, do you consider the appearance shown in the screenshot to be expected?

No, it's absolutely not expected and I couldn't even reproduce it. Here's how it looks like on my end:

edwardgushchin:
Thanks — I found the cause on my side.

My `~/.config/gtk-4.0/gtk.css` contained an unscoped rule:

```css
.background {
  color: @window_fg_color;
  background-color: color-mix(in srgb, @window_bg_color 96%, transparent);
}
```

GTK loads user CSS with [`GTK_STYLE_PROVIDER_PRIORITY_USER`](https://docs.gtk.org/gtk4/const.STYLE_PROVIDER_PRIORITY_USER.html) (800), which overrides lib_webview's application-level transparent background CSS ([`GTK_STYLE_PROVIDER_PRIORITY_APPLICATION`](https://docs.gtk.org/gtk4/const.STYLE_PROVIDER_PRIORITY_APPLICATION.html), 600). As a result, the normally transparent shadow margins were filled with an almost opaque `#323232` background. The pixel values in my screenshot match that 96% blend exactly.

After removing this rule and reinstalling the unmodified distribution package, the Mini App renders correctly with the upstream `412x782` geometry, transparent margins, and the intended shadow, matching your screenshot.

So this PR was masking a local GTK override by removing the intended shadow globally, rather than fixing a Telegram Desktop bug. Sorry for the noise, and thank you for checking. I'm closing the PR.

PR #31204, opened 2026-08-29
Release audio device when an active track is destroyed

### LINKED ISSUE #31203: Audio stream remains active after an incoming call ends elsewhere
## Steps to reproduce

1. Run Telegram Desktop and inspect its PipeWire streams with `wpctl status` or qpwgraph.
2. Receive an incoming Telegram call.
3. Before answering it in Desktop, either have the calling party hang up, or answer it on another device logged into the receiving account and then end the call.
4. Wait at least two seconds after Desktop has returned to its normal UI and inspect PipeWire again.

## Expected behaviour

Telegram Desktop should release its playback stream after the call ends. PipeWire should no longer report a running Telegram `Stream/Output/Audio` node.

## Actual behaviour

On the unmodified build, both terminal paths left a running OpenAL Soft playback stream until Telegram Desktop was restarted:

```text
state: "running"
media.class = "Stream/Output/Audio"
media.name = "Playback Stream"
media.role = "Game"
node.name = ".Telegram-wrapped"
node.always-process = "true"
node.loop.name = "ALSoftP0"
stream.is-live = "true"
```

A self-built Telegram Desktop 6.8.1 with the proposed fix was tested with both affected paths. In each case the stream was released.

## Operating system

NixOS 26.05 (Yarara), KDE Plasma 6.6.6, Wayland, PipeWire 1.6.6, WirePlumber 0.5.14.

## Version of Telegram Desktop

6.8.1

## Installation source

Other source (self-built from upstream 6.8.1 sources)

## Crash ID

Not applicable.

## Logs

No crash or relevant Telegram log error observed.

Related: #28628, #26386, #1044

MaxMatti:
Fixes #31203

Destroying an active `Media::Audio::Track` bypassed `trackFinished()`, so the audio instance did not schedule its idle device detach. Finish the track before it is detached and unregistered.

Tested with a self-built 6.8.1 binary on NixOS, KDE Plasma, and PipeWire:

- calling party hangs up while Desktop is ringing;
- recipient answers on another device, then ends the call.

The lingering OpenAL Soft playback stream was released in both cases.

### DIFF
--- a/Telegram/SourceFiles/media/audio/media_audio_track.cpp
+++ b/Telegram/SourceFiles/media/audio/media_audio_track.cpp
@@ -244,6 +244,7 @@ void Track::reattachToDevice() {
 }
 
 Track::~Track() {
+	finish();
 	detachFromDevice();
 	_instance->unregisterTrack(this);
 }


john-preston:
Thanks!

PR #31218, opened 2026-09-01
Leave the unread bar out of the message count for screen readers

rezabakhshilaktasaraei:
Depends on #31211 and builds on top of it: that PR exposes the unread bar as a separator, this one keeps it out of the "x of y" a screen reader announces for the messages around it.

With desktop-app/lib_ui#359 a painted list item reports its position within its set, by default its index among all rows. The unread bar is not a message, so both message lists override that: the bar itself reports no position, and the messages are numbered without counting it, so the first unread message is "5 of 20" whether or not the bar sits above it.

Needs desktop-app/lib_ui#359 (and desktop-app/patches#265 for the Qt side of it).

Tested with NVDA on Windows 10: arrowing through a chat with unread messages announces every message with its number and the total, the numbers do not skip across the bar, and the bar is still announced as "Unread messages, separator" with no position of its own.

### DIFF
--- a/Telegram/SourceFiles/history/history_inner_widget.cpp
+++ b/Telegram/SourceFiles/history/history_inner_widget.cpp
@@ -6635,6 +6635,20 @@ QAccessible::Role HistoryInner::accessibilityChildRoleAt(int index) const {
 		: accessibilityChildRole();
 }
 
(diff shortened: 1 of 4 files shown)


john-preston:
Fix the conflicts please.

PR #31237, opened 2026-09-03
Let Tab reach the settings and buttons of the poll and checklist boxes with a screen reader

rezabakhshilaktasaraei:
With a screen reader, Tab could not leave the text fields of the Create Poll box: every field passed Tab on to the next one and back to the first, so the settings (Show Who Voted, Set Correct Answer, …), the correct answer buttons, the attach and emoji buttons and Create/Cancel were reachable only by object navigation. The checklist, "Edit your name" and "Add link" boxes had the same ring.

- The fields leave a Tab request for the default order alone (see desktop-app/lib_ui#361), so in screen reader mode Tab walks each box in the order it is shown. Without a screen reader the rings behave as before.
- The poll content, the answers list and the checklist content order their Tab stops by position — rows created while the box is open landed at the end of the focus chain, past the settings.
- The emoji toggle of a field stays shown while the keyboard is in the field (its text, the attach button inside it, or the toggle itself) instead of fading out under the focus that arrives on it; it also gets the name "Emoji". Note: this applies to every box using `AddEmojiToggleToField`, not only the poll box.
- Names: the poll media button ("Add attachment"), the checklist task remove button ("Remove"), and the correct answer radio buttons/checkboxes of a quiz (the text of their answer, kept up to date). A task removed with its button from the keyboard passes the focus on to the next task.

Depends on desktop-app/lib_base#291 and desktop-app/lib_ui#361. Testing the attach menu path additionally needs desktop-app/lib_ui#363.

### DIFF
--- a/Telegram/SourceFiles/boxes/add_contact_box.cpp
+++ b/Telegram/SourceFiles/boxes/add_contact_box.cpp
@@ -1579,14 +1579,22 @@ void EditNameBox::prepare() {
 	_last->submits(
 	) | rpl::on_next([=] { submit(); }, _last->lifetime());
 
+	// Tab switches between the two names. A request for the default order
+	// is left alone: the buttons of the box are Tab stops then.
 	using TabbedRequest = Ui::InputField::TabbedRequest;
 	_first->tabbed(
 	) | rpl::on_next([=](not_null<TabbedRequest*> request) {
+		if (request->defaultOrder) {
+			return;
+		}
 		_last->setFocus();
 		request->handled = true;
 	}, _first->lifetime());
 	_last->tabbed(
 	) | rpl::on_next([=](not_null<TabbedRequest*> request) {
+		if (request->defaultOrder) {
+			return;
+		}
 		_first->setFocus();
 		request->handled = true;
 	}, _last->lifetime());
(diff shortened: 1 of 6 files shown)


john-preston:
**Verdict: CHANGES_REQUIRED.** Keep the navigation implementation. Remove one redundant quiz-name event. No substantive focus-order regression was established. Coordinate the emoji opener with the small field-focus fix already in #31240; the complete combined stack must still be exercised. This is static review, not runtime certification.

## Findings

**P3 — Remove the second quiz NameChanged event.** The [answer subscription](https://github.com/telegramdesktop/tdesktop/blob/40e23aa282f2de1c3bbd2a84dc8a53a2f42ded37/Telegram/SourceFiles/boxes/create_poll_box.cpp#L669-L674) calls `setAccessibleName()` immediately followed by `accessibilityNameChanged()`. Create a quiz, then edit an answer: each subscription emission sends the same name event twice. Configured Qt 5.15.19's QWidget setter already constructs and dispatches NameChanged at `qwidget.cpp:11409–11414`; the [RpWidget helper](https://github.com/desktop-app/lib_ui/blob/0f8abea17431d2f98ac41e48eca1266cca66664a/ui/rp_widget.cpp#L497-L500) repeats that operation. This confirms duplicate source notifications, not duplicate spoken output. Remove the explicit helper call; retain the initial name and live answer subscription. The setter handles the notification.

PR #31240, opened 2026-09-04
Walk the emoji panel with the keyboard and a screen reader

rezabakhshilaktasaraei:
The emoji panel is a painted grid with no keyboard handling and nothing a screen reader can read: opened from the keyboard, it appeared with no one to talk to, and pressing the "Emoji" button of a field seemed to do nothing.

- `EmojiListWidget` exposes every shown cell as an item of a list (the existing painted-list accessibility API): named by the emoji itself — the screen reader speaks it in its own words and language — described by its section (the category, or the title of a custom set); the "+N" cell of a collapsed set opens it.
- In screen reader mode the list takes the focus once the panel is shown and remembers where it came from. Left/Right, Up/Down (by column, on into the next section), Home/End, PageUp/PageDown walk the cells. Enter or Space chooses, Escape gives up; both return the focus to where it was and ask the panel to hide (`TabbedSelector::cancelled()` passes the request on, like the GIF tab's cancel). SetFocus/Invoke from the screen reader work the same way. A keyboard selection survives the mouse leaving the list.
- The "Emoji" toggle of a field gives the focus to its field before opening the panel: the fields of a box tell the target of a chosen emoji apart by the focus, which a toggle pressed from the keyboard held itself. With the mouse the field has it already.

Nothing changes without a screen reader. Standalone, no library changes needed. Verified with NVDA: emoji names, section names, all keys, insertion and focus return.

### DIFF
--- a/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
+++ b/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
@@ -29,6 +29,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 #include "ui/emoji_config.h"
 #include "ui/painter.h"
 #include "ui/power_saving.h"
+#include "ui/screen_reader_mode.h"
 #include "ui/ui_utility.h"
 #include "ui/cached_round_corners.h"
 #include "boxes/share_box.h"
@@ -1696,12 +1697,26 @@ void EmojiListWidget::afterShown() {
 		|| (_mode == Mode::UserpicBuilder);
 	if (_search && steal) {
 		_search->stealFocus();
+	} else if (Ui::ScreenReaderModeActive() && !hasFocus()) {
+		// The panel is opened from the keyboard and has no one to talk
+		// to: take the focus into the list, so the emoji are announced
+		// and walked with the arrows, and remember where it came from.
+		_focusReturn = window()->focusWidget();
+		setFocus();
 	}
 }
 
 void EmojiListWidget::beforeHiding() {
 	if (_search) {
 		_search->returnFocus();
+	}
+	returnFocus();
+}
+
+void EmojiListWidget::returnFocus() {
+	const auto was = base::take(_focusReturn);
+	if (was && hasFocus()) {
+		was->setFocus();
 	}
 }
 
@@ -3401,15 +3416,21 @@ void EmojiListWidget::mouseMoveEvent(QMouseEvent *e) {
 			_picker->clearSelection();
 		}
 	}
+	_keyboardSelection = false;
 	updateSelected();
 }
 
 void EmojiListWidget::leaveEventHook(QEvent *e) {
-	clearSelection();
+	// A selection made from the keyboard stays where the mouse is not.
+	if (!_keyboardSelection) {
+		clearSelection();
+	}
 }
 
 void EmojiListWidget::leaveToChildEvent(QEvent *e, QWidget *child) {
-	clearSelection();
+	if (!_keyboardSelection) {
+		clearSelection();
+	}
 }
 
 void EmojiListWidget::enterFromChildEvent(QEvent *e, QWidget *child) {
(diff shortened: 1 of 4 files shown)


john-preston:
**CHANGES_REQUIRED.** Keep the painted-grid implementation, with five functional corrections.

## Required corrections

**P2 — Restore the destination field for mouse selection too.** [The new afterShown focus takeover](https://github.com/telegramdesktop/tdesktop/blob/2818a1d4e3ca4accc08f916014d1128bf1405892/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L1694-L1706) runs for mouse-opened panels as well as keyboard-opened panels whenever reader mode is active. Open the poll question's emoji panel and click a plain emoji. [Mouse release still selects directly](https://github.com/telegramdesktop/tdesktop/blob/2818a1d4e3ca4accc08f916014d1128bf1405892/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L2958-L2978), while [only keyboard activation restores the field first](https://github.com/telegramdesktop/tdesktop/blob/2818a1d4e3ca4accc08f916014d1128bf1405892/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L3749-L3771). The [poll question handlers](https://github.com/telegramdesktop/tdesktop/blob/2818a1d4e3ca4accc08f916014d1128bf1405892/Telegram/SourceFiles/boxes/create_poll_box.cpp#L1404-L1416) require `question->hasFocus()`, so they discard the choice. Options and description use the same routing. Prepare destination focus for all committing paths, including mouse skin-tone selection, or use an explicit destination. Preserve repeated mouse insertion rather than forcing keyboard-style closure. This affects mixed mouse/reader use; it is not an initial no-reader mouse regression.

**P2 — Replace positional action identity.** [The identity and reverse lookup](https://github.com/telegramdesktop/tdesktop/blob/2818a1d4e3ca4accc08f916014d1128bf1405892/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L3595-L3614) encode only section number and offset. Retain a provider for an emoji in custom set A, then reorder/remove that set so B occupies its position: the same token now selects B. [Custom rebuild](https://github.com/telegramdesktop/tdesktop/blob/2818a1d4e3ca4accc08f916014d1128bf1405892/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L3913-L4066) follows mutable installed/featured order; recents and search also replace entries at existing positions. Pinned lib_ui's [Item::currentIndex and doAction](https://github.com/desktop-app/lib_ui/blob/90f0acbd9d3896ed631e54c5348a6cb770c706a2/ui/accessible/ui_accessible_item.cpp#L49-L69) deliberately trust stable identity, so an equal slot token defeats invalidation. Use semantic view/section/item keys, or fresh generation tokens that reject stale actions. Reconcile keyboard selection on rebuild too. This finding requires neither 32-bit overflow nor a large section count.

**P2 — Honor live mode deactivation.** [keyPressEvent](https://github.com/telegramdesktop/tdesktop/blob/2818a1d4e3ca4accc08f916014d1128bf1405892/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L3799-L3830) consumes arrows, paging and activation without checking `ScreenReaderModeActive()`. Focus the grid and stop the reader: the pinned FocusManager changes its policy to NoFocus, but configured Qt 5.15.19 `QWidget::setFocusPolicy` leaves existing focus intact. Keys still navigate and insert with project mode inactive. [Queued SetFocus](https://github.com/telegramdesktop/tdesktop/blob/2818a1d4e3ca4accc08f916014d1128bf1405892/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L3617-L3631) likewise lacks an execution-time mode check, and NoFocus does not prohibit explicit `setFocus()`. Gate the new reader-only input/focus paths and define focus return/selection cleanup on deactivation. Metadata exposure is a separate concern.

**P2 — Name Unicode children independently of painting.** [accessibleEmojiText](https://github.com/telegramdesktop/tdesktop/blob/2818a1d4e3ca4accc08f916014d1128bf1405892/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L3499-L3514) calls `lookupOverEmoji`, which reads the lazy `_emoji[section]` vector. The child count already includes every static category through `_counts`, but [ensureLoaded](https://github.com/telegramdesktop/tdesktop/blob/2818a1d4e3ca4accc08f916014d1128bf1405892/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L1899-L1916) populates that vector only from [the clipped paint loop](https://github.com/telegramdesktop/tdesktop/blob/2818a1d4e3ca4accc08f916014d1128bf1405892/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L2517-L2518); [offscreen sections are skipped](https://github.com/telegramdesktop/tdesktop/blob/2818a1d4e3ca4accc08f916014d1128bf1405892/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L2418-L2423). Enumerating an unpainted category therefore produces valid children with empty names. First navigation announces immediately after scrolling, before loading is guaranteed, and loading emits no NameChanged. Resolve category/variant metadata independently of paint, and use the same semantic resolution for actions.

**P2 — Preserve Escape in direct hosts.** [The new Escape branch](https://github.com/telegramdesktop/tdesktop/blob/2818a1d4e3ca4accc08f916014d1128bf1405892/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L3823-L3830) accepts the event after firing `hideRequests()`. Only TabbedSelector subscribes. [ChooseToneIconBox](https://github.com/telegramdesktop/tdesktop/blob/2818a1d4e3ca4accc08f916014d1128bf1405892/Telegram/SourceFiles/boxes/create_ai_tone_box.cpp#L108-L120) and the forum-topic icon box embed this widget directly without that subscription. Tab into one of these grids, or use accessible SetFocus, then press Escape: ordinary Tab entry has no saved `_focusReturn`, the event is swallowed, and the existing layer close handler never receives it. Reactions likewise [forward the old escapes() stream](https://github.com/telegramdesktop/tdesktop/blob/2818a1d4e3ca4accc08f916014d1128bf1405892/Telegram/SourceFiles/history/view/reactions/history_view_reactions_selector.cpp#L1160), which still represents search-field Escape only. Use one cancellation contract for search/grid and wire direct hosts, or preserve propagation where no owner handles cancellation. This persists after live reader deactivation.

PR #31241, opened 2026-09-04
Choose the skin tone of an emoji from the keyboard

rezabakhshilaktasaraei:
On top of #31240 (the emoji panel walked with the keyboard). The skin-tone picker was mouse-only: it comes up right away on the first press of an emoji none of whose variants was ever chosen, and under the button held for a long press (500 ms) after that, while a click inserts with the chosen tone.

- Enter and Space now do the same on an emoji with variants, with the same delay and the same "chosen before" memory. The picker takes the focus with the shown variant selected; it is a list for a screen reader, named by the emoji, its items the variants (Left/Right, Home/End, Enter or Space chooses — tone saved, emoji inserted, focus back to the field, panel hidden — Escape backs out to the same emoji).
- An emoji with variants reports itself `expandable` (`expanded` while its picker is up) and dispatches the standard `showMenu` action to open or close it (desktop-app/lib_ui#365). Platforms whose bridge surfaces the expandable state for list items announce it as collapsed and can open the picker through it; the Windows UIA bridge currently does that for tree items only.

Nothing changes without a screen reader except that Enter/Space in the panel gain the mouse's long-press behaviour. Verified with NVDA.

### DIFF
--- a/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
+++ b/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
@@ -3820,7 +3820,43 @@ QAccessible::State EmojiListWidget::accessibilityChildState(
 			state.focused = true;
 		}
 	}
+	// An emoji with variants opens the picker of its skin tones: a
+	// screen reader hears it as collapsed, and expanded while the picker
+	// is up for it.
+	if (over) {
+		const auto emoji = lookupOverEmoji(&*over);
+		if (emoji && emoji->hasVariants()) {
+			state.expandable = true;
+			const auto picked = std::get_if<OverEmoji>(&_pickerSelected);
+			state.expanded = picked
+				&& (*picked == *over)
+				&& !_picker->isHidden();
+		}
+	}
 	return state;
+}
+
+void EmojiListWidget::accessibilityChildShowMenu(quintptr identity) {
+	// Expand opens the picker of variants for the item, Collapse closes
+	// it - the bridge calls this for either, by the state it sees.
+	crl::on_main(this, [=] {
+		const auto index = accessibilityChildIndexByIdentity(identity);
+		const auto over = accessibleChild(index);
+		if (!over) {
+			return;
+		}
+		const auto picked = std::get_if<OverEmoji>(&_pickerSelected);
+		if (picked && *picked == *over && !_picker->isHidden()) {
+			_picker->hideAnimated();
+			return;
+		}
+		keyboardSelect(*over, false);
+		if (!hasFocus()) {
+			_focusReturn = window()->focusWidget();
+			setFocus();
+		}
+		[[maybe_unused]] const auto opened = openKeyboardPicker();
+	});
 }
 
 bool EmojiListWidget::accessibilityChildSupportsActions(int index) const {
--- a/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
+++ b/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
@@ -29,6 +29,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 #include "ui/emoji_config.h"
 #include "ui/painter.h"
 #include "ui/power_saving.h"
+#include "ui/screen_reader_mode.h"
 #include "ui/ui_utility.h"
 #include "ui/cached_round_corners.h"
 #include "boxes/share_box.h"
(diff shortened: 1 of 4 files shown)


john-preston:
**CHANGES_REQUIRED.** Keep the feature, with four corrections in the new picker/action paths.

## Required corrections

**P2 — Give picker children semantic identities.** [The new identity implementation](https://github.com/telegramdesktop/tdesktop/blob/c9e8099c0d68a9590008a41555d11215d24d5178/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L625-L634) returns `index + 1`, while [showEmoji](https://github.com/telegramdesktop/tdesktop/blob/c9e8099c0d68a9590008a41555d11215d24d5178/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L197-L220) reuses the same widget for different emoji. Retain a provider for a tone of A, dismiss it, and open B: that provider now names and activates B's same-numbered tone. The [lib_ui cache](https://github.com/desktop-app/lib_ui/blob/0988554b104fcdc3364c355e415df8a0a3c847e9/ui/rp_widget.cpp#L917-L935) only replaces a child when its identity changes, so this token defeats invalidation. This is the new picker, independently of corrected grid identities. Use the actual variant identity plus current-membership validation, or a popup-generation token.

**P2 — Preserve the external destination across picker-to-picker transitions.** [showMenu captures the current focus widget](https://github.com/telegramdesktop/tdesktop/blob/c9e8099c0d68a9590008a41555d11215d24d5178/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L3853-L3858) whenever the list lacks focus. Open A's tones from a field, then invoke showMenu for B while the picker owns focus. The saved field becomes the picker itself. When B's tone is chosen, [colorChosen calls returnFocus before emitting](https://github.com/telegramdesktop/tdesktop/blob/c9e8099c0d68a9590008a41555d11215d24d5178/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L3594-L3602), but the return target is already the focused picker; the original field is lost. Focus-sensitive field routing can discard the choice, and dismissal cannot restore that field. Treat grid and picker as one focus session: capture an external destination only on entry from outside, preserve it for internal transitions, and give the picker an explicit list return owner.

**P2 — Extend mode deactivation handling to the new popup.** Open the picker while reader mode is active, then stop the reader or disable that mode. FocusManager changes its focus policy to NoFocus, but configured Qt 5.15.19 `QWidget::setFocusPolicy` leaves existing focus intact. [The new key handler](https://github.com/telegramdesktop/tdesktop/blob/c9e8099c0d68a9590008a41555d11215d24d5178/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L528-L553) continues moving and committing tones. [Its queued SetFocus](https://github.com/telegramdesktop/tdesktop/blob/c9e8099c0d68a9590008a41555d11215d24d5178/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L637-L647) can explicitly focus it again; the new showMenu and [long-press timer callback](https://github.com/telegramdesktop/tdesktop/blob/c9e8099c0d68a9590008a41555d11215d24d5178/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L4140-L4146) also lack execution-time mode checks. Correcting #31240's grid handlers alone does not cover these additions. Apply the shared mode policy to the popup, cancel pending holds on deactivation, and release reader-owned focus while preserving ordinary mouse operation.

**P2 — Reject actions whose popup was dismissed before dispatch completes.** Queue a variant activation while visible, then dismiss it or hide its panel before the callback executes. [The new callback](https://github.com/telegramdesktop/tdesktop/blob/c9e8099c0d68a9590008a41555d11215d24d5178/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L650-L658) only resolves its slot; [hideFast](https://github.com/telegramdesktop/tdesktop/blob/c9e8099c0d68a9590008a41555d11215d24d5178/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L383-L389) and hideAnimated retain `_variants`. It therefore still calls chooseSelected/colorChosen, saving a tone and emitting an emoji after cancellation. A weak lifetime guard covers destruction, not a hidden or dismissing owner. This is separate from deactivation. [The new showMenu callback](https://github.com/telegramdesktop/tdesktop/blob/c9e8099c0d68a9590008a41555d11215d24d5178/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L3839-L3858) likewise has no current-owner visibility/enabled check before changing selection/focus and reopening its child. Validate an active popup generation and usable owner at execution; invalidate pending actions on hide/replacement. Validate variant capability before showMenu changes selection/focus.

PR #31242, opened 2026-09-05
Walk the categories and the search of the emoji panel with the keyboard

rezabakhshilaktasaraei:
On top of #31241. The row of category icons under the emoji list and the groups beside its search field were painted strips with nothing for a screen reader and no keyboard; the search field was reachable, but nothing led from it to the results.

- **Footer** (`StickersListFooter`): a list named "Categories" — Recently used, the seven categories (the "all categories" icon as one item per category), custom emoji sets by title, a group's set by the group's name; on the stickers tab also Favorites, Trending and the "Manage sticker sets" button. The section shown reports as selected. Left/Right, Home/End walk it; Enter or Space scrolls the list to the section and puts the keyboard on its first emoji.
- **Search groups** (`GroupsStrip`): a list named "Search groups", items named by the titles the server sends (`Ui::EmojiGroup` gains a `title` field; it was being dropped). Enter/Space runs the group's search and lands on the first result once results are in; the keyboard keeps its place through a refill of results.
- **Search field**: Enter or Down goes on to the first result (`TabbedSearch::submits()`). The search/back/clear buttons around it get names; the emoji list is named "Emoji".

Two new lang keys: `lng_emoji_categories`, `lng_emoji_search_groups`. Nothing changes without a screen reader except Enter/Down in the panel's search field. Verified with NVDA.

### DIFF
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -3225,6 +3225,8 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 "lng_emoji_category5" = "Travel & Places";
 "lng_emoji_category6" = "Objects";
 "lng_emoji_category7" = "Symbols & Flags";
+"lng_emoji_categories" = "Categories";
+"lng_emoji_search_groups" = "Search groups";
 "lng_emoji_manage_sets" = "Choose emoji set";
 "lng_emoji_set_ready" = "Downloaded";
 "lng_emoji_set_active" = "Current set";
--- a/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
+++ b/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
@@ -3820,7 +3820,43 @@ QAccessible::State EmojiListWidget::accessibilityChildState(
 			state.focused = true;
 		}
 	}
+	// An emoji with variants opens the picker of its skin tones: a
+	// screen reader hears it as collapsed, and expanded while the picker
+	// is up for it.
+	if (over) {
+		const auto emoji = lookupOverEmoji(&*over);
+		if (emoji && emoji->hasVariants()) {
+			state.expandable = true;
+			const auto picked = std::get_if<OverEmoji>(&_pickerSelected);
+			state.expanded = picked
+				&& (*picked == *over)
+				&& !_picker->isHidden();
+		}
+	}
 	return state;
+}
+
+void EmojiListWidget::accessibilityChildShowMenu(quintptr identity) {
+	// Expand opens the picker of variants for the item, Collapse closes
+	// it - the bridge calls this for either, by the state it sees.
+	crl::on_main(this, [=] {
+		const auto index = accessibilityChildIndexByIdentity(identity);
+		const auto over = accessibleChild(index);
+		if (!over) {
+			return;
+		}
+		const auto picked = std::get_if<OverEmoji>(&_pickerSelected);
+		if (picked && *picked == *over && !_picker->isHidden()) {
+			_picker->hideAnimated();
+			return;
+		}
+		keyboardSelect(*over, false);
+		if (!hasFocus()) {
+			_focusReturn = window()->focusWidget();
+			setFocus();
+		}
+		[[maybe_unused]] const auto opened = openKeyboardPicker();
+	});
 }
 
 bool EmojiListWidget::accessibilityChildSupportsActions(int index) const {
--- a/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
+++ b/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
@@ -29,6 +29,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 #include "ui/emoji_config.h"
 #include "ui/painter.h"
 #include "ui/power_saving.h"
+#include "ui/screen_reader_mode.h"
 #include "ui/ui_utility.h"
 #include "ui/cached_round_corners.h"
 #include "boxes/share_box.h"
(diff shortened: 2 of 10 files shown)


john-preston:
**CHANGES_REQUIRED.** Retain the painted-strip implementation, with six functional corrections.

## Required corrections

**P2 — Preserve modified-arrow editing.** [The unconditional customUpDown and Down filter](https://github.com/telegramdesktop/tdesktop/blob/a8a8169af71f29b909abac3d1d90f574d4c2ad2d/Telegram/SourceFiles/ui/controls/tabbed_search.cpp#L495-L503) affect ordinary input. With no reader running, type a matching query, place the caret mid-query, and press Shift+Down. Previously this selected through the end of the single line; now it enters emoji results. The actual inner editor [delegates to InputField](https://github.com/desktop-app/lib_ui/blob/90f0acbd9d3896ed631e54c5348a6cb770c706a2/ui/widgets/fields/input_field.cpp#L1647-L1649), whose [custom-Up/Down branch](https://github.com/desktop-app/lib_ui/blob/90f0acbd9d3896ed631e54c5348a6cb770c706a2/ui/widgets/fields/input_field.cpp#L4397-L4402) ignores these keys regardless of modifiers. Qt propagates the ignored event to the wrapper's new filter, which fires submits and cancels every Down. This bypasses both QTextEdit and InputField's own start/end selection fallback. Configured Qt 5.15.19 confirms Shift+Down maps to SelectNextLine and uses KeepAnchor/End on the final line. Intercept only intended unmodified Down on the inner editor; preserve modified arrows and other caret keys. Shared consumers without a submit subscriber currently just lose the editing operation.

**P2 — Replace the two new positional identities.** [Footer identity](https://github.com/telegramdesktop/tdesktop/blob/a8a8169af71f29b909abac3d1d90f574d4c2ad2d/Telegram/SourceFiles/chat_helpers/stickers_list_footer.cpp#L1129-L1140) and [group identity](https://github.com/telegramdesktop/tdesktop/blob/a8a8169af71f29b909abac3d1d90f574d4c2ad2d/Telegram/SourceFiles/ui/controls/tabbed_search.cpp#L350-L360) encode only index. Retain a provider for set/group A, then reorder/remove it so B occupies that slot: the token still resolves and activates B. [refreshIcons](https://github.com/telegramdesktop/tdesktop/blob/a8a8169af71f29b909abac3d1d90f574d4c2ad2d/Telegram/SourceFiles/chat_helpers/stickers_list_footer.cpp#L1375-L1405) replaces installed-set order; [GroupsStrip::set](https://github.com/telegramdesktop/tdesktop/blob/a8a8169af71f29b909abac3d1d90f574d4c2ad2d/Telegram/SourceFiles/ui/controls/tabbed_search.cpp#L149-L194) replaces server groups. The lib_ui identity cache trusts these tokens. Use set id plus category/settings kind and group icon id, or reject old generations. Reconcile the new keyboard indices with the same keys. This is independent of prerequisite grid/picker identity corrections.

**P2 — Preserve explicit SetFocus targets.** With group A chosen and focus elsewhere, invoke [SetFocus on B](https://github.com/telegramdesktop/tdesktop/blob/a8a8169af71f29b909abac3d1d90f574d4c2ad2d/Telegram/SourceFiles/ui/controls/tabbed_search.cpp#L363-L374). It selects B, calls setFocus, and [focusInEvent immediately selects A](https://github.com/telegramdesktop/tdesktop/blob/a8a8169af71f29b909abac3d1d90f574d4c2ad2d/Telegram/SourceFiles/ui/controls/tabbed_search.cpp#L403-L419). [Footer SetFocus](https://github.com/telegramdesktop/tdesktop/blob/a8a8169af71f29b909abac3d1d90f574d4c2ad2d/Telegram/SourceFiles/chat_helpers/stickers_list_footer.cpp#L1143-L1155) likewise loses its requested child to the [active section on entry](https://github.com/telegramdesktop/tdesktop/blob/a8a8169af71f29b909abac3d1d90f574d4c2ad2d/Telegram/SourceFiles/chat_helpers/stickers_list_footer.cpp#L1190-L1209). The announcement and subsequent Enter therefore concern A. Apply default entry selection for actual Tab/Backtab traversal; preserve a pending explicit child for OtherFocusReason/accessibility entry. Do not remove the explicit child-focus event needed by AT-SPI.

**P2 — Make result handoff explicit and query-scoped.** [showSearchResults](https://github.com/telegramdesktop/tdesktop/blob/a8a8169af71f29b909abac3d1d90f574d4c2ad2d/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L1340-L1404) infers activation from groupsHaveFocus(). Choose a group, return to/browse its strip while another cloud/set response is outstanding, and the response takes focus away without another activation. Conversely, activate the same chosen group after loading completes: the query is unchanged, [applyNextSearchQuery returns](https://github.com/telegramdesktop/tdesktop/blob/a8a8169af71f29b909abac3d1d90f574d4c2ad2d/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L899-L903), and no callback transfers focus. Local/cached application also fills results directly; field submission with zero results records no pending handoff. Track explicit field/group activation against its query/generation, service it from the common result-update path, and cancel it on deliberate focus departure or query replacement. Already populated groups should transfer immediately.

**P2 — Carry the search-owned return destination into results.** [The new submit handler](https://github.com/telegramdesktop/tdesktop/blob/a8a8169af71f29b909abac3d1d90f574d4c2ad2d/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L875-L882), and the group-result branch, directly focus the grid. For EmojiStatus, FullReactions and UserpicBuilder, [afterShown](https://github.com/telegramdesktop/tdesktop/blob/a8a8169af71f29b909abac3d1d90f574d4c2ad2d/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp#L1966-L1978) saved the opening control only inside SearchWithGroups::stealFocus; grid _focusReturn remains empty. On closing, [SearchWithGroups::returnFocus](https://github.com/telegramdesktop/tdesktop/blob/a8a8169af71f29b909abac3d1d90f574d4c2ad2d/Telegram/SourceFiles/ui/controls/tabbed_search.cpp#L750-L767) sees that its field no longer has focus, clears the destination, and restores nothing. The grid also has no return target, so TabbedSelector falls back to window focus. Preserve one external destination across internal panel handoffs, restoring only while the panel still owns focus. This is the new search transition, separate from #31241's picker-to-picker ownership problem.

**P2 — Apply live-mode policy to the new strips.** Focus either strip, then disable reader mode. FocusManager changes policy to NoFocus without clearing existing focus, while [the footer key handler](https://github.com/telegramdesktop/tdesktop/blob/a8a8169af71f29b909abac3d1d90f574d4c2ad2d/Telegram/SourceFiles/chat_helpers/stickers_list_footer.cpp#L1213-L1244) and [group key handler](https://github.com/telegramdesktop/tdesktop/blob/a8a8169af71f29b909abac3d1d90f574d4c2ad2d/Telegram/SourceFiles/ui/controls/tabbed_search.cpp#L422-L448) still navigate and activate. Their queued SetFocus callbacks also lack execution-time mode checks. Carry the prerequisite live-mode policy into these two new widgets and release their reader-owned focus on deactivation. Keep intentional ordinary search submission separate; correcting only the earlier grid/picker handlers cannot cover these additions.

PR #31243, opened 2026-09-05
Walk the stickers and the GIFs of the panel with the keyboard

rezabakhshilaktasaraei:
On top of #31242. The Stickers and GIFs tabs of the panel were as the Emoji tab was before #31240: painted grids with nothing for a screen reader and no keyboard.

- **Stickers** (`StickersListWidget`): a list named "Stickers"; each sticker named by the emoji it stands for, described by its set's title. Arrows, Home/End, page keys as in the emoji list; Enter or Space sends the sticker and hides the panel with the focus back in the composer; Escape gives up the same way. A set chosen in the footer lands on its first sticker; Enter/Down in the search field or a group chosen beside it lands on the first result once results are in.
- **GIFs** (`GifsListWidget`): a list named "GIFs"; items are named "GIF" (the app's own label for animations — they have no text) with their position from the set. Enter sends right away, as with Ctrl held, since the keyboard can't see whether the preview a click waits for has loaded. The GIF footer names its sections by their emoji.
- **Tab strip**: a tab switched to from the strip with a screen reader leaves the focus on the strip — the switch no longer resets it to the window (`TabbedSelector::beforeHiding`), the strip isn't hidden for the slide or its picture (it wasn't in the picture anyway), and the tab shown doesn't take the focus into its search then (`Inner::setKeepsFocusOnShow`). Opening the panel afresh behaves as before. This part matters once the strip is focusable (#31097); it compiles independently.

Nothing changes without a screen reader. Verified with NVDA.

### DIFF
--- a/Telegram/Resources/langs/lang.strings
+++ b/Telegram/Resources/langs/lang.strings
@@ -3225,6 +3225,8 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 "lng_emoji_category5" = "Travel & Places";
 "lng_emoji_category6" = "Objects";
 "lng_emoji_category7" = "Symbols & Flags";
+"lng_emoji_categories" = "Categories";
+"lng_emoji_search_groups" = "Search groups";
 "lng_emoji_manage_sets" = "Choose emoji set";
 "lng_emoji_set_ready" = "Downloaded";
 "lng_emoji_set_active" = "Current set";
--- a/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
+++ b/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
@@ -3820,7 +3820,43 @@ QAccessible::State EmojiListWidget::accessibilityChildState(
 			state.focused = true;
 		}
 	}
+	// An emoji with variants opens the picker of its skin tones: a
+	// screen reader hears it as collapsed, and expanded while the picker
+	// is up for it.
+	if (over) {
+		const auto emoji = lookupOverEmoji(&*over);
+		if (emoji && emoji->hasVariants()) {
+			state.expandable = true;
+			const auto picked = std::get_if<OverEmoji>(&_pickerSelected);
+			state.expanded = picked
+				&& (*picked == *over)
+				&& !_picker->isHidden();
+		}
+	}
 	return state;
+}
+
+void EmojiListWidget::accessibilityChildShowMenu(quintptr identity) {
+	// Expand opens the picker of variants for the item, Collapse closes
+	// it - the bridge calls this for either, by the state it sees.
+	crl::on_main(this, [=] {
+		const auto index = accessibilityChildIndexByIdentity(identity);
+		const auto over = accessibleChild(index);
+		if (!over) {
+			return;
+		}
+		const auto picked = std::get_if<OverEmoji>(&_pickerSelected);
+		if (picked && *picked == *over && !_picker->isHidden()) {
+			_picker->hideAnimated();
+			return;
+		}
+		keyboardSelect(*over, false);
+		if (!hasFocus()) {
+			_focusReturn = window()->focusWidget();
+			setFocus();
+		}
+		[[maybe_unused]] const auto opened = openKeyboardPicker();
+	});
 }
 
 bool EmojiListWidget::accessibilityChildSupportsActions(int index) const {
--- a/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
+++ b/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
@@ -1967,7 +1967,9 @@ void EmojiListWidget::afterShown() {
 	const auto steal = (_mode == Mode::EmojiStatus)
 		|| (_mode == Mode::FullReactions)
 		|| (_mode == Mode::UserpicBuilder);
-	if (_search && steal) {
+	if (keepsFocusOnShow()) {
+		return;
+	} else if (_search && steal) {
 		_search->stealFocus();
 	} else if (Ui::ScreenReaderModeActive() && !hasFocus()) {
 		// The panel is opened from the keyboard and has no one to talk
--- a/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
+++ b/Telegram/SourceFiles/chat_helpers/emoji_list_widget.cpp
@@ -29,6 +29,7 @@ https://github.com/telegramdesktop/tdesktop/blob/master/LEGAL
 #include "ui/emoji_config.h"
 #include "ui/painter.h"
 #include "ui/power_saving.h"
+#include "ui/screen_reader_mode.h"
 #include "ui/ui_utility.h"
 #include "ui/cached_round_corners.h"
 #include "boxes/share_box.h"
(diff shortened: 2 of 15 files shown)


elfelfdeewun933-sys (approved):

john-preston:
**CHANGES_REQUIRED.** Keep the feature and extractable tab-strip focus fix, with six functional corrections.

## Required corrections

**P2 — Preserve semantic identity and keyboard selection through rebuilds.** [Sticker tokens](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/stickers_list_widget.cpp#L3132-L3151) encode section/index; [GIF tokens](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/gifs_list_widget.cpp#L1038-L1051) encode mosaic position. Retain a provider, then remove/reorder a saved GIF or installed sticker set: its token now chooses the replacement in that slot. [Saved GIF rebuild](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/gifs_list_widget.cpp#L616-L642) and [installed-set rebuild](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/stickers_list_widget.cpp#L3445-L3460) establish ordinary triggers without overflow assumptions.

The same content-reconciliation fix must cover live keyboard selection. Navigate toward the end of GIF results with the mouse stationary elsewhere; pagination calls [refreshInlineRows](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/gifs_list_widget.cpp#L731-L768), clearing selection and replacing it with the mouse target or -1. The new showInlineRows restores only footer/group entry for new results, never a keyboard position for appended results. Sticker refill restores an ordinal, not a semantic item. Use stable semantic keys or invalidating generations for providers, and reconcile keyboard selection by the same key throughout refresh/reflow.

**P2 — Enumerate only rendered featured-pack cells.** [The new child mapping](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/stickers_list_widget.cpp#L3017-L3033) uses SectionInfo.count, the whole loaded sticker array. External-layout packs [allocate one row](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/stickers_list_widget.cpp#L518-L551) and [paint only its columns](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/stickers_list_widget.cpp#L1731-L1745), while appendSet retains the complete loaded pack. For a pack longer than one row, Right/Down reaches unpainted stickers; their rectangles overlap later packs or exceed allocated height, and Enter chooses invisible content. Clamp preview navigation/count/geometry to rendered columns. Use the existing set-opening action for access to the remaining pack.

**P2 — Close only after GIF choice acceptance.** [Keyboard activation](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/gifs_list_widget.cpp#L1150-L1160) always hides after selectInlineResult(..., true). Searched GIF layouts use ItemBase(context, result), storing _result rather than _document, so they reach [the getResult branch](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/gifs_list_widget.cpp#L539-L548). That branch ignores forceSend. [Result::onChoose](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/inline_bots/inline_bot_result.cpp#L324-L344) returns false for an active media view with an unloaded preview. No choice event is emitted, but the panel closes. Have the choice helper return acceptance and close only then, or implement the intended forced choice consistently for result-backed GIFs. This boundary is local choice emission, not network send completion.

**P2 — Make result handoff explicit and query-scoped.** [The GIF submit handler](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/gifs_list_widget.cpp#L870-L894) focuses the first existing item while [a new uncached query retains old results](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/gifs_list_widget.cpp#L898-L917). Withhold B's response, submit B, then press Enter again: A's old result can be chosen. Async callbacks also infer activation from footer/group focus, allowing late responses to interrupt browsing; reactivating an unchanged group need not generate a callback at all. [The new sticker consumer](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/stickers_list_widget.cpp#L770-L797) repeats the focus inference. Record explicit activation against a query generation, service matching cached/current results immediately or a pending handoff later, and cancel it when query/focus intent changes. Fixing #31242's emoji consumer alone is insufficient.

**P2 — Extend live-mode handling to these grids.** [Sticker keys](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/stickers_list_widget.cpp#L3335-L3366), [GIF keys](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/gifs_list_widget.cpp#L1188-L1220) and their queued SetFocus callbacks have no execution-time mode check. Stop the reader while either grid holds focus: NoFocus changes future traversal but retains current focus, so arrows and choice keys remain active. Queued setFocus can reacquire focus after deactivation. Extend the shared mode/focus policy to these classes. The new search submits also enter grids with no reader; if ordinary Enter/Down browsing is intentional, define its coherent behavior separately and narrow the description's “nothing changes” claim.

**P2 — Wire sticker cancellation in every host.** [Escape is accepted after hideRequests](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/stickers_list_widget.cpp#L3359-L3366), but [TabbedSelector wires only its Stickers instance](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/chat_helpers/tabbed_selector.cpp#L757-L774), omitting Masks, which uses the same class. Direct [message-effects](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/history/view/reactions/history_view_reactions_selector.cpp#L1148-L1173) and [userpic](https://github.com/telegramdesktop/tdesktop/blob/64f71017d8a722837ace5437096d13c21632323f/Telegram/SourceFiles/info/userpic/info_userpic_emoji_builder_widget.cpp#L267-L281) hosts also omit the stream. Tab into one of these grids and press Escape: normal traversal has no saved return widget, cancellation is unobserved, and the existing host key handler never receives the accepted event. Wire one cancellation contract for every host or preserve propagation where no owner handles it.

Issue #30489, opened 2026-03-29
100%+ CPU load when rendering animated avatar in profile preview


kolayne:
### Steps to reproduce

1. Open a profile of someone with animated avatar.

Or

1. Open the sidebar profile preview in a chat with someone with animated avatar.

### Expected behaviour

Low CPU load.

### Actual behaviour

~120% CPU load (where 100% is one core fully loaded).

Note that it only happens in profile preview or sidebar preview. The CPU load is fine when animated avatars appear in the chat history and when I open an animated avatar in fullscreen mode.

### Operating system

Arch Linux, Hyprland 0.54.2

### Version of Telegram Desktop

6.6.4 beta

### Installation source

Other (unofficial) source

### Logs

```shell

```

SwakshGupta:
i can fix dis

kolayne:
@SwakshGupta, if you can, please, submit your fix via a Pull Request!

ilya-fedin:
Can you reproduce with official binary? That's likely expected though since both video decoding and UI (including video) drawing happens on CPU. Only media viewer draws on GPU (still decodes on CPU by default though since GPU decoding is bugged).

kolayne:
@ilya-fedin, yes, reproduced with the official binary for Linux.

Note that animated avatars rendering in the chat history does not spike the CPU usage. It only happens in the profile view or the chat sidebar

ilya-fedin:
> Note that animated avatars rendering in the chat history does not spike the CPU usage. It only happens in the profile view or the chat sidebar

I believe you just mistest it

kolayne:
I'm don't think there's much to mistest... I just close the chat sidebar and the CPU usage drops; I open it again and the CPU usage spikes again.

Retested again more carefully on the official binary:
- Moving cursor slowly in a chat with no animated avatars: CPU usage ~1%.
- Moving cursor slowly in a chat with the same animated avatar appearing 5 times in the chat history: CPU usage ~4%
- Same plus open a sidebar: CPU usage ~115%

ilya-fedin:
I mean their size is very different. Both sidebar and profile have a larger drawing area for the avatar than the history widget. So you get different CPU load of course. The larger video, the larger CPU usage.

john-preston:
@kolayne That’s interesting, usually nothing can be done with high cpu load in TDesktop, because it happens just due to repainting, but the difference is very large in your case, maybe there is something really wrong going on in chat profile indeed..

[...]

PR #31244, opened 2026-09-05
Repaint video userpic in profile top bar only on new frames

hatomist:
With the profile column open on a chat that has a video userpic, `TopBar::paintUserpic()` re-queues itself via `update()` on every paint. Since the default surface format uses swap interval 0 on Linux, the loop is unthrottled.

Measured on Linux/Wayland (KWin, AMD Raphael iGPU, 3840x1100 window), Flatpak 7.1.3:
- before: ~420 backing-store flushes/s, each re-uploading the 584x473 top bar and redrawing the whole window; the process held 112-132% of the iGPU 3D engine busy time
- after: ~30 flushes/s (the video frame rate), 4-14%

Built with the CentOS build image and verified on the same setup.

### DIFF
--- a/Telegram/SourceFiles/info/profile/info_profile_top_bar.cpp
+++ b/Telegram/SourceFiles/info/profile/info_profile_top_bar.cpp
@@ -2782,7 +2782,6 @@ void TopBar::paintUserpic(QPainter &p, const QRect &geometry) {
 		if (!frame.isNull()) {
 			auto hq = PainterHighQualityEnabler(p);
 			p.drawImage(geometry, frame);
-			update();
 			return;
 		}
 	}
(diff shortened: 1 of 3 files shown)


ilya-fedin:
if i read the description right, this fixes https://github.com/telegramdesktop/tdesktop/issues/30489?

hatomist:
Yes, looks like it. Same rendering path, just spins either on CPU raster (RHI off), or spins on gpu side with opengl/vulkan rendering.

Just tested on cpu; unpatched vs patched dev (7.2.6): 97% -> 9% CPU, ~2100 -> ~70 window commits/s.

ilya-fedin:
> just spins either on CPU raster (RHI off), or spins on gpu side with opengl/vulkan rendering

Are you sure? Media on main window still always draws with CPU

hatomist:
Yes, frame itself is always drawn by CPU, I missed a number in the RHI ON comparison.

Pre patch:
- RHI on - ~90% CPU AND saturates iGPU 3D engine
- RHI off - ~100% CPU, no GPU

Post patch:
- RHI on: ~8% CPU, ~5% GPU
- RHI off: ~9% CPU, no GPU

Render is CPU in both cases, downstream of that RHI ON adds a GPU step per flush: upload the dirty region, redraw the whole window. So GPU load is technically a side effect, I didn't separate the two at first properly as it mostly showed up for me as desktop/compositor lag while the 9950x3d2 host cpu didn't suffer much from one spinning core out of 32

PR #31287, opened 2026-09-13
Show back button on My Profile when opened in a stack

meh7an:
My Profile draws its own top bar, which has a back button of its own. That button is shown through `_backToggles` in `Info::Stories::InnerWidget`, which `InnerWidget::enableBackButton()` sets.

When a page is pushed on top of another section, `WrapWidget::showNewContent()` calls `enableBackButton()` on the content widget. `Info::Profile::Widget` forwards that call to its inner widget, but `Info::Stories::Widget` did not override it, so the empty `ContentWidget` implementation ran and the back button never appeared. A stacked My Profile could only be left by closing the whole layer.

`Info::Stories::Widget` now forwards `enableBackButton()` to its inner widget, the same way `Info::Profile::Widget` does.

**Testing**

Built and tested on macOS (Debug), with My Profile pushed on top of Settings through #31288: the back button is shown and returns to Settings.

### DIFF
--- a/Telegram/SourceFiles/info/stories/info_stories_widget.cpp
+++ b/Telegram/SourceFiles/info/stories/info_stories_widget.cpp
@@ -283,6 +283,10 @@ void Widget::setupBottomButton(int wasBottomHeight) {
 	}
 }
 
+void Widget::enableBackButton() {
+	_inner->enableBackButton();
+}
+
 void Widget::showFinished() {
 	_shown = true;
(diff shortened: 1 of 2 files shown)


23rd:
@meh7an 
Please record a short demo-video how it worked before your code changes and a short demo-video how it works after your code changes, for all your opened pull requests. After that, you can tag me so I can pay attention again.

PR #31306, opened 2026-09-14
Fix crash sending many tracks from All Music in slow mode

meh7an:
`GetErrorForSending` checks slow mode limits before sending. When more than one message would be sent and there is no text, it assumes the messages are being forwarded and checks whether they form a single album, reading `request.forward->front()`.

The All Music box calls it with only `messagesCount`, the number of albums its selected tracks are split into, and without a forward list. Selecting more than ten tracks without a caption makes that count larger than one, so in a slow mode group the check dereferenced a null forward list and the app crashed.

A missing or empty forward list is now treated as not a single forwarded album, so a caller that passes only a message count gets the usual "Slow mode is enabled. You can't send more than one message at a time." error instead.

**Testing**

Built and tested on macOS (Debug), as a regular member of a supergroup with slow mode enabled:

- All Music with more than ten tracks and no caption: the slow mode error is shown instead of a crash.
- All Music with up to ten tracks: the tracks are sent as before.


john-preston:
Thanks!

PR #31307, opened 2026-09-14
Fix crash deleting selected messages removed elsewhere

meh7an:
`ListWidget` shows messages in forum topics, comments, saved sublists, and the pinned, scheduled and welcome messages sections. When a message was removed, `ListWidget::itemRemoved` cleared its view and hover state but kept it in the selection.

A message can be removed while it is selected in several ways: another admin or another device deletes it, its self-destruct timer destroys it, or, in the scheduled messages section, it gets sent when its time comes. The top bar then kept showing the old selection count, and the delete confirmation still counted the removed messages. `ConfirmDeleteSelectedItems` skips messages that no longer exist, so when every selected message was gone it passed an empty list to `CanCreateModerateMessagesBox`, and the `!items.empty()` assertion in `CalculateModerateOptions` crashed the app.

`ListWidget::itemRemoved` now removes the message from the selection with the existing `removeItemSelection()`, which also updates the top bar. This matches how `HistoryInner::itemRemoved` handles the main chat view.

**Testing**

- In Scheduled messages, select a message scheduled a minute ahead and wait until it is sent: the selection bar goes away and nothing crashes. Before the fix, pressing Delete at that point crashed with `Assertion Failed! "!items.empty()" moderate_messages_box.cpp:118`.


john-preston:
Thanks!

PR #31308, opened 2026-09-14
Fix crash archiving a chat inside a collapsed community

meh7an:
A chat that belongs to a community shown as one chat lives in the community's own chats list, which is what `Data::Session::chatsListFor()` returns for it. `History::setFolderPointer()` ignored that and moved the chat's row between the main chats list and the Archive through `chatsList(folder)`.

So when such a chat was archived or unarchived:

- its row was removed from a list that didn't contain it, leaving an orphan row in the community list and updating the wrong list's unread state;
- a second row was added to the Archive or the main list;
- the following `refreshChatListEntry()` adjusted the community list with that foreign row and failed the `row->index() >= 0 && row->index() < _rows.size()` assertion in `Dialogs::List::adjustByDate()`.

The chat menu hides Archive for these chats, but their folder can still change from another device, when an archived chat is unarchived by a new message, or through the swipe quick action. Once the server has such a chat archived, the crash also repeats shortly after every launch.

`setFolderPointer()` now resolves both lists through `chatsListFor()`, so a community chat stays in the community list, while other chats move between the main list and the Archive as before.

**Testing**

Built and tested on macOS (Debug), with a community shown as one chat:

- Archiving one of its chats with the swipe quick action no longer crashes. Before the fix it failed the `dialogs_list.cpp:86` assertion right away, and again about 20 seconds after each launch.
- The chat stays inside the community without a duplicate row, and unarchiving it doesn't crash.


john-preston:
Thanks!

PR #31310, opened 2026-09-14
Fix crash deleting a topic open in its own window

meh7an:
When a forum topic open in its own window is deleted, `ChatWidget` handles `ForumTopic::destroyed()` in `closeCurrent()` by calling `Window::Controller::close()` for that window. That goes through `MainWindow::closeEvent()` and `Core::Application::closeNonLastAsync()`, so the window is destroyed only on the next event loop iteration.

`Forum::applyTopicDeleted()` doesn't wait for that. Right after firing the destroyed event it erases the topic, freeing it, and then calls `History::destroyMessagesByTopic()`. The topic section in the still open window refreshes its message list and reads the freed topic, for example in `CornerButtons::updateUnreadThingsVisibility()` through `cornerButtonsThread()`, and the app crashes. Even without that crash, the queued close reads the topic again: `SavedWindows::windowClosed()` serializes it from the window id, and the `ChatWidget` destructor saves the list state into it.

`closeCurrent()` now closes such a window with `Core::Application::closeWindow()` right away, the same way `SessionController` closes separate forum, archive and community windows. The window and its section are destroyed while the topic is still alive.

**Testing**

Built and tested on macOS (Debug), with `MallocScribble=1` so any use of freed memory crashes reliably:

- Deleting a topic from the main window while it is open in its own window closes that window and the app keeps running. Before, it crashed in `CornerButtons::updateUnreadThingsVisibility()`.
- Deleting a topic from the menu of its own window closes that window without a crash.
- Deleting a topic open in the main window goes back to the topics list as before.


john-preston:
Thanks!

PR #31311, opened 2026-09-14
Fix crash returning to a live story with unread comments

meh7an:
In the stories viewer, `ReplyArea::show()` calls `ComposeControls::setToggleCommentsButton()` whenever the shown story switches between a live stream and a regular story. For a regular story it deletes the comments toggle button. The dot that marks new comments, `_commentsShownNewDot`, is a child of that button, so it was destroyed together with it, but the pointer kept its old value.

Returning to a live story creates a new button. While the comments state is `WithNew`, `setupCommentsShownNewDot()` sees the non-null pointer and doesn't create a dot. Any other non-empty state runs `delete base::take(_commentsShownNewDot)` on the destroyed dot. The rejoined stream first reports an empty comments list and then loads the comments, switching the state to `Shown`, so the app crashed right after coming back to the live story.

`setToggleCommentsButton()` now clears the dot pointer and deletes the previous button before creating a new one. This also covers moving from one live story straight to another, where a second button used to be created while the first one and its subscriptions stayed alive.

**Testing**

Built and tested on macOS (Debug), with `MallocScribble=1` so any use of freed memory crashes reliably:

- Opened a live story, hid the comments and waited for the new comments dot, then moved to a regular story and back. Before, the app crashed as soon as the comments loaded. Now it keeps running.


john-preston:
Thanks!

PR #31313, opened 2026-09-14
Fix crash in one-tab sticker pickers after adding stickers

meh7an:
`TabbedSelector` subscribes to `Data::Stickers::stickerSetInstalled()` when it has a stickers tab, and the handler calls `_tabsSlider->setActiveSection()` to switch to that tab. The event fires when a sticker set is added, and also when a sticker is added to favorites, because `Stickers::setIsFaved()` reports the favorites set as installed.

`_tabsSlider` is only created when the selector has more than one tab. Pickers in `Mode::StickersOnly` and `Mode::ChatIntro` have just the stickers tab, so the slider is null there. These are the poll option sticker picker from `CreatePollStickerPanel()`, used in the create poll box and when adding a poll option, the chat intro sticker picker in Business settings, and the adapt sticker picker of an owned emoji set. While any of them existed, adding a sticker set or a favorite sticker anywhere in the app crashed in `Ui::DiscreteSlider::setActiveSection()`.

The handler now switches the slider only when it exists. A single-tab picker already shows its stickers tab, and it still scrolls to the added set.

**Testing**

- Opened "Choose Sticker" for a poll option in the create poll box, kept the box open and added a sticker to favorites from another window. Before, the app crashed. Now it keeps running, and the picker shows the new favorite when opened again.
- Adding a sticker to favorites with the regular emoji and sticker panel still switches it to the stickers tab.


23rd:
Your fix looks clean and logical. But to be honest I can't to reproduce the crash.

meh7an:
> Your fix looks clean and logical. But to be honest I can't to reproduce the crash.

A detailed STR:
1. Open a chat in a new window
2. Send a sticker there in the new window
3. Go back to main window and open a chat and try to create a poll.
4. In poll creation, click on attach button, then "Choose sticker" and then nothing else, just switch back to new window.
5. Try to add the sticker you had sent before to favorites. 
App will now crash.

Here is the Crash ID: ```ad10f643-483c-4a47-9b0d-b46c9b0b2923```

PR #31316, opened 2026-09-14
Fix crash sending rotated videos from the photo editor

meh7an:
The photo editor can place videos on an edited photo. When the result is sent, `Media::Encode` renders each video item with `FFmpeg::FrameGenerator`, requesting frames at the item's size. A video with rotation metadata, like a portrait phone video, crashed the app there on `Assert(dx >= 0 && dy >= 0 && (!dx || !dy))` in `FrameGenerator::Impl::renderCurrent()`.

`renderCurrent()` draws the frame in its stored orientation and rotates the whole image by `_rotation` at the end. But it fitted the stored frame size into the requested size, which is in display orientation, and then swapped the width and height of the result for rotations by 90 or 270 degrees. For a requested size that isn't square, the swapped result is larger than the requested size in one dimension. For example, a 960x540 frame rotated by 90 degrees and requested at 270x480 was scaled to 270x480 and swapped to 480x270, so the horizontal offset became negative and the assertion failed.

Now the requested size is swapped for such rotations before scaling. The frame is scaled, centered and padded in its stored orientation, and the existing rotation at the end turns it into an image of the requested size. This also keeps the aspect ratio of rotated frames requested with `Qt::KeepAspectRatio`, which were squeezed into the swapped box before.

**Testing**

Built and tested on macOS (Debug), with a 960x540 H.264 video marked as rotated by 90 degrees:

- Attached a photo, added the video in the photo editor and sent it. Before, the app crashed while preparing the upload. Now it sends, and the sent video is portrait and upright, the same as in the editor.


john-preston:
Thanks!

PR #31317, opened 2026-09-14
Fix crash opening bot web views for unavailable chats

meh7an:
461872c1d2 added `Window::CanShowSeparateWindow()` and made `Core::Application::ensureSeparateWindowFor()` assert it. The check fails for chats with an unavailable reason, and while the app is locked by passcode or the session is locked by terms of service. That commit added the check to the callers of `ensureSeparateWindowFor()`, but `WindowForThread()` in `bot_attach_web_view.cpp` still calls it directly. It does so when there is no separate window for the chat yet and the window that started the flow is gone or has no chats list, for example a chat opened in its own window.

`WindowForThread()` finds the window for a chat when choosing a chat for an attachment menu bot link with `startattach` and `choose`, when a web app's privacy policy sends the bot's command, and when a web app's request to join a chat is approved. The attachment menu chooser doesn't filter out unavailable chats, so choosing one of them from such a window would fail the assertion.

`WindowForThread()` now returns no window when a separate window can't be shown, and all callers already handle that. The web app isn't opened for the chosen chat, the privacy policy command isn't sent, and the approved chat isn't opened, while its toast is still shown.

**Testing**

Built on macOS (Debug). I couldn't reproduce the crash itself, since it needs a chat that is unavailable on desktop, an attachment menu bot that lets you choose chats, and a flow started from a window without a chats list. Opening a bot's Mini App from a chat opened in its own window still works as before.


john-preston:
Thanks!

PR #31318, opened 2026-09-14
Fix crash forwarding to a topic from a separate window

meh7an:
When a single chat is chosen in the forward box, it finds a window to put the forward draft into. The window id used `SeparateType::Forum` whenever the chosen chat belongs to a forum that lists its topics instead of showing tabs, even when the chosen chat is one of its topics.

`Core::Application::windowFor()` finds no window with such an id. If the topic's account has no main window, for example when forwarding from a chat of one account opened in its own window while the main window shows another account, it falls back to the main window of the other account, and the forward box calls `ensureSeparateWindowFor()` with the forum id. The new window's `SessionNavigation::showByInitialId()` then calls `showForum(id.forum())`, but `SeparateId::forum()` is null for a topic, so constructing `not_null<Data::Forum*>` fails the assertion. Such an id also never matches a window already open for that topic.

The forward box now uses `SeparateType::Forum` only when the chosen chat isn't a topic, the same condition as the "Open in new window" menu item in `Filler::addNewWindow()`. A topic gets a chat window, so the forward draft goes to a new window for that topic, or to the topic's window if it is already open.

**Testing**

Built and tested on macOS (Debug) with two accounts:

- Opened a chat of one account in its own window, switched the main window to the other account, forwarded a message from that window and chose a topic of a forum with the list layout. Before, the app crashed. Now a window for the topic opens with the forward draft.
- With that topic already open in its own window, forwarding to it the same way puts the draft into that window.
- Forwarding to a topic while the main window shows the same account still shows the forward in the main window.


john-preston:
Thanks!

PR #31323, opened 2026-09-15
Fix crash clicking the add photo suggestion

meh7an:
For an account without a profile photo the server can offer the `USERPIC_SETUP` suggestion, shown above the chat list as "Add your photo!". Clicking it crashed the app, including the official 7.2.8 build.

`suggestion_userpic_setup.cpp` creates a `Ui::UserpicButton` for choosing the photo and puts it into the suggestion with `TopBarSuggestionContent::setLeadingWidget()`. The suggestion's clicked callback forwarded the click to that button with `Ui::SendSynteticMouseEvent()`, sending a move, a press and a release to the window at the button's top left corner. Such events are delivered to the widget found under that point, but `setLeadingWidget()` makes the leading widget `Qt::WA_TransparentForMouseEvents`, so the widget found is the suggestion itself. Its clicked callback ran again and sent the events again, until the stack overflowed. The crash dumps of the official build show the main thread overflowing its stack inside a long repeated chain of the same frames.

The callback now calls `upload->clicked({}, Qt::LeftButton)`, which runs the button's click handler and opens its photo menu without any hit testing. The `ui/ui_utility.h` include was only needed for the synthetic events and is removed.

**Testing**

Built and tested on macOS (Debug). The suggestion is shown only when the server offers it, so it was tested with a local build that always showed it:

- Clicking the suggestion text opens the photo menu instead of crashing.
- Clicking the photo icon of the suggestion opens the same menu.
- Choosing "Photo or video" in the menu opens the file picker.


john-preston:
Thanks!

PR #31331, opened 2026-09-17
Fix crash sending from the Music box after its topic was deleted

meh7an:
The Music item of the attach menu opens `MusicAttachBox` with the `actionFactory` the menu got from the chat. The other items of that menu call the factory when they are clicked, but the box keeps it and calls it only when sending (`sendSelected()`, `sendPreparedMusic()` and the files box it opens for "Choose from files"). For a topic, comments and other `ChatWidget` sections, this factory calls `ChatWidget::prepareSendAction()` on a raw `this`.

The box is shown in the window, not in the section. When a topic open in the main window is deleted, for example from another device, `ChatWidget::closeCurrent()` goes back from the stack with `anim::activation::background`, which destroys the section without hiding the window's layers. The box stays open, and pressing Send calls `prepareSendAction()` on the destroyed section, crashing in `ComposeControls::replyingToMessage()`.

The attach menu now closes the Music box when its parent widget is destroyed. For sections this is the `ChatWidget` itself. For regular chats it is `HistoryWidget`, which stays alive when switching chats, so nothing changes there. The files box opened from the Music box already calls back through `crl::guard(box, ...)`, so it stops sending once the Music box is gone. The factory is still called only when sending, because `prepareSendAction()` reserves a new topic in forums of bots where the user creates topics, and that shouldn't happen for a box that gets cancelled.

**Testing**

- With a topic open in the main window and the Music box open, deleting the topic from another device and pressing Send in the box crashed before. Now the box closes together with the topic.


john-preston:
Thanks!

PR #31344, opened 2026-09-20
Play in-chat sounds for sent and received messages

rezabakhshilaktasaraei:
Adds the in-chat sounds the mobile apps have: a short tick when a message arrives in the chat you are looking at, and another when your message is sent. Requested for the desktop app in #476 (https://github.com/telegramdesktop/tdesktop/issues/476#issuecomment-5521753795, point 1). It is a regular feature for everyone, not tied to the screen reader mode.

**Behaviour**, following Telegram for Android (`NotificationsController.playInChatSound` / `playOutChatSound`):
- Incoming: played when a new message lands in the chat, topic or reply thread that is open while the window is active, the same moment it is marked as read and no notification is shown. Not played for silent messages, muted chats or while recording a voice message. At most once per 500 ms.
- Outgoing: played when the server confirms a sent message (the local id becomes a real one), not on pressing Send. Not played for messages sent without sound or scheduled ones. At most once per 100 ms.
- The sounds are Android's `sound_in.wav` and `sound_out.wav`, unchanged.

**Setting**: "In-chat sounds" in Settings > Notifications, under "Allow sound". On by default, like on Android; easy to flip if you prefer it off for desktop. The ticks also respect "Allow sound" and system quiet hours.

**Implementation**
- `Window::Notifications::System::playInChatSound()` owns the two lazily created tracks and the throttling. It goes through a new `Manager::maybePlayInChatSound()`, because `maybePlaySound()` is gated by `volumeSupported()`, which is false when native toasts play the notification sound themselves.
- `HistoryWidget` plays the incoming tick in `newItemAdded()` inside the `markingMessagesRead()` branch, and the outgoing one on `Data::Session::itemIdChanged()`.
- `HistoryView::ChatWidget` does the same from `listItemsAddedToEnd()` and its own `itemIdChanged()` subscription, limited to messages of the shown thread or sublist.
- The setting is stored through the generic prefs (`in-chat-sounds`, read with `true` as the fallback), so the settings stream is untouched.

**Second commit**, separate so it can be dropped on its own: a reply that arrives in a plain reply thread that is open and active is marked read in the thread, but today it still shows a notification and plays the full sound, because the chat itself stays unread. With the tick that would be two sounds for one message you are already looking at. `System::clearIncomingFromItem()` cancels that single notification (the toast, the queued item and its alert time), the way reading the chat or a topic already does for theirs. Other pending notifications of the chat are untouched.


Telegram/SourceFiles/core/core_settings.h
```diff
@@ -39,6 +39,8 @@ namespace Core {
 
 inline constexpr auto kScreenReaderModeDisabledKey
 	= "screen-reader-mode-disabled"_cs;
+inline constexpr auto kInChatSoundsDisabledKey
+	= "in-chat-sounds-disabled"_cs;
```
ilya-fedin: Why do you store inverted value? Name it just `in-chat-sound` and  remove the inversion in getter/setter?

rezabakhshilaktasaraei: Done, thanks. I copied the inverted form from `screen-reader-mode-disabled` and missed that `readPref` takes a fallback. It is now `in-chat-sounds`, read with `true` as the default and written as is.

